PR #6, opened 2025-08-07
Cleanup: v2 AST rename directories

rjb32:


rjb32:
@LucLabarriere I just did a simple rename of AST/attribution into AST/decl. Not sure it should be the definitive organization either. And a few random stuff

PR #23, opened 2025-09-01
Build StringIndex using temp ID mapping

cyrusknopf:
The string index was previously built using the temporary IDs of nodes/edges instead of the permanent ones, meaning sometimes it would return incorrect results. This PR fixes that issue.

### DIFF
--- a/storage/indexers/StringPropertyIndexer.h
+++ b/storage/indexers/StringPropertyIndexer.h
@@ -11,6 +11,10 @@
 namespace db {
 
 class StringPropertyIndexer {
+private:
+    template <TypedInternalID IDT>
+    using TempIDMap = std::unordered_map<IDT, IDT>;
+
 public:
     StringPropertyIndexer() = default;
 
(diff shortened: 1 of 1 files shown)


storage/indexers/StringPropertyIndexer.h
```diff
@@ -41,8 +43,55 @@ class StringPropertyIndexer {
     bool _initialised {false};
 
     void initialiseIndexTrie(PropertyTypeID propertyID);
+
+    template <TypedInternalID IDT>
     void addStringPropertyToIndex(
         PropertyTypeID propertyID,
-        const TypedPropertyContainer<types::String>& stringPropertyContainer);
+        const TypedPropertyContainer<types::String>& stringPropertyContainer,
+        const std::unordered_map<IDT, IDT>& tempIDMap);
```
rjb32: For the unordered_map use an already defined type or use a type alias to inform about the purpose of the data structure

storage/indexes/StringIndex.h
```diff
@@ -48,7 +49,16 @@ class StringIndex {
         static PrefixTreeNode* create(StringIndex& idx);
 
         static inline size_t charToIndex(char c);
-        static inline char indexToChar(size_t idx);
+
+        static inline char indexToChar(size_t idx) {
+            if (idx > ALPHABET_SIZE) [[unlikely]] {
+                throw TuringException("Invalid index: " + std::to_string(idx));
+            }
+            char c = idx < NUM_ALPHABETICAL_CHARS
```
rjb32: c should be const

PR #24, opened 2025-09-04
Fix chunking in planCreate, add new test (!plan-create-chunking)

cyrusknopf:
Fixing https://turingbio.atlassian.net/browse/TURING-796?atlOrigin=eyJpIjoiODk3YzFjOTE4YzBiNGNkZjhmNTEzMTQ2ZDUzMWRmYjkiLCJwIjoiaiJ9

### DIFF
--- a/plan/QueryPlanner.cpp
+++ b/plan/QueryPlanner.cpp
@@ -6,7 +6,6 @@
 #include <range/v3/view/chunk.hpp>
 #include <range/v3/all.hpp>
 
-#include "BioAssert.h"
 #include "DataPart.h"
 #include "Expr.h"
 #include "FilterStep.h"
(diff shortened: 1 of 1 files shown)


plan/QueryPlanner.cpp
```diff
@@ -206,11 +206,10 @@ bool QueryPlanner::planCreate(const CreateCommand* createCmd) {
             continue;
         }
 
-        for (auto step : pathElements | rv::chunk(3)) {
+        for (auto step : pathElements | rv::drop(1) | rv::chunk(2)) {
             // Create the target + the edge (the source is already created)
-            const auto* src = step[0];
-            const auto* edge = step[1];
-            const auto* tgt = step[2];
+            const auto* edge = step[0];
```
rjb32: Please add the types of edge, tgt, edgeDecl and tgtDecl here so that it's more readable

rjb32: And also for auto line 179, src and srcDecl line 188-189

PR #26, opened 2025-09-04
Add support for injecting existing NodeIDs in CREATE queries

cyrusknopf:
Motivated by [this Jira](https://turingbio.atlassian.net/browse/TURING-704?atlOrigin=eyJpIjoiZTkwMDAxNjY4MmFlNDYyYWIzNjE1YjkyYzM2YzM1ZTAiLCJwIjoiaiJ9).

### DIFF
--- a/pipeline/operations/CreateNodeStep.cpp
+++ b/pipeline/operations/CreateNodeStep.cpp
@@ -2,7 +2,6 @@
 #include "ExecutionContext.h"
 #include "Expr.h"
 #include "ExprConstraint.h"
-#include "InjectedIDs.h"
 #include "PathPattern.h"
 #include "Profiler.h"
 #include "TypeConstraint.h"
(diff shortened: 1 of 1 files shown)


interpreter/QueryAnalyzer.cpp
```diff
@@ -359,21 +367,44 @@ void QueryAnalyzer::analyzeBinExprConstraint(const BinExpr* binExpr,
...
+void QueryAnalyzer::analyzeEntityPattern(DeclContext* declContext, EntityPattern* entity,
+                                         bool isCreate) {
     VarExpr* var = entity->getVar();
     // Handle the case where the entity is unlabeled edge (--)
     if (!var) {
         var = VarExpr::create(_ctxt, createVarName());
         entity->setVar(var);
     }
 
+    uint64_t idToSet = entity->getEntityID();
```
rjb32, line 379: idToSet should be const. Don't we have a better type for IDs?

cyrusknopf: idToSet is later reassigned, so cannot be const. getEntityID() returns uint64_t, I can change idToSet to be EntityID, but later uses of it then require a call to .getValue(). What do you think is better?

PR #32, opened 2025-09-10
Add S3 Connect, Pull, and Push Commands to the query language.

ghost:
The syntax for the commands:

S3_CONNECT <ACCESS_ID> <SECRET_KEY> <AWS_REGION>
S3_PULL <S3_URL> <LOCAL_DIR>
S3_PUSH <LOCAL_DIR> <S3_URL>

### DIFF
--- a/io/filecache/FileCache.h
+++ b/io/filecache/FileCache.h
@@ -15,7 +15,7 @@ class FileCache {
 public:
     FileCache(const fs::Path& graphDir,
               const fs::Path& dataDir,
-              ClientType&& clientWrapper);
+              S3::TuringS3Client<ClientType>& clientWrapper);
 
     FileCacheResult<void> saveGraph(std::string_view graphName);
     FileCacheResult<void> loadGraph(std::string_view graphName);
@@ -43,7 +43,7 @@ public:
 private:
     fs::Path _graphsDir;
     fs::Path _dataDir;
-    S3::TuringS3Client<ClientType> _s3Client;
+    S3::TuringS3Client<ClientType>& _s3Client;
     std::string _userId = "turingDefault";
     std::string _bucketName = "turing-disk-test";
 };
--- a/io/s3/AwsManager.h
+++ b/io/s3/AwsManager.h
@@ -1,21 +1,12 @@
-#include <aws/core/Aws.h>
+#include <memory>
+
+namespace Aws {
+class SDKOptions;
+}
+
+namespace S3 {
 
 class AWSManager {
-private:
-    Aws::SDKOptions _options;
-
-    AWSManager() {
-        _options.loggingOptions.logLevel = Aws::Utils::Logging::LogLevel::Trace;
-        _options.httpOptions.initAndCleanupCurl = true;
-        _options.httpOptions.installSigPipeHandler = true; // Let AWS handle SIGPIPE
-
-        Aws::InitAPI(_options);
-    }
-
-    ~AWSManager() {
-        Aws::ShutdownAPI(_options);
-    }
-
 public:
     static AWSManager& getInstance() {
         static AWSManager instance; // Only initialised once
@@ -26,4 +17,12 @@ public:
     AWSManager& operator=(const AWSManager&) = delete;
     AWSManager(AWSManager&&) = delete;
     AWSManager& operator=(AWSManager&&) = delete;
+
+private:
+    std::unique_ptr<Aws::SDKOptions> _options;
+
+    AWSManager();
+    ~AWSManager();
 };
+
+}
--- a/pipeline/operations/S3ConnectStep.cpp
+++ b/pipeline/operations/S3ConnectStep.cpp
@@ -5,8 +5,6 @@
 #include "ExecutionContext.h"
 #include "Profiler.h"
 #include "SystemManager.h"
-#include "AwsS3ClientWrapper.h"
-
 
 using namespace db;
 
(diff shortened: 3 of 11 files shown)


interpreter/QueryAnalyzer.cpp
```diff
@@ -452,3 +454,35 @@ void QueryAnalyzer::analyzeLoadGraph(LoadGraphCommand* cmd) {
     }
 }
 
+void parseS3Url(std::string_view s3URL, std::string_view& bucket, std::string_view& prefix, std::string_view& fileName) {
```
rjb32, line 457: Put the function in an anonymous namespace at the top of the file if it is purely defined in the cpp file

io/filecache/FileCache.h
```diff
@@ -15,7 +15,7 @@ class FileCache {
 public:
     FileCache(const fs::Path& graphDir,
               const fs::Path& dataDir,
-              ClientType& clientWrapper);
+              ClientType&& clientWrapper);
```
rjb32: Why change from a reference to an rvalue reference?

io/s3/AwsManager.h
```diff
@@ -0,0 +1,29 @@
+#include <aws/core/Aws.h>
```
rjb32: It's best to hide aws includes in the cpp file so that we also don't polute the consumers of this header with aws stuff

io/s3/AwsS3ClientWrapper.h
```diff
@@ -9,6 +9,8 @@
 #include <aws/s3-crt/model/ListObjectsV2Request.h>
 #include <aws/core/Aws.h>
 
+#include "AwsManager.h"
+
 namespace S3 {
 
 template <typename ClientType = Aws::S3Crt::S3CrtClient>
```
rjb32, line 16: I am not sure there is really a value to have a templated implementation here. If you don't template you can hide more stuff in a cpp file

pipeline/operations/S3ConnectStep.cpp
```diff
@@ -0,0 +1,37 @@
+#include "S3ConnectStep.h"
+
+#include <sstream>
+
+#include "ExecutionContext.h"
+#include "Profiler.h"
+#include "SystemManager.h"
+#include "AwsS3ClientWrapper.h"
+
+
```
rjb32, line 8: Too many blank lines

pipeline/operations/S3PullStep.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <sstream>
+
+#include "ExecutionContext.h"
+#include "PipelineException.h"
+#include "Profiler.h"
+#include "TuringS3Client.h"
+#include "SystemManager.h"
+#include "spdlog/spdlog.h"
+
+
```
rjb32: Too many blank lines

pipeline/operations/S3PullStep.cpp
```diff
@@ -0,0 +1,54 @@
...
+    _s3Client = _sysMan->getS3Client();
+    if (_s3Client == nullptr) {
+        throw PipelineException("TuringS3Client has not been created");
+    }
+}
+
+void S3PullStep::execute() {
+    Profile profile {"S3PullStep::execute"};
+    if (!_s3File.empty()) {
+        auto awsRes = _s3Client->downloadFile(_localPath, std::string(_s3Bucket), std::string(_s3File));
```
rjb32, line 36: Why do you recreate string objects to pass to the function?

pipeline/operations/S3PullStep.cpp
```diff
@@ -0,0 +1,54 @@
...
+    _sysMan = ctxt->getSystemManager();
+    _s3Client = _sysMan->getS3Client();
+    if (_s3Client == nullptr) {
+        throw PipelineException("TuringS3Client has not been created");
+    }
+}
+
+void S3PullStep::execute() {
+    Profile profile {"S3PullStep::execute"};
+    if (!_s3File.empty()) {
```
rjb32, line 35: Why this condition determines if it's a directory or a file?

system/SystemManager.h
```diff
@@ -65,10 +67,21 @@ class SystemManager {
...
+    void setS3Client(std::unique_ptr<S3::TuringS3Client<S3::AwsS3ClientWrapper<>>> newClient) {
+        // Move assignment
+        _s3Client = std::move(newClient);
+    }
+
+    S3::TuringS3Client<S3::AwsS3ClientWrapper<>>* getS3Client() {
+        return _s3Client.get();
+    }
+
+
```
rjb32, line 79: Too many blank lines

system/SystemManager.h
```diff
@@ -65,10 +67,21 @@ class SystemManager {
                                               CommitHash commitHash,
                                               ChangeID changeID);
 
+    void setS3Client(std::unique_ptr<S3::TuringS3Client<S3::AwsS3ClientWrapper<>>> newClient) {
```
rjb32: Who sets the client? Is it the best idea to expect this to be set from the outside instead of being created by SystemManager itself?

io/s3/AwsS3ClientWrapper.h
ghost: It's templated mainly for the mock client, the alternative would be to have inherited classes which feels weird having unnecessary Vtable lookup overhead just because of testing. (but I guess performance isn't that big of a concern here?)

pipeline/operations/S3PullStep.cpp
ghost: These are string_views on the S3URL string, and Aws::string doesn't cast string_views well.

io/s3/AwsS3ClientWrapper.h
rjb32: Ah right indeed, I forgot. Yes indeed sounds good

PR #34, opened 2025-09-15
Improve Graph Loading

ghost:
Pass Graph Names As String Constants, Load Neo4j Archvies, Add Ability To Name Loaded Graphs Differently.

### DIFF
--- a/import/neo4j/Neo4jImporter.cpp
+++ b/import/neo4j/Neo4jImporter.cpp
@@ -384,7 +384,7 @@ bool Neo4jImporter::fromDumpFileToJsonDir(JobSystem& jobSystem,
                                           std::size_t nodeCountPerQuery,
                                           std::size_t edgeCountPerQuery,
                                           const DumpFileToJsonDirArgs& args) {
-    std::lock_guard<std::mutex> lock(_mutex);
+    std::unique_lock<std::mutex> lock(_mutex);
 
     Neo4jInstance instance(args._workDir);
     if (Neo4jInstance::isRunning()) {


query/parser/parser.y
```diff
@@ -527,7 +527,8 @@ list_graph_cmd: LIST GRAPH { $$ = ListGraphCommand::create(ctxt); }
            ;
 
 // LOAD GRAPH
-load_graph_cmd: LOAD GRAPH ID { $$ = LoadGraphCommand::create(ctxt, $3); }
+load_graph_cmd: LOAD GRAPH STRING_CONSTANT { $$ = LoadGraphCommand::create(ctxt, $3); }
+           |    LOAD GRAPH STRING_CONSTANT STRING_CONSTANT { $$ = LoadGraphCommand::create(ctxt, $3, $4); }
```
ghost, line 540: I am not sure if we want the graph name to be a STRING_CONSTANT or an ID type

rjb32: I think it may be good for now to be an ID. Just also for considerations of directory names for dumping.

CMakeLists.txt
```diff
@@ -151,6 +151,7 @@ if(APPLE)
     message(STATUS "Using Homebrew Bison include directory: ${BISON_INCLUDE_DIRS}")
 endif()
 
+find_package(ZLIB)
```
rjb32: Add a blank line after

interpreter/QueryAnalyzer.cpp
```diff
@@ -437,18 +437,18 @@ std::string QueryAnalyzer::createVarName() {
...
 
 void QueryAnalyzer::analyzeLoadGraph(LoadGraphCommand* cmd) {
-    const std::string& name = cmd->getName();
+    const std::string& name = cmd->getFileName();
     if (name.empty()) {
-        throw AnalyzeException("No graph name provided.");
+        throw AnalyzeException("No graph file name provided.");
     }
 
-    // Check that the graph name is only [A-Z0-9_]+
```
rjb32, line 447: I think this was a good idea actually

query/parser/parser.y
rjb32: Actually regarding LOAD_GRAPH, we do have this jira https://turingbio.atlassian.net/issues/?filter=10009&selectedIssue=TURING-795 which requires strings for gml files

LucLabarriere: Makes sense to use a string yes. Also in parser v2, we can make it an expression so that it can take string literals, ids or even variables

interpreter/QueryAnalyzer.cpp
LucLabarriere: We need to allow "." then right? for blabla.gml

LucLabarriere:
@rjb32 @suhassn-turing-bio

Seems good to me, do you think in parser v2 we could use the CALL syntax? This would allow to have a cleaner syntax for arguments, possibly even with named arguments:

```
CALL loadGraph("reactome", type="NEO4j")
```

And even put those lines inside scripts, stuff like this
```
# Load additional graph storing the information on APOE-4
MATCH (n:Protein { name= "APOE-4" })
WITH DISTINCT n.database AS db
CALL loadGraph(db)
```

CMakeLists.txt
LucLabarriere: By the way, is this necessary?

interpreter/QueryAnalyzer.cpp
ghost: Do we not assume that files can all sorts of characters?

ghost:
> @rjb32 @suhassn-turing-bio
> 
> Seems good to me, do you think in parser v2 we could use the CALL syntax? This would allow to have a cleaner syntax for arguments, possibly even with named arguments:
> 
> ```
> CALL loadGraph("reactome", type="NEO4j")
> ```
> 
> And even put those lines inside scripts, stuff like this
> 
> ```
> # Load additional graph storing the information on APOE-4
> MATCH (n:Protein { name= "APOE-4" })
> WITH DISTINCT n.database AS db
> CALL loadGraph(db)
> ```

I do like the idea, it would work well if we had connector objects to other data sources like in tiger graph : [https://docs.tigergraph.com/tigergraph-server/3.6/data-loading/s3-loader-user-guide]

.clang-format
LucLabarriere, line 1: You added the clang-format and clangd files? Did you discuss this with Remy? I think would be nice having them but maybe it's weird if we don't all use them in the company

load_env.sh
LucLabarriere, line 1: What does this do? Is it the introduction of a an env var system for turingdb?

pipeline/PipelineStep.cpp
```diff
@@ -230,9 +230,15 @@ PipelineStep::PipelineStep(CallLabelSetStep::Tag,
...
+PipelineStep::PipelineStep(LoadGraphStep::Tag, const std::string& graphFileName)
     : _opcode(PipelineOpcode::LOAD_GRAPH),
-    _impl(std::in_place_type<LoadGraphStep>, graphName)
+    _impl(std::in_place_type<LoadGraphStep>, graphFileName)
+{
+}
+
+PipelineStep::PipelineStep(LoadGraphStep::Tag, const std::string& graphFileName, const std::string& graphName)
+    : _opcode(PipelineOpcode::LOAD_GRAPH),
+    _impl(std::in_place_type<LoadGraphStep>, graphFileName, graphName)
```
LucLabarriere: I've gone through the code a bit but I struggle a bit understand the difference between graphName and graphFileName. If the user loads a graph, how is the graphName generated/retrieved? Do you parse the filepath in system manager then deduce the graph name from it ? (I mean in the case of loading a gml for example). And if we load our own format, do you read the graph name from the class Graph?

unstaged.patch
LucLabarriere, line 1: I assume this was not supposed to be committed?

.clang-format
ghost: oh weird, that shouldn't be the case. Let me fix that.

load_env.sh
ghost: These should not be here at all 😓

unstaged.patch
ghost: yes! I will remove these!

pipeline/PipelineStep.cpp
ghost: so initially we were just parsing the file path to get the graph name :

LOAD GRAPH <GRAPHFILENAME> - which leads to the situation where we might have graphs called data.gml or pole-43.dump

So I added the option: LOAD GRAPH <GRAPHFILENAME> <GRAPHNAME> - so now we can choose the name of a graph we are loading in.

I think renaming these will bring some clarity!

CMakeLists.txt
ghost: @LucLabarriere I needed to add this after changing the aws libraries to statically loaded in.

pipeline/PipelineStep.cpp
LucLabarriere: I think it's honestly! I just needed a bit more clarification

import/neo4j/Neo4jImporter.cpp
```diff
@@ -382,8 +384,9 @@ bool Neo4jImporter::fromDumpFileToJsonDir(JobSystem& jobSystem,
                                           std::size_t nodeCountPerQuery,
                                           std::size_t edgeCountPerQuery,
                                           const DumpFileToJsonDirArgs& args) {
-    Neo4jInstance instance(args._workDir);
+    std::lock_guard<std::mutex> lock(_mutex);
```
rjb32: Please std::unique_lock

interpreter/QueryAnalyzer.cpp
rjb32: Yes indeed

PR #38, opened 2025-09-16
Implement a WriteBuffer for applying CREATE commands at submit-time

cyrusknopf:
This pull requests implements a CommitWriteBuffer system which replaces the existing implementation of using DataPartBuilders at time of CREATE query execution.

Create{Node/Edge}Steps are replaced by a new WriteStep which writes the entire CreateTarget to the CommitWriteBuffer.

At time of Change::submit, the associated commit is synced with main (currently this is just rebasing), and the CommitWriteBuffer is "flushed", which involves filling a DataPartBuilder with the changes stored in the buffer.

### DIFF
--- a/regress/complex_rebases/main.py
+++ b/regress/complex_rebases/main.py
@@ -9,7 +9,7 @@ MAX_EDGES : int = MAX_NODES
 def make_local_changes(client : TuringDB, num_nodes : int, num_edges : int) -> tuple[int, str]:
   ts = int(time.time())
 
-  change = client.query("CHANGE NEW")["Change ID"][0]
+  change = client.query("CHANGE NEW")[0][0]
   client.checkout(change=change)
 
   for _ in range(num_nodes):
@@ -60,8 +60,8 @@ def main():
     expected_nodes += rand_node_counts[change] + (2 * rand_edge_counts[change])
     expected_edges += rand_edge_counts[change]
 
-    num_nodes : int = len(client.query("match (n) return n")["n"])
-    num_edges : int = len(client.query("match (n)-[e]-(m) return e")["e"])
+    num_nodes : int = len(client.query("match (n) return n")[0])
+    num_edges : int = len(client.query("match (n)-[e]-(m) return e")[0])
 
     assert expected_nodes == num_nodes
     assert expected_edges == num_edges


pipeline/operations/WriteStep.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "versioning/CommitBuilder.h"
+
+namespace rg = ranges;
+namespace rv = rg::views;
+
+using namespace db;
+
+namespace {
+
+using CWB = CommitWriteBuffer;
```
rjb32: I am usually not a big fan of this type abbreviations for clarity sake

pipeline/operations/WriteStep.cpp
```diff
@@ -0,0 +1,218 @@
...
+    CWB::UntypedProperties nodeProperties;
+
+    const ExprConstraint* patternProperties = nodePattern->getExprConstraint();
+    if (!patternProperties) { // Early exit if no properties
+        // Add this node to the write buffer, and record its offset
+        CWB::PendingNodeOffset thisNodeOffset =
+            _writeBuffer->nextPendingNodeOffset();
+        const std::string& nodeVarName = nodePattern->getVar()->getName();
+
+        _writeBuffer->addPendingNode(nodeLabels, nodeProperties);
```
rjb32: It looks like duplication with what happens at the main exit of the function. Shouldn't it be encapsulated in some kind of method?

pipeline/operations/WriteStep.h
```diff
@@ -0,0 +1,78 @@
...
+     * @brief Adds a pending node to @ref _writeBuffer specified by @param nodePattern,
+     * and returns the offset at which it can be found in the @ref _pendingNodes vector.
+     */
+    CommitWriteBuffer::PendingNodeOffset writeNode(const EntityPattern* nodePattern);
+
+    /**
+     * @brief Adds a pending edge to @ref _writeBuffer specified by @param edgePattern and
+     * using @ref src and @ref tgt as the source and target nodes.
+     */
+    void writeEdge(const ContingentNode src, const ContingentNode tgt,
```
rjb32: Put each argument on a different line

storage/reader/GraphReader.cpp
```diff
@@ -9,20 +11,30 @@
...
 size_t GraphReader::getNodeCount() const {
     if (_view.dataparts().empty()) {
         return 0;
     }
-    const auto& part = _view.dataparts().back();
-    return part->getFirstNodeID().getValue() + part->getNodeCount();
+
+    // Due to deletions, NodeIDs are no longer guaranteed to be contiguous across
+    // dataparts. Hence we must manually now sum up the nodes in each datapart.
+    return rg::accumulate(_view.dataparts(), 0, [](int acc, const WeakArc<DataPart>& dp) {
```
rjb32: Can we use std algorithms instead of the ranges library to compute the sum?

storage/versioning/Change.cpp
```diff
@@ -85,23 +91,46 @@ CommitResult<void> Change::rebase(JobSystem& jobsystem) {
         _base->commits().back().openTransaction().readGraph().getEdgeCount();
 
     MetadataRebaser metadataRebaser;
-    DataPartRebaser dataPartRebaser(oldNextNodeID, oldNextEdgeID, newNextNodeID,
+    DataPartRebaser dataPartRebaser(branchTimeNextNodeID, branchTimeNextEdgeID, newNextNodeID,
```
rjb32, line 91: Thanks, this is clearer

storage/versioning/CommitBuilder.cpp
```diff
@@ -1,18 +1,24 @@
 #include "CommitBuilder.h"
 
+#include <unordered_map>
+#include <range/v3/view/enumerate.hpp>
+
 #include "reader/GraphReader.h"
 #include "Profiler.h"
 #include "Graph.h"
+#include "spdlog/spdlog.h"
```
rjb32: Put this as a library include below the range include

storage/versioning/CommitBuilder.cpp
```diff
@@ -84,6 +95,42 @@ CommitResult<std::unique_ptr<Commit>> CommitBuilder::build(JobSystem& jobsystem)
...
+    // If there is nothing to flush, return early without creating a new builder
+    if (wb.empty()) {
+        return;
+    }
+
+    // TODO: Needs to be a better warning mode here
+    // Ensure this CommitBuilder is in a fresh state, otherwise there are builders which
+    // have used a potentially outdated metadata
+    if (!isEmpty() || pendingCount() != 0) {
+        spdlog::warn(
```
rjb32: It should probably be an exception

storage/versioning/CommitBuilder.cpp
```diff
@@ -84,6 +95,42 @@ CommitResult<std::unique_ptr<Commit>> CommitBuilder::build(JobSystem& jobsystem)
...
+            "CommitBuilder to be flushed is not empty; there are {} pending builders and {} datapartcount",
+            pendingCount(), dpCount());
+    }
+
+    // We create a single datapart when flushing the buffer, to ensure it is synced with
+    // the metadata provided when rebasing main
+    DataPartBuilder& dpBuilder = newBuilder();
+
```
rjb32: Remove blank line

storage/versioning/CommitBuilder.cpp
```diff
@@ -107,6 +162,9 @@ void CommitBuilder::initialize() {
     // Create metadata builder
     _metadataBuilder = MetadataBuilder::create(_view.metadata(), &_commitData->_metadata);
 
+    // Create the write buffer for this commit
+    _writeBuffer = std::make_unique<CommitWriteBuffer>();
+
     // Create datapart builder
-    this->newBuilder();
+    // this->newBuilder(); XXX: Do not do this here, it need be created at submit time
```
rjb32: What about this?

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,119 @@
...
+     using ContingentNode = std::variant<NodeID, PendingNodeOffset>;
+ private:
+     struct PendingEdge {
+         ContingentNode src;
+         ContingentNode tgt;
+         std::string edgeLabelTypeName;
+         UntypedProperties properties;
+    };
+
+public:
```
rjb32: Why is there multiple public/private sections?

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,119 @@
...
+     struct PendingNode {
+         std::vector<std::string> labelNames;
+         UntypedProperties properties;
+     };
+
+     using PendingNodeOffset = size_t;
+
+     // A node: either exists in previous commit (materialised as NodeID),
+     // or to be created in this commit (materialised as PendingNode)
+     using ContingentNode = std::variant<NodeID, PendingNodeOffset>;
```
rjb32: Why is it called a ContigentNode?

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,119 @@
...
+
+    PendingNodeOffset nextPendingNodeOffset() { return _pendingNodes.size(); }
+
+    void addPendingNode(std::vector<std::string>& labels,
+                        UntypedProperties& properties);
+
+    void addPendingEdge(ContingentNode src, ContingentNode tgt, std::string& edgeType,
+                        UntypedProperties& edgeProperties);
+
+    std::vector<PendingNode>& pendingNodes() { return _pendingNodes; }
```
rjb32: Use a type alias for pending nodes

test/query/QueryTest.cpp
```diff
@@ -622,6 +628,78 @@ TEST_F(QueryTest, ChangeWithRebaseFromEmpty) {
         .execute();
 }
 
+TEST_F(QueryTest, threeCreateTargets) {
```
rjb32, line 627: I think we need to add scenarios where main has progressed more between the base commit of the change and submit time into main

rjb32:
I think overall it's quite clear at a high level, also from a code style point of view. Thank you for commenting well as well.

storage/versioning/CommitWriteBuffer.h
cyrusknopf: Change name

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,119 @@
...
+
+     struct UntypedProperty {
+        std::string propertyName;
+        SupportedTypeVariant value;
+     };
+
+     using UntypedProperties = std::vector<UntypedProperty>;
+
+     struct PendingNode {
+         std::vector<std::string> labelNames;
```
rjb32: Should we use a labelset?

pipeline/operations/WriteStep.h
```diff
@@ -0,0 +1,78 @@
...
+    CommitWriteBuffer* _writeBuffer {nullptr};
+
+    /**
+     * @brief Mapping from variable names to offsets in @ref _writeBuffer 's @ref
+     * _pendingNodes vector.
+     * @detail Allows nodes which may be referred to by edges in the same or a different
+     * CreateTarget to be correctly referred to when creating PendingEdges in the
+     * CommitWriteBuffer.
+     */
+    std::unordered_map<std::string, CommitWriteBuffer::PendingNodeOffset> _varOffsetMap;
```
rjb32: We should use VarDecl* as a key here instead of string because hashing a VarDecl* is cheaper than hashing strings

storage/versioning/CommitWriteBuffer.h
cyrusknopf: This type is meant to represent a node which either will be created in this commit, or already exists in a previous commit.

Do you think "PotentialNode" or "OptionalNode" is clearer? Or something else?

test/query/QueryTest.cpp
cyrusknopf: I could do this in a Python regression test? Seems high-level enough to warrant it?

storage/reader/GraphReader.cpp
cyrusknopf: Sure, is that in case we want to remove ranges-v3 as a dependency? Or is there another reason you prefer STL > ranges-v3?

rjb32: Yes I would like to not increase our dependency on ranges-v3. And the stl now has a bunch of convenient algorithms that we can use without relying on an external dependency

test/query/QueryTest.cpp
rjb32: Yes a regress is good

pipeline/operations/WriteStep.cpp
cyrusknopf: Removed

pipeline/operations/WriteStep.cpp
cyrusknopf: I've adjusted  the control flow so there isn't this repetition

pipeline/operations/WriteStep.h
cyrusknopf: Done

storage/reader/GraphReader.cpp
cyrusknopf: Sounds good, done

pipeline/operations/WriteStep.h
cyrusknopf: Agree, done

storage/versioning/CommitWriteBuffer.h
cyrusknopf: Done

storage/versioning/CommitBuilder.cpp
cyrusknopf: Fixed, spdlog no longer needed

storage/versioning/CommitBuilder.cpp
cyrusknopf: Removed this check as it was for debugging purposes

storage/versioning/CommitBuilder.cpp
cyrusknopf: Left this for the sake of the code review, removed it now

storage/versioning/CommitWriteBuffer.h
cyrusknopf: Fixed the structure of the class sections to use 1 public section and 1 private section

storage/versioning/Change.cpp
```diff
@@ -70,9 +76,9 @@ CommitResult<void> Change::rebase(JobSystem& jobsystem) {
     Profile profile {"Change::rebase"};
 
     // Get the next Edge and Node IDs at the time this change branched
-    NodeID oldNextNodeID =
+    NodeID branchTimeNextNodeID =
         _base->commits().back().openTransaction().readGraph().getNodeCount();
```
LucLabarriere: Store the transaction and GraphReader once

storage/versioning/CommitWriteBuffer.h
rjb32: ExistingOrPendingNode

pipeline/operations/WriteStep.cpp
```diff
@@ -0,0 +1,206 @@
...
+CommitWriteBuffer::PendingNodeOffset WriteStep::writeNode(const EntityPattern* nodePattern) {
+    // Build a @ref CommitWriteBuffer::PendingNode:
+
+    // Labels to pass to the PendingNode
+    const TypeConstraint* patternLabels = nodePattern->getTypeConstraint();
+    std::vector<std::string> nodeLabels;
+    if (patternLabels) {
+        nodeLabels.reserve(patternLabels->getTypeNames().size());
+        for (const VarExpr* name : patternLabels->getTypeNames()) {
+            nodeLabels.emplace_back(name->getName());
```
LucLabarriere: Since the WriteBuffer is tied to a commit, we should be able to retrieve the LabelMap of the current commit and use that to only store labelIDs. However, I assume that means that le label IDs have to updated during the rebase so it's not an easy fix. I let you guys decide if we should do it for this PR or not @cyrusknopf @rjb32

Later on we could even have a LabelSet instead of a vector to avoid the heap allocation and pointer indirection

storage/versioning/CommitWriteBuffer.h
LucLabarriere: It would definitely be better, no pointer indirection, no heap allocation. Only problem is that the rebase needs to update the pending nodes with the fixed label IDs

storage/versioning/CommitWriteBuffer.h
cyrusknopf: Done

storage/versioning/CommitBuilder.cpp
cyrusknopf: Done

storage/writers/DataPartBuilder.h
```diff
@@ -1,12 +1,14 @@
...
 
 #include <memory>
+#include <unordered_map>
 #include <unordered_set>
 
 #include "EdgeRecord.h"
 #include "ID.h"
 #include "properties/PropertyManager.h"
 #include "views/GraphView.h"
+#include "versioning/CommitWriteBuffer.h"
```
LucLabarriere: We should probably forward declare CommitWriteBuffer. Although this might need to extract PendingNode from the CommitWriteBuffer
But to be fair, I don't think the DataPartBuilder need to know anything about the CommitWriteBuffer

storage/writers/DataPartBuilder.cpp
```diff
@@ -102,6 +124,120 @@ const EdgeRecord& DataPartBuilder::addEdge(EdgeTypeID typeID, NodeID srcID, Node
     return edge;
 }
 
+NodeID DataPartBuilder::addPendingNode(CommitWriteBuffer::PendingNode& node) {
```
LucLabarriere: Ok so I really think it's weird that DataPartBuilder now has dependencies to the CommitWriteBuffer. If we want to keep the DataPartBuilder in this PR, we need the WriteBuffer to use a DataPartBuilder to generate a datapart. But here it looks like the DataPartBuilder somehow uses the CommitWriteBuffer. I'd have assumed we would have this kind of API:
```
auto& partBuilder = commitBuilder.getPartBuilder();
wb.generateDataPartBuilder(partBuilder);
```

And then this whole logic goes into the CommitWriteBuffer (or an intermediate class between the DataPartBuilder and the WriteBuffer)

Here it's really weird that we do `partBuilder->addPendingEdge()` and that in `addPendingEdge()` we check if the node was deleted and if so just return null. Instead it should the WriteBuffer that checks if the src/target is deleted, then called `partBuilder->addEdge()`

storage/writers/DataPartBuilder.cpp
```diff
@@ -102,6 +124,120 @@ const EdgeRecord& DataPartBuilder::addEdge(EdgeTypeID typeID, NodeID srcID, Node
...
+                using T = std::decay_t<decltype(propValue)>;
+
+                if constexpr (std::is_same_v<T, types::Int64::Primitive>) {
+                    this->addNodeProperty<types::Int64>(nodeID, propID, propValue);
+                } else if constexpr (std::is_same_v<T, types::UInt64::Primitive>) {
+                    this->addNodeProperty<types::UInt64>(nodeID, propID, propValue);
+                } else if constexpr (std::is_same_v<T, types::Double::Primitive>) {
+                    this->addNodeProperty<types::Double>(nodeID, propID, propValue);
+                } else if constexpr (std::is_same_v<T, std::string>) {
+                    this->addNodeProperty<types::String>(nodeID, propID, propValue);
```
LucLabarriere: I'm not sure that the condition is valid `std::is_same_v<T, std::string>`. Shouldn't it be `std::is_same_v<T, types::String::Primitive>` ? If not then I assume we need to forward the propValue: `std::forward<std::string>(propValue)`

storage/writers/DataPartBuilder.cpp
```diff
@@ -102,6 +124,120 @@ const EdgeRecord& DataPartBuilder::addEdge(EdgeTypeID typeID, NodeID srcID, Node
...
+}
+
+std::optional<EdgeID> DataPartBuilder::addPendingEdge(const CommitWriteBuffer& wb,
+                                                      const CommitWriteBuffer::PendingEdge& edge,
+                                                      const std::unordered_map<CommitWriteBuffer::PendingNodeOffset, NodeID>& tempNodeIDMap) {
+    using CWB = CommitWriteBuffer;
+    // If this edge has source or target which is a node in a previous datapart, check
+    // if it has been deleted. NOTE: Deletes currently not implemented
+    if (std::holds_alternative<NodeID>(edge.src)) {
+        const NodeID srcID = std::get<NodeID>(edge.src);
```
LucLabarriere: You can contract that into something like that: 
```c++
if (const NodeID* srcID = std::get_if<NodeID>(edge.src)) {
    // use srcID
}
```

storage/writers/DataPartBuilder.cpp
```diff
@@ -102,6 +124,120 @@ const EdgeRecord& DataPartBuilder::addEdge(EdgeTypeID typeID, NodeID srcID, Node
...
+                        {newEdgeID, srcID, tgtID, edgeTypeID}, propID, propValue);
+                } else if constexpr (std::is_same_v<T, std::string>) {
+                    this->addEdgeProperty<types::String>(
+                        {newEdgeID, srcID, tgtID, edgeTypeID}, propID, propValue);
+                } else if constexpr (std::is_same_v<T, types::Bool::Primitive>) {
+                    this->addEdgeProperty<types::Bool>(
+                        {newEdgeID, srcID, tgtID, edgeTypeID}, propID, propValue);
+                }
+            },
+            prop.value);
```
LucLabarriere: Yeah you see, if we adopt a more Data oriented design, the properties are decoupled from the entities that own them. So you have all the names, then all the ages, etc and you can do the type check once for the whole collection. Moreover for updates it will be easier to just remove or modify existing properties. But that might require a little bit too much work for this PR

storage/versioning/VersionController.cpp
```diff
@@ -62,25 +66,24 @@ CommitResult<void> VersionController::submitChange(Change* change, JobSystem& jo
 
     std::scoped_lock lock(_mutex);
 
-    const auto& head = _head.load();
-    if (head->hash() != change->baseHash()) {
+    // atomic load main
+    Commit* mainState = _head.load();
```
LucLabarriere, line 70: @rjb32 @cyrusknopf
We have naming issues tbh. We're trying to mimick git for certain things (commits), but stay away from it for others (changes instead of branches). I think we need to settle on our definitions, maybe write a notion page an accept it once and for all. When I called that thing "head" I had in mind the HEAD of git, but it's not the same thing anyway. I still think it's a good name because it points to a very specify commit which is the tip of main (and I don't like mainTip).

In my opinion, we either try to mimick 100% git and rename Change into Branch and try to find a better name for head. Or we accept that our definitions are similar but not the same as git's and we accept something like "head" for this (or find a better name).

There was also the issue for commitBuilders, we can probably come up with a better name (commitBuilder is a wrapper around a commit, so in fact a better name would be PendingCommit).

In my opinion the CommitWriteBuffer also needs a better name because it sounds way more important than just a buffer. It's literaly the information necessary to build the commit in a way

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,123 @@
...
+
+class CommitWriteBuffer {
+
+struct PendingEdge;
+
+public:
+    using SupportedTypeVariant =
+        std::variant<types::Int64::Primitive, types::UInt64::Primitive,
+                     types::Double::Primitive, std::string,
+                     types::Bool::Primitive>;
```
LucLabarriere, line 27: Are you sure this shouldn't be types::String::Primitive instead of std::string? I don't remmember exactly at which point we store the string in a string buffer and can reference it as std::string_view

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,123 @@
...
+
+     struct UntypedProperty {
+        std::string propertyName;
+        SupportedTypeVariant value;
+     };
+
+     using UntypedProperties = std::vector<UntypedProperty>;
```
LucLabarriere, line 34: Yeah that sounds extrememely expensive. For each node, we have a vector of std::string for labels + a vector of {std::string + prop value (which maybe strings)} for properties. The memory is broken into tiny pieces

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,123 @@
...
+
+     void addPendingNode(std::vector<std::string>& labels, UntypedProperties& properties);
+
+     void addPendingEdge(ExistingOrPendingNode src, ExistingOrPendingNode tgt, std::string& edgeType,
+                         UntypedProperties& edgeProperties);
+
+     PendingNodes& pendingNodes() { return _pendingNodes; }
+     PendingEdges& pendingEdges() { return _pendingEdges; }
+
+     const std::set<NodeID> deletedNodes() const { return _deletedNodes; }
```
LucLabarriere: You wanted to return a const ref?

storage/versioning/CommitWriteBuffer.cpp
```diff
@@ -0,0 +1,63 @@
+#include "CommitWriteBuffer.h"
+#include "ID.h"
+#include <variant>
+
+using namespace db;
+
+void CommitWriteBuffer::addPendingNode(std::vector<std::string>& labels,
+                                       std::vector<UntypedProperty>& properties) {
+    _pendingNodes.emplace_back(labels, properties);
```
LucLabarriere: You're making a copy here, you need to use std::move()

storage/versioning/CommitWriteBuffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void CommitWriteBuffer::addPendingEdge(ExistingOrPendingNode src, ExistingOrPendingNode tgt,
+                                       std::string& edgeType,
+                                       std::vector<UntypedProperty>& edgeProperties) {
+    _pendingEdges.emplace_back(src, tgt, edgeType, edgeProperties);
```
LucLabarriere: std::move()

storage/versioning/CommitWriteBuffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            return wbID;
+        }
+    };
+
+    for (CommitWriteBuffer::PendingEdge& e : _buffer.pendingEdges()) {
+        // We only care about edges that refer to NodeIDs
+        if (std::holds_alternative<NodeID>(e.src)) {
+            NodeID oldSrcID = std::get<NodeID>(e.src);
+            e.src = NodeID {rebaseNodeID(oldSrcID)};
```
LucLabarriere: use if (const NodeID* oldSrcId = std::get_if<NodeID>(e.src)) {...}

storage/versioning/CommitBuilder.cpp
```diff
@@ -55,10 +60,16 @@ CommitResult<void> CommitBuilder::buildAllPending(JobSystem& jobsystem) {
...
+        // Change::commit we do not want to perform this sync, and can continue with our
+        // local next ID.
+        auto part = _controller->createDataPart(_nextNodeID, _nextEdgeID);
 
-        _firstNodeID += builder->nodeCount();
-        _firstEdgeID += builder->edgeCount();
+        // Update these values so the next builder which is created starts where the last
+        // left off
+        _nextNodeID += builder->nodeCount();
+        _nextEdgeID += builder->edgeCount();
```
LucLabarriere, line 68: How do you make the distinction that you mention in the comments?

storage/versioning/CommitBuilder.cpp
```diff
@@ -84,6 +94,32 @@ CommitResult<std::unique_ptr<Commit>> CommitBuilder::build(JobSystem& jobsystem)
     return std::move(_commit);
 }
 
+void CommitBuilder::flushWriteBuffer(JobSystem& jobsystem) {
+    CommitWriteBuffer& wb = writeBuffer();
```
LucLabarriere, line 94: Double check that multithreading does not cause issues here. Everything was properly wrapped in scoped_locks, but not anymore so we need to double check. What happens if you call flush + at the same time people write to the buffer ? Maybe that's not possible because flush is called in protecteed environments but from the point of view of this class, it's not the case

storage/versioning/CommitBuilder.cpp
```diff
@@ -84,6 +94,32 @@ CommitResult<std::unique_ptr<Commit>> CommitBuilder::build(JobSystem& jobsystem)
...
+
+    // TODO: Remove this by using a mapping of NodeID = offset + _firstNodeID
+    std::unordered_map<CommitWriteBuffer::PendingNodeOffset, NodeID> tempIDMap;
+
+    // Add pending nodes to be created to the current DataPartBuilder
+    for (const auto& [offset, node] : wb.pendingNodes() | rv::enumerate) {
+        NodeID nodeID = dpBuilder.addPendingNode(node);
+        tempIDMap[offset] = nodeID;
+    }
+
```
LucLabarriere: Wouldn't it be easy to extract the logic that you had in DataPartBuilder and put it here? And only call addNode/addNodeProperty here, so that DataPartBuilder does not have to be aware of the write buffer

storage/versioning/Change.cpp
```diff
@@ -52,6 +56,8 @@ PendingCommitReadTx Change::openReadTransaction(CommitHash commitHash) {
 CommitResult<void> Change::commit(JobSystem& jobsystem) {
     Profile profile {"Change::commit"};
 
+    _tip->flushWriteBuffer(jobsystem);
+
     if (auto res = _tip->buildAllPending(jobsystem); !res) {
         return res;
     }
```
LucLabarriere, line 61: Could buildAllPending be called in flushWriteBuffer? Or maybe could we have a function that does the flush + the build and call this one? This might help for locks

storage/versioning/DataPartRebaser.h
```diff
@@ -45,6 +45,12 @@ class DataPartRebaser {
     NodeID _targetCommitNextNodeID {0};
     EdgeID _targetCommitNextEdgeID {0};
 
+    NodeID _baseCommitFirstNodeID {0};
+    EdgeID _baseCommitFirstEdgeID {0};
+
+    NodeID _targetCommitFirstNodeID {0};
+    EdgeID _targetCommitFirstEdgeID {0};
```
LucLabarriere: Do you need this?

storage/writers/DataPartBuilder.cpp
cyrusknopf: I agree this pattern is better:
```
auto& partBuilder = commitBuilder.getPartBuilder();
wb.generateDataPartBuilder(partBuilder);
```

I will start refactoring this

storage/writers/DataPartBuilder.cpp
cyrusknopf: String::Primitive is string_view, and in the current implementation the PendingNode cannot store a string_view because it gets deallocated by the owner before we flush (I think the owner is the EntityPattern).

The UntypedProperty variant contains std::string and not String::Primitive for this reason.

addNodeProperty then implicitly casts the std::string to a string_view, which is then forwarded around until the StringPropertyContainer allocates the chars for it in a bucket (if I have understood correctly).

Although this now requires an additional owning std::string in the PendingNode - as opposed to the current implementation where the property is passed to the DataPartBuilder as a view form the EntityPattern - I think it is necessary, as the WriteBuffer flush is at some unknown point in time in the future, certainly after the destruction of the EntityPattern, and therefore the buffer itself needs to own the string.

LucLabarriere: Ok makes perfect sense, then forward the value with std::forward<std::string>(propValue). This is more evidence that the CommitWriteBuffer is gonna have to go through some heavy optimisations because it's at the core of the write mechanism. It's the owner of many things, and has complex logic

storage/versioning/CommitWriteBuffer.cpp
cyrusknopf: I decided to refactor these methods (addPendingEdge/Node) to take rvalue reference to make it explicit to the caller that they must std::move the labels and properties and therefore make it obvious the WriteBuffer is taking ownership

storage/versioning/Change.cpp
cyrusknopf: Done

pipeline/operations/WriteStep.cpp
cyrusknopf: I think LabelIDs is hardly better than string? We would need something like
```
vector<vector<LabelID> _pendingNodesLabels
```
so again we an extra heap allocation and indirection, essentially the same as vec<string>.

I think it would only be worth changing to
```
vector<LabelSet> _pendingNodesLabelSets
```

storage/writers/DataPartBuilder.h
cyrusknopf: Removed the dependency, DataPartBuilder now completely separate from CommitWriteBuffer

storage/writers/DataPartBuilder.cpp
cyrusknopf: Done

storage/writers/DataPartBuilder.cpp
cyrusknopf: What does forwarding do here? And why is it beneficial? propValue comes in as a string& and is then implicitly converted to string_view as a parameter of addNodeProperty.

storage/writers/DataPartBuilder.cpp
cyrusknopf: Done

storage/writers/DataPartBuilder.cpp
cyrusknopf: At the moment I'm imagining an AoS for each property type, for both nodes and edges:

```cpp
struct pendingIntNodeProperty {
  NodeID nodeID;
  PropertyTypeID propID;
  types::Int val; 
};

// UInt property struct
// Double
// ...

vector<pendingIntNodeProperty> _pendingIntProperties
// UInt pending struct array
// ... 
```

because this matches the access pattern best (we need NodeID, Property ID, and value for each property at once to call DataPartBuilder::addNodeProperty).

Is this the kind of thing you had in mind?

storage/versioning/CommitWriteBuffer.cpp
cyrusknopf: Done, see above

storage/versioning/CommitWriteBuffer.cpp
cyrusknopf: Done

storage/versioning/CommitBuilder.cpp
cyrusknopf: Done

storage/versioning/DataPartRebaser.h
cyrusknopf: Nope, good catch, thanks

pipeline/operations/WriteStep.cpp
LucLabarriere: Ok if you think it's too much work, let's keep this way for now, we'll see in a later PR if we can optimize things out

storage/writers/DataPartBuilder.cpp
LucLabarriere: Yes something like that, or split by property type id, but that should be sufficient already (the propID is duplicated accross all properties which might be too much memory, but it maybe fine)

storage/versioning/CommitWriteBuffer.cpp
LucLabarriere: You still need to move the values here, otherwise it makes a copy
```c++
pendingNodes.emplace_back(std::move(labels), std::move(properties))
```

If you have clangd, you can add this clang-tidy check to get diagnostics: cppcoreguidelines-rvalue-reference-param-not-moved

Relevant stackoverflow post: [std::move or rvalue reference parameter](https://stackoverflow.com/questions/58761148/reason-to-use-stdmove-on-rvalue-reference-parameter)

storage/versioning/CommitWriteBuffer.cpp
```diff
@@ -0,0 +1,223 @@
...
+    ValueType operator()(types::Int64::Primitive propValue) const {
+        return types::Int64::_valueType;
+    }
+    ValueType operator()(types::UInt64::Primitive propValue) const {
+        return types::UInt64::_valueType;
+    }
+    ValueType operator()(types::Double::Primitive propValue) const {
+        return types::Double::_valueType;
+    }
+    ValueType operator()(std::string propValue) const {
```
LucLabarriere: const std::string&
With clangd, you can have this check: performance-unnecessary-value-param

storage/writers/DataPartBuilder.cpp
LucLabarriere: Yeah you're right that it shouldn't change anything here, but if we leave like that, there is a change that, one day we change the implementation to pass in a std::string instead, which would turn this into a copy without us realizing it. I assume it's usually a good idea to use std::forward<...>(...) whenever you pass an argument as a template value auto&& or T&& (even for integers):

```c++
template <typename T>
void doSth(T&& arg) {
    anotherFunc(std::forward<T>(arg));
}
```

The std::forward call allows to do perfect forwarding, so that, it automatically detects how it should pass the argument. Right now it's fine because it passes it as an std::string_view but maybe that will change.

For integers here it does not matter because you handle the cases with the if constexpr, but I would still add the forward for the std::string

storage/versioning/VersionController.cpp
rjb32: Yes we will discuss better names soon. I agree for confusion around "head" and the notion of "change". In fact a branch is perfectly acceptable.

storage/versioning/CommitWriteBuffer.cpp
rjb32: Beware of what you do with rvalue references and be always extremely obvious and plain with the way you manage ownership. Please do not rely on extremely subtle ways of doing std::move or not.

storage/versioning/CommitWriteBuffer.cpp
cyrusknopf: Fixed both, thanks

storage/writers/DataPartBuilder.cpp
cyrusknopf: Makes sense, fixed, thanks

storage/versioning/CommitWriteBuffer.cpp
cyrusknopf: @LucLabarriere  thanks my mistake
@rjb32 My thinking is making the function take r value reference is explicit  in the function moving the ownership of the arguments. Is that what you mean? Or do you consider this too subtle?

LucLabarriere: At least, if a function takes an rvalue reference, the program does not compile unless you explicitly move or create a temporary. People don't always do that because it makes the API hard to use if what you actually want to do is a copy. Consider this:
```c++
std::string myString; // Need to keep a version of it in the current scope;

doSth(std::string{myString}); // Only way to use doSth(std::string&&) is to explicitely make a copy
```

But it puts a strong contraint and adds intention to the code so maybe it's what we want?

rjb32: This is exactly what we want. We should never violate what is stated in the interface of a function or class.

storage/versioning/Change.cpp
cyrusknopf: How do you imagine it helping for locking? Presumably the write buffer need be protected by a lock whenever writing to it, and whenever a change calls Commit or Submit the buffer also need be locked.

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,140 @@
...
+
+     struct UntypedProperty {
+        PropertyTypeID propertyID;
+        SupportedTypeVariant value;
+     };
+
+     using UntypedProperties = std::vector<UntypedProperty>;
+
+     struct PendingNode {
+         LabelSetHandle labelsetHandle;
```
cyrusknopf, line 37: @rjb32 @LucLabarriere 
Key change: string -> LabelSetHandle

storage/versioning/CommitWriteBuffer.h
```diff
@@ -0,0 +1,140 @@
...
+    void addDeletedNodes(const std::vector<NodeID>& newDeletedNodes);
+
+private:
+    friend DataPartBuilder;
+    friend CommitWriteBufferRebaser;
+
+    struct PendingEdge {
+         ExistingOrPendingNode src;
+         ExistingOrPendingNode tgt;
+         EdgeTypeID edgeType;
```
cyrusknopf, line 91: @rjb32 @LucLabarriere 
Key change: string -> EdgeTypeID

storage/versioning/MetadataRebaser.cpp
```diff
@@ -104,22 +107,41 @@ bool MetadataRebaser::rebase(const GraphMetadata& theirs, MetadataBuilder& ours)
...
-    if (_edgeTypesChanged) {
-        ours._metadata->_edgeTypeMap = std::move(newEdgeTypes);
+        for (auto& [id, value] : node.properties) {
+             id = _propTypeMapping[id]._id;
+        }
     }
 
-    if (_propTypesChanged) {
-        ours._metadata->_propTypeMap = std::move(newPropTypes);
+    // Rebase WriteBuffer edge metadata
```
cyrusknopf, line 132: @rjb32 @LucLabarriere 
MetadataRebaser now rebases the metadata in the CommitWriteBuffer

storage/versioning/CommitBuilder.cpp
cyrusknopf: @rjb32 How do we want to handle locking here? Considering a CommitWriteBuffer exists within a Commit, which exists within a Change, does the ChangeAccessor only allow one thread to modify the buffer at once? If not, then I will need to handle concurrent writes to the buffer, either through a lock or lock-free structures.

Similarly, as @LucLabarriere mentioned, the flushWriteBuffer is called only within Change::commit or VersionController::submitChange, which are both guarded by either the ChangeAccessor or by a caller which takes the _changesLock.

pipeline/operations/WriteStep.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+}
+
+void WriteStep::prepare(ExecutionContext* ctxt) {
+    Transaction* rawTx = ctxt->getTransaction();
+    if (!rawTx->writingPendingCommit()) {
+        throw PipelineException("WriteStep: Cannot perform writes outside of a write transaction");
+    }
+
+    auto& tx = rawTx->get<PendingCommitWriteTx>();
```
rjb32, line 99: Do not use auto for readability here because it's quite important to see explicitly which kind of transaction

storage/versioning/CommitBuilder.cpp
rjb32: * The WriteStep of the pipeline to add new entities in the buffer makes sure to be in a write transaction.
* When we are in a write transaction, we have a ChangeAccessor alive for the lifetime of this write tx
* Whenever a ChangeAccessor is alive, it holds a unique lock on the change mutex
Thus a WriteStep can only be executed while holding a lock on the change.

Furthermore, commit and submit can only be called through ChangeAccessor methods because it is the only friend class allowed to do so. And ChangeAccessor can only be alive when a unique lock on the change mutex is held because this is done in ChangeAccessor constructor.

So one of them will be blocked in the acquisition of the unique lock of ChangeAccessor, either the write step or commit/submit.

PR #40, opened 2025-09-18
Expr constraints on nodes and edges for plan v2

LucLabarriere:


v2/AST/decl/PatternData.h
```diff
@@ -11,9 +11,10 @@ namespace db::v2 {
 class CypherAST;
 class Expr;
 
+using ExprConstraints = std::vector<std::pair<PropertyType, Expr*>>;
```
rjb32: Make NodePatternData and EdgePatternData derive from the same parent class and put the ExprConstraints type in there, so that you can insulate it a bit

v2/AST/expr/ExprEvaluator.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <cstdint>
```
rjb32: Use stdint.h

v2/interpreter/QueryInterpreterV2.cpp
```diff
@@ -77,7 +77,7 @@ db::QueryStatus QueryInterpreterV2::execute(const InterpreterContext& ctxt,
     }
 
     // Generate plan graph
-    PlanGraphGenerator planGen(view, ctxt.getQueryCallback());
+    PlanGraphGenerator planGen(ast, view, ctxt.getQueryCallback());
     try {
```
rjb32: Why don't we have anything here?

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,404 +163,234 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
-    if (!targetPattern) {
-        throw PlannerException("Target pattern element must be a node pattern");
-    }
-                                                        
+void PlanGraphGenerator::generatePatternElementEdge(const EdgePattern* edge) {
     // Get out edges
     GetOutEdgesNode* getOutEdges = _tree.create<GetOutEdgesNode>();
-    currentNode->connectOut(getOutEdges);
-    currentNode = getOutEdges;
+    _currentNode->connectOut(getOutEdges);
```
rjb32: Is it really useful to add _currentNode as a class member now instead of passing in a functional way?

v2/plan/PropertyMapExpr.h
```diff
@@ -0,0 +1,42 @@
...
+#include "PlanGraph.h"
+
+#include <vector>
+
+namespace db::v2 {
+
+class Expr;
+class PlanGraphNode;
+
+class PropertyMapExpr : public PlanGraphNode {
```
rjb32: This is a PlanGraphNode, its name should end with "Node", something like PropertyMapExprNode or something more clear for its purpose

v2/AST/expr/ExprEvaluator.h
```diff
@@ -0,0 +1,26 @@
...
+#include <vector>
+
+namespace db::v2 {
+
+class ExprEvaluator {
```
rjb32: What is this for?

LucLabarriere: To be removed

v2/plan/PlanGraphGenerator.cpp
LucLabarriere: I think so, it allows better split the generate functions for the origin and edge/target pairs, it could also be put into the argument of the functions, but I think it put more emphasis on the "state machine" behaviour of the plan graph generator. If at some point we end in the case where the _currentNode member function becomes annoying and we actually need to pass that around, it means we deviate from a state machine behaviour and that we need to be careful. I felt it when trying to connect back the VarNode n to a GetProperty (the example I sent you), it really highlights that it is a pathological pattern in the design

v2/interpreter/QueryInterpreterV2.cpp
LucLabarriere: No idea actually ahah, it's your code I think, maybe you didn't want to trigger the planner for now when we use the #v2 pragma ?

v2/plan/PlanGraphGenerator.cpp
rjb32: That's it, I don't want to emphase a state machine behavior. It's more like generate functions that may have a branching or graph behavior, I am not sure that it is the same "current node" everywhere. So when I pass it to functions I am sure that it is the one I intended to be

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -177,119 +211,235 @@ PlanGraphNode* PlanGraphGenerator::generatePatternElementOrigin(const EntityPatt
...
+    currentNode = propMapExpr;
+
+    // All dependencies are branches feeding into propMapExpr, e.g. for property expressions
+    //
+    //       ->  GetProperty ----
+    //       |                  |
+    // varNode -> * -> prev -> propMapExpr -> filter
+
+    for (const auto& [propType, expr] : exprConstraints) {
+        auto* getPropertyNode = _tree.create<GetPropertyNode>();
```
rjb32: Give GetPropertyNode the propType

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -177,119 +211,235 @@ PlanGraphNode* PlanGraphGenerator::generatePatternElementOrigin(const EntityPatt
...
+    if (!varDecl) {
+        throwError("Property expression is missing its variable declaration", expr);
+    }
+
+    PlanGraphNode* varNode = getVarNode(varDecl);
+    if (!varNode) {
+        throwError("Property expression refers to an undefined variable", expr);
+    }
+
+    auto* getPropertyNode = _tree.create<GetPropertyNode>();
```
rjb32: Give PropertyType

LucLabarriere:
@rjb32
- Right now, I'm creating a pair of `Filter`/`VarNode` whether there are filters or not.

Will see later if I can delay the filter to avoid creating it if not needed. I think it's trivial as a small post process to remove empty filters, or ignore them in the PipelineGeneration step.
However, we may as well skip the allocation entirely, will see if that's possible

- Had to create an unnamed `VarDecl` for nodes and edges that don't have a name, will see if I can remove that later or if the design actually requires it. For now it's interesting for debugging
purposes. However, the implementation is ugly so I hope we can remove unnamed variables entirely at some point (storing `unique_ptr<std::string>` in `DeclContext`)

- Moved the VarDecl/Filter management into a PlanGraphVariables class to lighten PlanGraphGenerator

- Removed the difference between `ScanNodes`/`ScanNodesByLabel`.

With the addition of the filter node, it is trivial for the pipelinegenerator to check if the filter has a LabelConstraint, and if so, turn the root into a `ScanNodesByLabel` instead
This is better because if multiple branches define n with different labels, they are all stored in the same filter node: `MATCH (n:Person), (n:Officer)` -> Both labels Person and Officer
are in the same `FilterNode` instead of split into two different `ScanNodesByLabel`. The Pipeline generator will also trivially be able to turn the root into a `ScanNodeProperties` instead if it
thinks it's fastet

- Moved individual nodes into their own files, we have way fewer nodes now but they will get more complex than they originally were

- Right now, property constraints are added whether we need a cartesian product or not. Will see if I can make the difference between: trivial prop constraints, prop constraints with join, prop constraints with cartsian products in the next commits

Queries that will need to be tested once we make the difference between the 3 types of prop constraints:
- `MATCH (n)--(m { prop: n:Person })`: match all n--m pairs where m has a boolean property "prop" equal to true if n is a Person, false otherwise
- `MATCH (n:Person)--(m) WHERE n:Officer`: the where should add a label constraint to n
- `MATCH (n:Person), (p:Officer) WHERE (n)-->(m)`: not supported in the analyzer  right now, but we may want to start considering path expression

Example query:
```
MATCH (n:Person)-->(m { name: n.name }),
      (m:Person),
      (m:Founder { age: 30 })
RETURN n.name, m.name
```

image

v2/AST/decl/DeclContext.h
```diff
@@ -19,9 +22,18 @@ class DeclContext {
...
     VarDecl* getDecl(std::string_view name) const;
 
+    std::string_view storeUnnamedVarName(std::string&& name) {
+        auto inserted = std::make_unique<std::string>(std::move(name));
+        std::string& toreturn = *inserted;
+        _unnamedVarNames.emplace_back(std::move(inserted));
+
+        return toreturn;
+    }
+
```
LucLabarriere: That's a bit ugly but I'm basically creating names for unnamed variables v0, v1, v2 etc. I think it's really helpful for debugging. If it's too ugly we can find a better way

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +130,409 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
+        throwError("Only one edge type constraint is supported for now", edge);
+    }
+
+    auto [var, filter] = _variables.getVarNodeAndFilter(decl);
+    if (!var) {
+        std::tie(var, filter) = _variables.createVarNodeAndFilter(decl);
     } else {
-        currentNode = _tree.create<ScanNodesByLabelNode>(&labelSet);
+        incrementDeclOrders(prevNode->getDeclOdrer(), filter);
     }
```
LucLabarriere, line 341: IncrementDeclOrders is called whenever a branch joins another on a varNode, it goes through the joined branch nodes following its outgoing edges and increment the declOrders so that the var nodes appear to have been defined after. Say we join those two branches by connecting F to A (decl orders are written in parentheses)

```
A(0) -> B(1) -> C(2)
D(0) -> E(1) -> F(2)
```

We get:
```
D(0) -> E(1) -> F(2) -> A(3) -> B(4) -> C(5)
```

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +130,409 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
 
-    // Expression constraints
+    currentNode->connectOut(filter);
+    auto* typedFilter = static_cast<FilterEdgeNode*>(filter);
+
     for (const auto& [propType, expr] : exprConstraints) {
-        FilterNodeExprNode* filter = _tree.create<FilterNodeExprNode>(expr);
-        currentNode->connectOut(filter);
-        currentNode = filter;
+        _propConstraints.push_back(std::make_unique<PropertyConstraint>(var, propType._id, expr));
```
LucLabarriere: Property constraints are stored and placed at the end, once we have the whole graph

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +130,409 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
-        {
-            auto expandEdges = _tree.create<GetInEdgesNode>();
-            currentNode->connectOut(expandEdges);
-            currentNode = expandEdges;
-            break;
+
+    return var;
+}
+
+void PlanGraphGenerator::unwrapWhereExpr(const Expr* expr) {
```
LucLabarriere, line 396: By far the trickest method of planGraphGenerator. It unwraps all conditions separated by an AND:

```
WHERE cond1 AND cond2 AND cond3 AND ...
```

Label/edge type constraints are also treated separately to push them down to the right filter node

```
WHERE n:Person AND e:KNOWS AND ...
```

Once we encounter any expression other than a type constraint or an AND, we store it as a predicate. This query:

```
WHERE n:Person
  AND n.name = "Luc" 
  AND (n.duration = m.duration OR p.duration = 25)
  AND n.duration = 20
```

Gets unwrapped as:

```
|WHERE n:Person AND n.name = "Luc" AND (n.duration = m.duration OR p.duration = 25) AND n.duration = 20
|      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|----WHERE n:Person AND n.name = "Luc" AND (n.duration = m.duration OR p.duration = 25) AND n.duration = 20
|    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|    |----WHERE n:Person AND n.name = "Luc" AND (n.duration = m.duration OR p.duration = 25) AND n.duration = 20
|    |    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
|    |    |----WHERE n:Person AND n.name = "Luc" AND (n.duration = m.duration OR p.duration = 25) AND n.duration = 20
|    |    |          ^^^^^^^^
|    |    |----WHERE n:Person AND n.name = "Luc" AND (n.duration = m.duration OR p.duration = 25) AND n.duration = 20
|    |    |                       ^^^^^^^^^^^^^^
|    |----WHERE n:Person AND n.name = "Luc" AND (n.duration = m.duration OR p.duration = 25) AND n.duration = 20
|                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|----WHERE n:Person AND n.name = "Luc" AND (n.duration = m.duration OR p.duration = 25) AND n.duration = 20
                                                                                        ^^^^^^^^^^^^^^^
```

Whitout the parentheses around the OR, the whole expression is a single predicate evaluated in one step

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +130,409 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
+        if (node->getOpcode() == PlanGraphOpcode::VAR) {
+            auto* varNode = static_cast<VarNode*>(node);
+            varNode->setDeclOdrer(varNode->getDeclOdrer() + declOrder + 1);
+        }
+
+        for (auto* out : node->outputs()) {
+            q.push(out);
+        }
     }
+}
```
LucLabarriere, line 508: Increment all declOrders down the branch

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +130,409 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
+        for (; it != depContainer.end(); ++it) {
+            if (order < it->_var->getDeclOdrer()) {
+                order = it->_var->getDeclOdrer();
+                var = it->_var;
+            }
+        }
+
+        // The constraint is evaluated on var (latest dependency)
+        auto* filter = _variables.getNodeFilter(var);
+        filter->addPropertyConstraint(prop.get());
```
LucLabarriere, line 540: At this point we know that that filter node will have everything it needs to evaluate the expression, so we place it here

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +130,409 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
+                continue;
+            }
+
+            insertDataFlowNode(var, dep._var);
+        }
     }
+}
 
-    return currentNode;
+void PlanGraphGenerator::insertDataFlowNode(const VarNode* node, VarNode* dependency) {
```
LucLabarriere, line 566: Also a tricky function,
I tested many algorithms but this is the only one that I could make work.
- Find the shortest path between the node that evaluates the expr and the dependency
- If the dependency corresponds to the same var node we evaluate the expr on, there is nothing to be done
- If the dep is on the same "branch" (we only had to go backward on the graph) -> MATERIALIZE
- If we had to change direction while exploring, it means the dep is on a parallel branch -> JOIN
- If we couldn't find a path, it means the dep is on a separate "island" -> CARTESIAN PRODUCT (or rather value hash join)

It's important to notice that the joins are made from the tip of the branch the dependency is on, to the current node. We can't just connect the dependency to the filter

Example of a dependency on another island: if B is a dependency required on D, we connect the tip of the branch B is on (which is C): to a JOIN right before D

```
A --> B --> C
N --> M --> D --> E --> F

// After the join
A --> B --> C ---⌄
N --> M -------> JOIN --> D --> E --> F
```

v2/plan/PlanGraphTopology.h
```diff
@@ -0,0 +1,116 @@
...
+class PlanGraphTopology {
+public:
+    enum class PathToDependency {
+        SameVar,
+        BackwardPath,
+        UndirectedPath,
+        NoPath
+    };
+
+    static PathToDependency getShortestPath(const PlanGraphNode* origin, const PlanGraphNode* target) {
```
LucLabarriere: Two stage algorithm, try to find the shortest path between two nodes following outgoing edges. If we can't find one, also follow the incoming edges

v2/plan/nodes/VarNode.h
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    void setDeclOdrer(uint32_t declOdrer) { _declOdrer = declOdrer; }
+
+    uint32_t getDeclOdrer() const { return _declOdrer; }
+    const VarDecl* getVarDecl() const { return _varDecl; }
+
+private:
+    const VarDecl* _varDecl {nullptr};
+    uint32_t _declOdrer {0};
```
cyrusknopf: Should this be `_declOrder`? Also would be prefer default-initialising this to something like INTMAX32 to distinguish it from a possibly valid value (assuming that 0 is a valid `_declOdrer`). Currently this is overwritten in the initialiser list but if this changes it might be useful to catch bugs later.

LucLabarriere: Thank's good catch!
Also, about INT_MAX32, good point. Actually I think that default initializing to 0 is safer because the declOrder is used to find the latest dependencies of predicates. If we use int max value and the initialization goes wrong, it will always get picked up as the last dependecy which could cause even more harm

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +128,410 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
+    }
+
+    if (ends.empty()) {
+        throwError("No end points found", stmt);
+    }
+
+    PlanGraphNode* a = (*ends.begin());
+
+    // If there is only one end point, the loop is skipped
+
```
cyrusknopf: Would this be more readable with an explicit ends.size() == 1 check into early return? Also might be more efficient than calculating and comparing iterators to determine if we should loop or not

v2/plan/nodes/VarNode.h
cyrusknopf: `VarNode::getDeclOdrer` and `VarNode::getDeclOdrer` also have the same small spelling error :)

v2/AST/decl/DeclContext.h
rjb32: I think it's better to treat them as any kind of variable so as a VarDecl. It's fairly common indeed in compilers to create variables that have been created as part of the de-sugarification of syntactic sugar in a language.

v2/AST/decl/PatternData.h
```diff
@@ -11,47 +11,54 @@ namespace db::v2 {
 class CypherAST;
 class Expr;
 
-class NodePatternData {
+class PatternData {
```
rjb32, line 14: Good refactoring

v2/AST/expr/PropertyExpr.h
```diff
@@ -16,11 +17,16 @@ class PropertyExpr : public Expr {
...
+    PropertyTypeID getPropertyType() const { return _propTypeID; }
+
     void setDecl(VarDecl* decl) { _decl = decl; }
 
+    void setPropertyType(PropertyTypeID propTypeID) { _propTypeID = propTypeID; }
+
 private:
     QualifiedName* _name {nullptr};
     VarDecl* _decl {nullptr};
+    PropertyTypeID _propTypeID;
```
rjb32, line 29: Add a meaningful default value if the default constructor of PropertyTypeID does not take care of it

v2/interpreter/QueryInterpreterV2.cpp
rjb32: Indeed, it's my fault. You can add a call to invoke plan gen here

v2/plan/PlanGraphTopology.h
```diff
@@ -0,0 +1,116 @@
...
+        if (target == origin) {
+            return PathToDependency::SameVar;
+        }
+
+        std::queue<const PlanGraphNode*> phase1;
+        std::queue<const PlanGraphNode*> phase2;
```
rjb32: Are the 2 queues really necessary? Could the first queue be reused for the second phase?

v2/plan/ExprDependencies.h
```diff
@@ -0,0 +1,111 @@
+#pragma once
+
+#include <variant>
+
+#include "metadata/LabelSet.h"
+
+#include "PlanGraphVariables.h"
+#include "PlannerException.h"
```
rjb32, line 8: Order of includes! Why PlannerException which is a utility class is actually before the expr hierarchy?

v2/plan/ExprDependencies.h
```diff
@@ -0,0 +1,111 @@
...
+
+            case Expr::Kind::SYMBOL:
+                // throwError("Symbol expression not supported yet", expr);
+                // TODO Find a way to get access to throwError
+                throw PlannerException("Symbol expression not supported yet");
+                break;
+
+            case Expr::Kind::LITERAL:
+                // Reached end
+                break;
```
rjb32, line 78: No default case, this is a bit dangerous

v2/plan/ExprDependencies.h
```diff
@@ -0,0 +1,111 @@
...
+    };
+
+    using Dependency = std::variant<LabelDependency, PropertyDependency>;
+
+    struct ExprDependency {
+        VarNode* _var {nullptr};
+        Dependency _dep;
+    };
+
+    using Container = std::vector<ExprDependency>;
```
rjb32, line 36: Container is just a vector here instead of something with set behavior. If a variable appears several times in an expression, can it be added several times?

v2/plan/PlanGraphTopology.h
rjb32: Put this implementation in a cpp file.

v2/plan/PlanGraphVariables.h
```diff
@@ -0,0 +1,54 @@
...
+        _nextDeclOrder = 0;
+    }
+
+    VarNode* getVarNode(const VarDecl* varDecl) const;
+    FilterNode* getNodeFilter(const VarNode* varNode) const;
+
+    std::tuple<VarNode*, FilterNode*> getVarNodeAndFilter(const VarDecl* varDecl);
+    std::tuple<VarNode*, FilterNode*> createVarNodeAndFilter(const VarDecl* varDecl);
+
+    std::unordered_map<const VarDecl*, VarNode*>& getVarNodesMap() {
```
rjb32: Returning a map by reference should not be done

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -24,104 +48,64 @@
...
+        throwError("Update statements are not supported yet", query);
+    } else {
+        if (!returnStmt) {
+            // Return statement is mandatory if there are no update statements
+            throwError("Return statement is missing", query);
+        }
     }
 
-    // TODO: Return statement
+    // Place joins on vars that have more than one input
```
rjb32, line 99: This is very clean I like it.

v2/plan/PlanGraphGenerator.cpp
rjb32: Yes this is not very readable. And a loop which is not very readable has a very high potential for bugs. The ++ends.begin() expression is hard to understand and to ensure correct. What does a means? This is very much unclear. Please make it plainly obvious.

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +128,410 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
+    if (ends.empty()) {
+        throwError("No end points found", stmt);
+    }
+
+    PlanGraphNode* a = (*ends.begin());
+
+    // If there is only one end point, the loop is skipped
+
+    for (auto itB = ++ends.begin(); itB != ends.end(); itB++) {
+        PlanGraphNode* b = *itB;
```
rjb32: Find a more explicit name than a and b

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +128,410 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
+                a->connectOut(join);
+                b->connectOut(join);
+
+                a = join;
+            } break;
+        }
+    }
+
+    // Connect the last node to the output
+    auto* results = _tree.create<ProduceResultsNode>();
```
rjb32: Avoid using auto to increase readability

v2/plan/PlanGraphGenerator.cpp
```diff
@@ -144,152 +128,410 @@ void PlanGraphGenerator::generateMatchStmt(const MatchStmt* stmt) {
...
+        std::tie(var, filter) = _variables.createVarNodeAndFilter(decl);
+    } else {
+        incrementDeclOrders(prevNode->getDeclOdrer(), filter);
     }
 
-    const NodePattern* targetPattern = dynamic_cast<const NodePattern*>(target);
-    if (!targetPattern) {
-        throw PlannerException("Target pattern element must be a node pattern");
+    currentNode->connectOut(filter);
+    auto* typedFilter = static_cast<FilterNodeNode*>(filter);
```
rjb32: Why is it called typedFilter here? Also avoid using auto for readability

v2/AST/decl/DeclContext.h
LucLabarriere: So you mean it's fine like this?

v2/plan/PlanGraphTopology.h
LucLabarriere: I don't think so because we explore the whole graph following only outputs first. If I put the inputs in the same queue, I have no way to split the exploration, after the first step, I will already be going through outgoing edges instead of only incoming

v2/plan/ExprDependencies.h
LucLabarriere: What do you mean? do you prefer this?

```c++
#include <variant>

#include "expr/BinaryExpr.h"
#include "expr/Expr.h"
#include "expr/NodeLabelExpr.h"
#include "expr/PropertyExpr.h"
#include "expr/StringExpr.h"
#include "expr/UnaryExpr.h"

#include "metadata/LabelSet.h"
#include "PlanGraphVariables.h"
#include "PlannerException.h"
```

v2/plan/ExprDependencies.h
LucLabarriere: No actually it's a safeguard, if we add a kind, the code does not compile. Default cases are actually code smells because you can forget to handle the new cases, I think default cases should always be forbidden for switch cases on enums that cannot be bit fields. For example if I remove the Literal case:
```
/.../ExprDependencies.h:43:16: error: enumeration value ‘LITERAL’ not handled in switch [-Werror=switch]
   43 |         switch (expr->getKind()) {
```

v2/plan/ExprDependencies.h
LucLabarriere: Yes exactly, that might be an issue. I'm not sure what the design should be because it will depend on the executor I think. We need to store multiple dependencies for the same variable because we could have something like this: `a.name = "sth" OR a.age > 20` which puts two dependencies on a, with two different properties. Maybe `std::unordered_set<Dependency>` could work, where we ensure that each var+propID or var+label dependency is stored only once, but we may also make a cache for the getProperties (so that if n.name appears in many predicates, we only do getproperty once). I'll try to think what should be the use cases

v2/plan/PlanGraphVariables.h
LucLabarriere: Replaced by a range:
```c++
class PlanGraphVariables {
public:
   /* ... */
   using FilterNodeMap = std::unordered_map<const VarNode*, FilterNode*>;

    struct FilterNodeMapRange {
        FilterNodeMap::const_iterator _begin;
        FilterNodeMap::const_iterator _end;

        FilterNodeMap::const_iterator begin() const {
            return _begin;
        }

        FilterNodeMap::const_iterator end() const {
            return _end;
        }
    };

    /* ... */

    FilterNodeMapRange getNodeFiltersMap() {
        return {_nodeFiltersMap.begin(), _nodeFiltersMap.end()};
    }

private:
    /* ... */

    // Map of node filters
   FilterNodeMap _nodeFiltersMap;
};
```

What do you think?

v2/plan/PlanGraphGenerator.cpp
LucLabarriere: Even for trivial cases like this one ?
``` c++
ProduceResultsNode* results = _tree.create<ProduceResultsNode>();
std::unique_ptr<Something> sth = std::make_unique<Something>();
// etc
```

v2/plan/PlanGraphGenerator.cpp
LucLabarriere: Yeah it's not really good... That because we make the distinction between FilterNodeNode and FilterEdgeNode... Trying to come up with clearer code

LucLabarriere: What do you think of this:
- Rename FilterNodeNode/FilterEdgeNode -> NodeFilterNode/EdgeFilterNode
- Keep FilterNode -> (Node/Edge)FilterNode hiearchy
- add asNodeFilter and asEdgeFilter methods to FilterNode
```c++
class FilterNode {
    /* ... */
    NodeFilterNode* asNodeFilter();
    const NodeFilterNode* asNodeFilter() const;

    EdgeFilterNode* asEdgeFilter();
    const EdgeFilterNode* asEdgeFilter() const;
    /* ... */
};

/* ... */
/* Defined once we have the definition of NodeFilterNode and EdgeFilterNode */
inline NodeFilterNode* FilterNode::asNodeFilter() {
    return static_cast<NodeFilterNode*>(this);
}

inline const NodeFilterNode* FilterNode::asNodeFilter() const {
    return static_cast<const NodeFilterNode*>(this);
}

inline EdgeFilterNode* FilterNode::asEdgeFilter() {
    return static_cast<EdgeFilterNode*>(this);
}

inline const EdgeFilterNode* FilterNode::asEdgeFilter() const {
    return static_cast<const EdgeFilterNode*>(this);
}


/// In PlanGraphGenerator.cpp

// Instead of:
auto* edgeFilter = static_cast<EdgeFilterNode*>(filter);

// This:
EdgeFilterNode* edgeFilter = filter->asEdgeFilter();
```

v2/plan/PlanGraphGenerator.cpp
LucLabarriere: I tried to come up with a clearer code and better namings + added an explanation of the algorithm, what do you guys think @cyrusknopf @rjb32 

```c++
void PlanGraphGenerator::generateReturnStmt(const ReturnStmt* stmt) {
    // Step 1: Find all end points
    std::vector<PlanGraphNode*> ends;
    for (const auto& node : _tree.nodes()) {
        if (node->outputs().empty()) {
            ends.push_back(node.get());
        }
    }

    if (ends.empty()) {
        throwError("No end points found", stmt);
    }

    if (ends.size() == 1) {
        // No joins needed, just generate the last ProduceResultsNode
        auto* results = _tree.create<ProduceResultsNode>();
        ends.back()->connectOut(results);
    }

    // Step 2: Generate all joins
    // Algorithm:
    // - Pick the first endpoint (= rhs)
    // - For each other endpoint (= lhs):
    //     - Find the shortest path between the lhs and rhs
    //     - If the path is undirected, JOIN the two endpoints
    //     - If no path is found, CARTESIAN_PRODUCT the two endpoints
    //     - rhs becomes the join node
    /*       A              A              A
     *      / \            / \            / \
     *     B   C          B   C          B   C
     *    /     \    ->  /     \    ->  /     \
     *   D       F      D       F      D       F
     *    \     / \      \     / \      \     / \
     *     H   I   J      H   I   J      H   I   J
     *                         \ /        \   \ /
     *                         [u]         \  [u]
     *                                      \ /
     *                                      [u]   */

    PlanGraphNode* rhsNode = ends[0];

    for (size_t i = 1; i < ends.size(); i++) {
        PlanGraphNode* lhsNode = ends[i];

        const auto path = PlanGraphTopology::getShortestPath(rhsNode, lhsNode);

        switch (path) {
            case PlanGraphTopology::PathToDependency::SameVar:
                // Should not happen
                throwError("Unknown error", rhsNode);
                continue;
            } break;

            case PlanGraphTopology::PathToDependency::UndirectedPath: {
                // Join
                JoinNode* join = _tree.create<JoinNode>();
                rhsNode->connectOut(join);
                lhsNode->connectOut(join);

                rhsNode = join;
            } break;
                // Cartesian product
                CartesianProductNode* join = _tree.create<CartesianProductNode>();
                rhsNode->connectOut(join);
                lhsNode->connectOut(join);

                rhsNode = join;
            } break;
        }
    }

    // Step 3: Connect the last node to the output
    auto* results = _tree.create<ProduceResultsNode>();
    rhsNode->connectOut(results);
}
```

v2/AST/expr/PropertyExpr.h
LucLabarriere: It does take care of it, initialised to an invalid PropertyTypeID (max uint64_t):
image

v2/plan/PlanGraphGenerator.cpp
cyrusknopf: This seems much clearer to me. As an aside from the code I think it would be helpful to have a page on Notion which justifies/explains why each case results in a Join/CartProd, etc. so that it is documented somewhere for future - if there isn't already.

v2/plan/PlanGraphVariables.h
cyrusknopf: > Returning a map by reference should not be done

Why not?

v2/plan/PlanGraphGenerator.cpp
LucLabarriere: Yeah you're right, in fact it's not 100% clear yet ahah, we know a few things already:
- If one node is the parent of the other (was found by only walking backward on the plan graph), it means the column we need is already in our current block (it was generated in the same pipeline "branch"), so we should just be able to "MATERIALIZE", if you're not sure what it means, don't worry, we're gonna have to talk about it in a meeting anyway because it's a side effect of our column oriented design and it's a bit tricky. That only applies when we place predicates because there is no way that an endpoint is the parent of another node (an endpoint has no children)
- If no path could be found the nodes are on completely isolated "islands", there is no relation between the two data blocks, so mathematically, the only thing we can do is to make a cartesian product. Now, if there is a condition (predicate) to apply during the cartesian product, there are optimization techniques that we can use  (ValueHashJoin), but that does not apply at this stage, it's used when we place joins for predicate evaluation
- If a path can be found between the two nodes but we had to following outgoing edges (undirected path), it means the two nodes are on parallel branches and we need to join them. We could do a CARTESIAN_PRODUCT here too (as if nodes were on two different islands), but I suspect one branch filters the other (since they have a common parent), so I need to think a bit more about this

v2/AST/decl/DeclContext.h
rjb32: We should create VarDecl objects for them

v2/plan/PlanGraphTopology.h
rjb32: Indeed I didn't see the write to phase2 queue inside the first loop

v2/plan/ExprDependencies.h
rjb32: Yes exactly

v2/plan/ExprDependencies.h
rjb32: Ok the vector maybe fine for now

v2/plan/PlanGraphGenerator.cpp
rjb32: This is much much clearer, thanks.

v2/plan/PlanGraphGenerator.cpp
rjb32: I like the EdgeFilterNode/NodeFilterNode names

v2/plan/PlanGraphGenerator.cpp
rjb32: For the unique_ptr & std::make_unique this is a good use

v2/AST/decl/DeclContext.h
LucLabarriere: Ok so that's what's done right now, the only ugly thing is that we create a name for unnamed variable (interesting mostly for debugging purposes). Do you want to keep it ? Or should I try to make it work without allocating a new string?

rjb32: If you can make it work without allocating a new string without too much uglyness then why not, otherwise that's fine

v2/plan/PlanGraphVariables.h
```diff
@@ -0,0 +1,66 @@
...
+
+class VarDecl;
+class VarNode;
+class FilterNode;
+class PlanGraphNode;
+class PlanGraph;
+
+class PlanGraphVariables {
+public:
+    explicit PlanGraphVariables(PlanGraph& tree);
```
rjb32: Pass pointer for uniformity of passing mechanism

v2/AST/decl/DeclContext.h
```diff
@@ -19,9 +22,18 @@ class DeclContext {
 
     VarDecl* getDecl(std::string_view name) const;
 
+    std::string_view storeUnnamedVarName(std::string&& name) {
```
rjb32: Move to CypherAST for context

rjb32:
TODO:
* Keep vector of dependencies
* Add exception for loops in patterns -> add solution in comments
* Store the strings for temporary names in CypherAST

v2/AST/CypherAST.cpp
```diff
@@ -92,6 +92,17 @@ CypherAST::~CypherAST() {
     for (EdgePatternData* data : _edgePatternDatas) {
         delete data;
     }
+
+    for (std::string* name : _unnamedVarIdentifiers) {
+        delete name;
+    }
+}
+
+std::string* CypherAST::createUnnamedVarIdentifier() {
```
rjb32: Rename this function into createString()

v2/AST/CypherAST.cpp
```diff
@@ -92,6 +92,17 @@ CypherAST::~CypherAST() {
...
+    std::string* name = new std::string;
```
rjb32: Use parenthesis for the constructor new std::string()

PR #41, opened 2025-09-18
BUG: Edges spanning commits incorrectly returned in MATCH

cyrusknopf:
New test added:
1. Create new change 1
2. `CREATE (n:NEWNODE)` <-- local ID 0
3. Change 1 is committed
4. `CREATE (n @ 0) - [:NEWEDGE] - (m:NEWNODE)` <-- between local node 0 and a new node
    `CREATE (m:NEWNODE) - [:NEWEDGE] - (n @ 0)` <-- between new node and local node 0
6. Change 1 is committed
7. Create change 2
8. `CREATE (n:NEWNODE)` <-- local ID 0
9. Change 2 is committed
10. `CREATE (n @ 0) - [:NEWEDGE] - (m:NEWNODE)` <-- between local node 0 and a new node
    `CREATE (m:NEWNODE) - [:NEWEDGE] - (n @ 0)` <-- between new node and local node 0
(i.e. Change1 and Change2 do the same thing)
11. Change 2 is submitted
12. On main: `(0)-[0]-(1)` and `(2)-[1]-(0)` exist, this is correct
14. Checkout change 1
15. Change 1 is submitted
16. On main, we should have 6 nodes, 4 edges, since both changes each created 3 nodes and 2 edges which did not exist when the change was branched
i.e. we should have
```
0 - 0 - 1
2 - 1 - 0
3 - 2 - 4
5 - 3 - 3
```
so nodes 3,4,5 on main are what Change 1 thought of as 0,1,2.

The result is that actually the edges from Change 1 point to Node 0 which was created by Change 2, and not Node 3, which is the materialisation of Change 1's node 0 on main.

Output of `ctest -R edgeSpanningCommits --output-on-failure`

```
Testing query: match (n) return n
Testing query: match (n)-[e]-(m) return n,e,m
/home/dev/turingdb/test/query/QueryTester.h:113: Failure
// Actual vs expected source NodeIDS
Expected equality of these values:
  c->getRaw()
    Which is: { 0, 2, 0, 5 }
  ec->getRaw()
    Which is: { 0, 2, 3, 5 }
/home/dev/turingdb/test/query/QueryTester.h:113: Failure
// Actual vs expected target NodeIDs
Expected equality of these values:
  c->getRaw()
    Which is: { 1, 0, 4, 0 }
  ec->getRaw()
    Which is: { 1, 0, 4, 3 }
```

### DIFF
--- a/storage/versioning/DataPartRebaser.h
+++ b/storage/versioning/DataPartRebaser.h
@@ -14,12 +14,14 @@ public:
     DataPartRebaser() = default;
     ~DataPartRebaser() = default;
 
-    DataPartRebaser(NodeID baseNxtNodeID, EdgeID baseNxtEdgeID, NodeID tgtNxtNodeID,
+    DataPartRebaser(NodeID baseNxtNodeID,
+                    EdgeID baseNxtEdgeID,
+                    NodeID tgtNxtNodeID,
                     EdgeID tgtNxtEdgeID)
         : _baseCommitNextNodeID(baseNxtNodeID),
-          _baseCommitNextEdgeID(baseNxtEdgeID),
-          _targetCommitNextNodeID(tgtNxtNodeID),
-          _targetCommitNextEdgeID(tgtNxtEdgeID)
+        _baseCommitNextEdgeID(baseNxtEdgeID),
+        _targetCommitNextNodeID(tgtNxtNodeID),
+        _targetCommitNextEdgeID(tgtNxtEdgeID)
     {
     }
 


rjb32:
Very very interesting!

LucLabarriere:
Really nice catch ! Super tricky bug

storage/versioning/Change.cpp
```diff
@@ -68,10 +69,24 @@ CommitResult<void> Change::commit(JobSystem& jobsystem) {
...
+    // Get the next Edge and Node IDs at the time this change branched
+    NodeID oldNextNodeID =
+        _base->commits().back().openTransaction().readGraph().getNodeCount();
+    EdgeID oldNextEdgeID =
+        _base->commits().back().openTransaction().readGraph().getEdgeCount();
+
+    // Get current state of main
     _base = _versionController->openTransaction().commitData();
 
+    // Get the current next Edge and Node IDs on main
```
rjb32, line 81: What about atomicity or locking? How do we know the node and edge count won't move in the meantime?

storage/versioning/DataPartRebaser.h
```diff
@@ -14,6 +14,15 @@ class DataPartRebaser {
     DataPartRebaser() = default;
     ~DataPartRebaser() = default;
 
+    DataPartRebaser(NodeID baseNxtNodeID, EdgeID baseNxtEdgeID, NodeID tgtNxtNodeID,
```
rjb32: Put them each on their own line or not at all

storage/versioning/DataPartRebaser.h
```diff
@@ -14,6 +14,15 @@ class DataPartRebaser {
...
+                    EdgeID tgtNxtEdgeID)
+        : _baseCommitNextNodeID(baseNxtNodeID),
+          _baseCommitNextEdgeID(baseNxtEdgeID),
```
rjb32: Align with the ":" character

storage/versioning/Change.cpp
cyrusknopf: Change::rebase is only called by VersionController::submitChange which takes a lock for submitting to main. This is the same way the existing rebase is ensured to be atomic

storage/versioning/DataPartRebaser.h
cyrusknopf: Fixed

storage/versioning/DataPartRebaser.h
cyrusknopf: Fixed

PR #42, opened 2025-09-18
Fix Python Json Parse Error Bug

ghost:
[Jira Issue](https://turingbio.atlassian.net/browse/TURING-800)

### DIFF
--- a/regress/CMakeLists.txt
+++ b/regress/CMakeLists.txt
@@ -18,6 +18,7 @@ endmacro()
 
 # subdirectories
 add_subdirectory(create_turingdb_object)
+add_subdirectory(json_error_msg)
 add_subdirectory(list_graphs)
 add_subdirectory(change_submit_alice)
 add_subdirectory(change_10)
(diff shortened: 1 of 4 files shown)


ghost:
I didn't have to touch the JsonEncoder.h code so I decided to leave it as it was ( It might be confusing in the future as most of the other network processing code is all in header files)

query/parser/QueryParser.cpp
```diff
@@ -35,14 +35,33 @@ QueryCommand* QueryParser::parse(std::string_view query) {
...
         throw ParserException(std::move(errMsg));
     }
-    
+
     return _astCtxt->getRoot();
 }
 
+std::string sanitizeJsonString(const std::string& input) {
+    std::string result;
+    result.reserve(input.size() * 1.2); // Pre-allocate
```
rjb32, line 44: Why times 1.2? Why not using the golden ratio for example? 1.618

ghost: I think a 20% increase in query string size is already pretty generous, right?

PR #45, opened 2025-09-19
Dumping Commits On Submit

ghost:
This PR lets us dump the new commits in a change at submit time.

### DIFF
--- a/storage/GraphSerializer.cpp
+++ b/storage/GraphSerializer.cpp
@@ -14,11 +14,11 @@ GraphSerializer::GraphSerializer(Graph* graph)
 }
 
 DumpResult<void> GraphSerializer::load() const {
-    spdlog::info("Loading graph {} from {}", _graph->getName(), _graph->getPath().get());
+    spdlog::info("Loading graph {}", _graph->getName());
     return GraphLoader::load(_graph, _graph->getPath());
 }
 
 DumpResult<void> GraphSerializer::dump() const {
-    spdlog::info("Dumping graph {} to {}", _graph->getName(), _graph->getPath().get());
+    spdlog::info("Dumping graph {}", _graph->getName());
     return GraphDumper::dump(*_graph, _graph->getPath());
 }
--- a/system/SystemManager.cpp
+++ b/system/SystemManager.cpp
@@ -40,10 +40,8 @@ void SystemManager::init() {
     const fs::Path defaultPath = _config->getGraphsDir() / "default"; auto found = std::find(list.value().begin(), list.value().end(), defaultPath);
 
     if (found != list->end()) {
-        spdlog::info("loading default");
         _defaultGraph = loadGraph("default");
     } else {
-        spdlog::info("creating default");
         _defaultGraph = createGraph("default");
     }
 


ghost:
@rjb32 I have a bigger issue that will take maybe more architectural/thinking time than I have now :

When loading a graph we build it by making and submitting changes - but there is no graph directory for these changes to be dumped.

Before I leave I will see if I can write the parent graph directory and the metadata files. But I am not sure if I will get it done - so I have left my work at this point. What works is:

- we can load in a turing binary and make changes which will be committed.
- the same applies for the default graph - which acts as a scratch pad.

ghost:
Actually it looks like there are many cases in our testing where we create graphs that shouldn't be dumped. So we should probably have a parameter to check that before dumping (probably in dump and load manager).

storage/CMakeLists.txt
```diff
@@ -74,5 +74,7 @@ target_link_libraries(turing_db_storage_s PRIVATE
 
 target_link_libraries(turing_db_storage_s PUBLIC
         turing_db_base_s
+        turing_db_io_fs_s
+        turing_db_dump_s
```
LucLabarriere: @rjb32 
Are we sure we want the storage library to depend on dump? There would be a circular dependency storage <-> dump

LucLabarriere:
@rjb32 

Ready for review. Big changes on the architecture, storage doesn't depend on the dump library anymore. Everything is moved to SystemManager. Each graph gets attribute a GraphSerializer (that has the mutex). in the ChangeStep, on submit, we call systemManager->dumpGraph() which dumps the graph (only the missing the commits). In order to do so, each graph has a new ID that is written in the info file on disk. When we try to dump, if the folder already exists, we check if the graph is the same one. If it is we just dump the missing commits.

I also simplified the systemManager loadGraph calls. We now have only loadGraph(graphName) which loads a turing graph, and if we want to load a graph with possibly another format, we need to specify the file name and call loadGraphFromFile. This way, in the query language, when we do "LOAD GRAPH reactome", it can only load reactome if it is in a turing format. If not, we have to run 'LOAD GRAPH "reactome_dump" "reactome"' which imports the graph, and dumps it as a turing graph.

I also added an inMemory only mode (new argument to turingdb tool + option in TuringConfig for unit tests)

system/TuringConfig.cpp
```diff
@@ -8,55 +8,56 @@
 
 using namespace db;
 
-TuringConfig::TuringConfig()
-{
-    init();
+TuringConfig::TuringConfig() {
```
rjb32: Prefer the braces on next line syntax for constructors

system/SystemManager.h
```diff
@@ -26,7 +27,7 @@ class Change;
 
 class SystemManager {
 public:
-    explicit SystemManager(const TuringConfig& config);
```
rjb32, line 29: Why is config passed as a non-cost reference? This is always a bit of code smell when we see a config object passed as non-const

system/TuringConfig.h
```diff
@@ -13,14 +13,18 @@ class TuringConfig {
...
+
+    void init();
 
 private:
     fs::Path _turingDir;
     fs::Path _graphsDir;
     fs::Path _dataDir;
 
-    void init();
+    bool _syncedOnDisk = true;
```
rjb32: Prefer braces initialization style for default values

system/GraphSerializer.h
```diff
@@ -0,0 +1,32 @@
...
+
+    DumpResult<void> load() const;
+    DumpResult<void> dump() const;
+
+    DumpResult<void> dumpCommit(const Commit& commit) const;
+
+private:
+    mutable std::mutex _mutex;
+
+    Graph* _graph;
```
rjb32: Give a nullptr default value for pointer members

system/GraphSerializer.cpp
```diff
@@ -0,0 +1,28 @@
+#include "GraphSerializer.h"
+
+#include <mutex>
+
+#include "Graph.h"
+#include "GraphLoader.h"
+#include "GraphDumper.h"
+#include "spdlog/spdlog.h"
```
rjb32: Fix the order of includes: spdlog is an external library, put the include right below the system includes

system/SystemManager.cpp
```diff
@@ -12,44 +12,128 @@
...
+    init();
 }
 
 SystemManager::~SystemManager() {
 }
 
+void SystemManager::init() {
+    const auto list = _config.getGraphsDir().listDir();
+
+    if (std::find(list->begin(), list->end(), _config.getGraphsDir() / "default") != list->end()) {
```
rjb32: Bad style: the condition in the if is too complicated, prefer to get the result of the find in a temporary variable before the if statement and test equality with end iterator

system/SystemManager.cpp
```diff
@@ -12,44 +12,128 @@
...
+        _defaultGraph = createAndDumpGraph("default");
+    }
+
+    if (!_defaultGraph) {
+        throw TuringException("Could not initialise the default graph");
+    }
+}
+
+Graph* SystemManager::loadGraph(const std::string& name) {
+    const fs::Path path = _config.getGraphsDir() / name;
```
rjb32: Rename path into graphPath

system/SystemManager.cpp
```diff
@@ -12,44 +12,128 @@
...
+    auto graphSerializer = std::make_unique<GraphSerializer>(graph.get());
+
+    auto* graphPtr = graph.get();
+
+    if (auto res = graphSerializer->load(); !res) {
+        spdlog::error(res.error().fmtMessage());
+        return nullptr;
+    }
+
+    if (!addGraph(std::move(graph), std::move(graphSerializer))) {
```
rjb32: Please use an ownership style where the static create method for Graph registers the Graph into SystemManager instead of relying on unique_ptr and moving it later.

rjb32: This is a bit ugly that the graphSerializer depends on what graph but it is somewhat outside it. We should also find a way for graphSerializer to not rely on the std::move & unique_ptr style and make it reside and be owned by some clear owner.

system/SystemManager.cpp
```diff
@@ -12,44 +12,128 @@
...
+    const fs::Path path = _config.getGraphsDir() / name;
+
+    auto graph = Graph::create(name, path.c_str());
+    auto graphSerializer = std::make_unique<GraphSerializer>(graph.get());
+
     auto* rawPtr = graph.get();
 
-    if (!addGraph(std::move(graph), name)) {
+    if (_config.isSyncedOnDisk()) {
+        if (auto res = graphSerializer->dump(); !res) {
```
rjb32: From what I can see so far, it seems like GraphSerializer would make sense to be owned by Graph

system/SystemManager.cpp
```diff
@@ -12,44 +12,128 @@
...
+    const auto& name = graph->getName();
+
     // Search if a graph with the same name exists
     const auto it = _graphs.find(name);
     if (it != _graphs.end()) {
         return false;
     }
 
+    // Search if a serializer with the same graph exists
+    const auto serializerIt = _serializers.find(graph.get());
```
rjb32: Yeah the fact that we have a separate map with maybe a serializer or not for a graph, separate from the graph is a bit fishy. It adds a state which is separate from graph but also dependent on it at the same time and it adds moving pieces.

query/AST/QueryCommand.cpp
```diff
@@ -84,8 +84,8 @@ ListGraphCommand* ListGraphCommand::create(ASTContext* ctxt) {
 
 // LoadGraphCommand
 
-LoadGraphCommand::LoadGraphCommand(const std::string& fileName)
-    :_fileName(fileName)
+LoadGraphCommand::LoadGraphCommand(const std::string& graphName)
+    :_graphName(graphName)
```
rjb32: Put a space between the : character and the underscore

storage/Graph.h
```diff
@@ -33,16 +34,18 @@ class Graph {
...
 
+    [[nodiscard]] GraphID getID() const { return _graphID; }
     [[nodiscard]] CommitHash getHeadHash() const;
 
     [[nodiscard]] static std::unique_ptr<Graph> create();
-    [[nodiscard]] static std::unique_ptr<Graph> create(const std::string& name);
+    [[nodiscard]] static std::unique_ptr<Graph> create(const std::string& name, const std::string& path);
     [[nodiscard]] static std::unique_ptr<Graph> createEmptyGraph();
-    [[nodiscard]] static std::unique_ptr<Graph> createEmptyGraph(const std::string& name);
+    [[nodiscard]] static std::unique_ptr<Graph> createEmptyGraph(const std::string& name, const std::string& path);
```
rjb32: I don't think it looks very good to have a distinction between create() and createEmptyGraph()? What does this even mean? At first glance it's hard to tell

storage/versioning/VersionController.h
```diff
@@ -73,8 +75,7 @@ class VersionController {
     std::unique_ptr<ArcManager<CommitData>> _dataManager;
     std::unique_ptr<ArcManager<DataPart>> _partManager;
 
-    void lock();
-    void unlock();
+    std::unique_lock<std::mutex> lock();
```
rjb32, line 78: Please don't make functions that return a unique_lock. Why is this necessary?

system/SystemManager.h
```diff
@@ -36,8 +37,14 @@ class SystemManager {
 
     const TuringConfig& getConfig() const { return _config; }
 
+    void init();
+
+    Graph* loadGraph(const std::string& graphName);
+
     Graph* createGraph(const std::string& graphName);
 
+    Graph* createAndDumpGraph(const std::string& graphName);
```
rjb32: Why 2 different create functions that are subtly different? This is not looking very good

system/SystemManager.h
```diff
@@ -81,11 +87,12 @@ class SystemManager {
 
 
 private:
-    const TuringConfig& _config;
+    TuringConfig& _config;
     mutable RWSpinLock _graphsLock;
     Graph* _defaultGraph {nullptr};
     std::unique_ptr<S3::TuringS3Client<S3::AwsS3ClientWrapper<>>> _s3Client {nullptr};
     std::unordered_map<std::string, std::unique_ptr<Graph>> _graphs;
+    std::unordered_map<Graph* , std::unique_ptr<class GraphSerializer>> _serializers;
```
rjb32: I really don't like this

system/SystemManager.h
LucLabarriere: Indeed... That's because I was calling init() on the config inside SystemManager but that's bad...

The `TuringConfig::init()` function creates the paths (.turing, .turing/data and .turing/graphs) if they don't exist. I think we have a few choices here:

- We keep the current system but remove the init from system manager, we just need make sure to not forget the call init when we create the config, before turingDB is initialized:
```c++
    TuringConfig config;
    config.setSyncedOnDisk(true);
    config.init() // Creates the directory structure if it does not exist
    TuringDB turingDB(config);
```

- We put everything inside TuringDB, we keep it as the only possible entry point to the database and we have an interface like this one:

```c++
TuringDB turing;
TuringConfig& config turing.getConfig();
config.setSyncedOnDisk(true);

turing.run(); // Calls config.init() and does other things
```

The problem with this approach is that most unit tests will have to depend on the db library instead of just System but maybe that's fine? Actually, TuringDB seems to simply be a wrapper around JobSystem + SystemManager, maybe we could merge the db/ and system/ libraries

- We make use a builder pattern for turingDB, something like

```c++
TuringDB turing = TuringDBBuilder {}
    .config()
    .setSyncedOnDisk(true)
    ...
    .build(); // Calls init
```

- We make TuringDB call config.init() in the constructor, or maybe even make SystemManager create the directory structure itself based on the config given. It's a bit weird that a config has an init() function anyway
```c++
TuringConfig config;
config.setSyncedOnDisk(true)
TuringDB turing { config };
```

Maybe this last one is better

system/SystemManager.cpp
LucLabarriere: The owner of the GraphSerializer is the owner of the unique_ptr, which is SystemManager. SystemManager owns the GraphSerializers the same way that it owns the graphs. I'll see if I can rework a bit so that we don't have to call this addGraph, and that the graph is simply created and owned by SystemManager from the beginning to the end.

The truth is that SystemManager is really ugly right now because it does too much (It's basically TuringDB if we think about it). I think we need to sit down and think about its responsibilities. Here's what it does:
- Store the graphs
- Store the graph serializers
- Store the changes
- Has a "defaultGraph" pointer
- Handles imports (has a Neo4jImporter member)
- Handles s3 downloads/uploads
- Handles graph loading/dumping (with a loading status manager)

I think we should rework that:
- For graphs
    - The changes need to be owned by the graphs themselves
    - We can argue that the graph serializers also need to be owned by the Graph. It requires the merge the dump and storage libraries but honestly it makes sense, especially if we want to read from disk later on
   - All graph operations (opening a transaction, listing loaded graphs, loading a graph, creating a graph etc) should be done by a `GraphManager`
- For imports
    - This may be handled by the GraphManager, but I think it's already too much so we should have an ImportManager class that converts any incoming graph format, to our own format. I think a design where "ImportManager creates a graph and gives the ownership to the GraphManager once it's built" would work well, but I know that you don't like passing ownership around so can find a better architecture for that, maybe the GraphManager creates an empty graph and gives a reference of it to the ImportManager would be better
- For s3 operations, obviously something like a DataManager of some sort
- For loading/dumping/importing status managers, this looks a bit like a job system with a queue where people can submit "please load this graph", or "please make a backup of this graph", "plaese import my data", etc... So it's bit complex because it touches many areas of the system, but I'm pretty sure we can have a centralized job monitoring system that tracks the status of jobs

If we look at the members of TuringDB, we see this:
```c++
private:
    std::unique_ptr<SystemManager> _systemManager;
    std::unique_ptr<JobSystem> _jobSystem;
```

To be fair, it could very easily be this:
```c++
private:
    std::unique_ptr<GraphManager> _graphManager;
    std::unique_ptr<JobMonitor> _JobMonitor; // contains JobSystem and
                                             // monitors job statuses (find better name)
    std::unique_ptr<DataManager> _dataManager; // s3 stuff, sync with the data/ folder
    std::unique_ptr<ImportManager> _importManager; // imports graphs from data/ into graphs/
```

system/SystemManager.cpp
LucLabarriere: Agree yeah, if you want to have a quick call today let me know

storage/Graph.h
LucLabarriere: Yeah it's really bad actually. I really thought of this but I struggle to find a better solution (although I think it's critical that I find one). The problem is that a graph has to be created with at least one commit so that it's in a valid state (we need to be able to get the head atomically). I'll try to come up with something today

storage/versioning/VersionController.h
LucLabarriere: This is necessary to be able to load/dump a graph while it's locked by one user. I can try to come up with a better way to do that, but at least replacing the lock()/unlock() pair functions by this is a much better improvement because there's no way to forget to unlock... I'll think of a better solution today to avoid this, we could use the transaction system for this but not sure. it's particularly important since we want to be able to dump missing commits and we really don't want someone to submit a change while we do that

system/SystemManager.h
LucLabarriere: Yeah it's bad, I'll see if I can change that but it might not be easy

system/SystemManager.h
rjb32: Use the following style:
```c++
TuringDBConfig config;
config.setSyncedOnDisk(true);

TuringDB turing(config);
turing.run(); // Creates directories
```

pipeline/operations/ChangeStep.cpp
```diff
@@ -111,8 +112,18 @@ ChangeResult<void> ChangeStep::acceptChange() const {
...
-    
-    return changeMan.acceptChange(tx.changeAccessor(), *_jobSystem);
+
+    // Step 1: Submit the change
+    if (auto res = changeMan.submitChange(tx.changeAccessor(), *_jobSystem); !res) {
+        return res;
+    }
+
+    // Step 2: Dump newly created commits
+    if (auto res = _sysMan->dumpGraph(_graph->getName()); !res) {
```
rjb32: Put GraphSerializer into Graph and call serializer->dump()

system/SystemManager.h
rjb32: Remove createAndDumpGraph() and put dump into create and do an if on the config

storage/CMakeLists.txt
rjb32: We take the decision to merge dump and storage

db/TuringDB.cpp
```diff
@@ -1,19 +1,74 @@
...
     _jobSystem(JobSystem::create())
 {
 }
 
-TuringDB::~TuringDB()
-{
+TuringDB::~TuringDB() {
+}
+
+void TuringDB::run() {
```
rjb32, line 24: This looks really good I like it

pipeline/PipelineStep.cpp
```diff
@@ -230,15 +230,21 @@ PipelineStep::PipelineStep(CallLabelSetStep::Tag,
...
-PipelineStep::PipelineStep(LoadGraphStep::Tag, const std::string& fileName)
+PipelineStep::PipelineStep(LoadGraphStep::Tag, const std::string& graphName)
     : _opcode(PipelineOpcode::LOAD_GRAPH),
-    _impl(std::in_place_type<LoadGraphStep>, fileName)
+    _impl(std::in_place_type<LoadGraphStep>, graphName)
 {
 }
 
-PipelineStep::PipelineStep(LoadGraphStep::Tag, const std::string& fileName, const std::string& graphName)
+PipelineStep::PipelineStep(LoadGraphStep::Tag, const fs::Path& filePath, const std::string& graphName)
```
rjb32: While is there still a LoadGraphStep that can take a filePath?

pipeline/operations/ImportGraphStep.cpp
```diff
@@ -0,0 +1,48 @@
...
+void ImportGraphStep::execute() {
+    Profile profile {"ImportGraphStep::execute"};
+
+    bool res {false};
+
+    if (_filePath.empty() || _graphName.empty()) {
+        throw PipelineException("Invalid graph name or file path");
+    }
+
+    res = _sysMan->importGraph(_graphName, _filePath, *_jobSystem);
```
rjb32: Do a const res here

storage/Graph.cpp
```diff
@@ -25,34 +26,29 @@ CommitHash Graph::getHeadHash() const {
...
 std::unique_ptr<Graph> Graph::create() {
-    auto* graph = new Graph;
+    auto* graph = new Graph();
     graph->_versionController->createFirstCommit();
     return std::unique_ptr<Graph> {graph};
 }
 
-std::unique_ptr<Graph> Graph::create(const std::string& name) {
-    auto* graph = new Graph(name);
+std::unique_ptr<Graph> Graph::create(const std::string& name, const fs::Path& path) {
```
rjb32, line 34: Why is there this second variant?

storage/GraphSerializer.cpp
```diff
@@ -0,0 +1,25 @@
+#include "GraphSerializer.h"
+
+#include <spdlog/spdlog.h>
+
+#include "Graph.h"
+#include "dump/GraphLoader.h"
+#include "dump/GraphDumper.h"
+
+
```
rjb32: Too many blank lines

storage/versioning/VersionController.cpp
```diff
@@ -93,12 +93,8 @@ std::unique_ptr<Change> VersionController::newChange(CommitHash base) {
     return Change::create(this, ChangeID {_nextChangeID.fetch_add(1)}, base);
 }
 
-void VersionController::lock() {
-    _mutex.lock();
-}
-
-void VersionController::unlock() {
-    _mutex.unlock();
+std::unique_lock<std::mutex> VersionController::lock() {
```
rjb32, line 96: This is ugly

system/TuringConfig.h
```diff
@@ -9,18 +9,27 @@ class TuringConfig {
...
+
     const fs::Path& getTuringDir() const { return _turingDir; }
     const fs::Path& getGraphsDir() const { return _graphsDir; }
     const fs::Path& getDataDir() const { return _dataDir; }
 
+    bool isSyncedOnDisk() const { return _syncedOnDisk; }
     void setTuringDirectory(const fs::Path& turingDir);
+    void setSyncedOnDisk(bool syncedOnDisk) { _syncedOnDisk = syncedOnDisk; }
+
+    [[nodiscard]] static TuringConfig createDefault();
```
rjb32, line 25: It's a bit weird that we have a public default constructor and a static createDefault method. I think it's better if we have a TuringConfig() and a init method or something of the sort. But not 2 ways that represent creation

system/SystemManager.cpp
```diff
@@ -12,44 +12,98 @@
...
+void SystemManager::init() {
+    const auto list = _config->getGraphsDir().listDir();
+
+    if (!list) {
+        throw TuringException(fmt::format("Can not list graphs in turing directory '{}'",
+                                          _config->getGraphsDir().get()));
+    }
+
+    const fs::Path defaultPath = _config->getGraphsDir() / "default";
+    auto found = std::find(list.value().begin(), list.value().end(), defaultPath);
```
rjb32: found can be const

pipeline/operations/ChangeStep.cpp
```diff
@@ -87,7 +88,7 @@ ChangeResult<Change*> ChangeStep::createChange() const {
 
     const auto& graphName = std::get<std::string>(_changeInfo);
 
-    const auto res = _sysMan->newChange(graphName);
+    auto res = _sysMan->newChange(graphName);
```
rjb32, line 85: res can be const

pipeline/operations/LoadGraphStep.cpp
```diff
@@ -9,14 +9,14 @@
...
-    :_fileName(fileName),
+LoadGraphStep::LoadGraphStep(const fs::Path& filePath, const std::string& graphName)
+    : _filePath(filePath),
     _graphName(graphName)
 {
 }
 
-LoadGraphStep::LoadGraphStep(const std::string& fileName) 
-    :_fileName(fileName)
+LoadGraphStep::LoadGraphStep(const std::string& graphName)
```
rjb32, line 12: Let's not put both capabilities of load & import into load but really separate the two

storage/dump/GraphInfoLoader.h
```diff
@@ -0,0 +1,84 @@
...
+    [[nodiscard]] DumpResult<void> load(Graph& graph) {
+        _reader.nextPage();
+
+        if (_reader.errorOccured()) {
+            return DumpError::result(DumpErrorType::COULD_NOT_READ_GRAPH_INFO,
+                                     _reader.error().value());
+        }
+
+        // Start reading metadata page
+        auto it = _reader.begin();
```
rjb32: it can be const

system/SystemManager.cpp
```diff
@@ -12,44 +12,98 @@
...
+    if (!_defaultGraph) {
+        throw TuringException("Could not initialise the default graph");
+    }
+}
+
+Graph* SystemManager::loadGraph(const std::string& name) {
+    const fs::Path graphPath = _config->getGraphsDir() / name;
+
+    auto graph = Graph::create(name, graphPath);
+    auto* graphPtr = graph.get();
```
rjb32: Don't use auto here for Graph*

system/SystemManager.cpp
```diff
@@ -12,44 +12,98 @@
...
+    return graphPtr;
+}
+
 Graph* SystemManager::createGraph(const std::string& name) {
-    std::unique_ptr<Graph> graph = Graph::create(name);
-    auto* rawPtr = graph.get();
+    const fs::Path path = _config->getGraphsDir() / name;
+
+    auto graph = Graph::create(name, path);
+    auto* graphPtr = graph.get();
```
rjb32: Same here don't use auto for Graph*

system/SystemManager.h
```diff
@@ -26,15 +27,19 @@ class Change;
...
+    explicit SystemManager(const TuringConfig* config);
     ~SystemManager();
 
     SystemManager(const SystemManager&) = delete;
     SystemManager(SystemManager&&) = delete;
     SystemManager& operator=(const SystemManager&) = delete;
     SystemManager& operator=(SystemManager&&) = delete;
 
-    const TuringConfig& getConfig() const { return _config; }
+    const TuringConfig& getConfig() const { return *_config; }
```
rjb32: If it is a pointer inside, return a pointer

system/SystemManager.h
```diff
@@ -81,7 +84,7 @@ class SystemManager {
 
 
 private:
-    const TuringConfig& _config;
+    const TuringConfig* _config;
```
rjb32: Put _config below the lock to symbolize the locking guarding the entirety of the class.
Default initialize _config pointer to nullptr

pipeline/PipelineStep.cpp
LucLabarriere: Yes I can remove that... It's just for backward compatibility, I'm a bit scared it could break code here and there but I think it's fine. Just wanted to make sure before we definitely remove it

pipeline/operations/ChangeStep.cpp
LucLabarriere: The problem is that making it const disables the move constructor here, which means it has to be copied when returned. Clang-tidy catches that:
image

After experimenting quite a lot with error types like this, I don't think it suits your style of avoiding std::move and automatic copy elision with RVO (At least in all code related to queries), I can rework slightly this file to throw immediately instead of returning result types back to the ChangeStep::execute() function and if you want we can discuss what to do with this error handling mechanism.

storage/Graph.cpp
LucLabarriere: Would it be better a single function:
```c++
static std::unique_ptr<Graph> Graph::create(const std::string& name = "default",
                                            const fs::Path& path = "/dev/null");
```

storage/dump/GraphInfoLoader.h
LucLabarriere: No actually, it's an iterator, it advances x bytes each at each get() call

PR #47, opened 2025-09-23
Fix bug with StringIndexDumper/Loader for large vectors

cyrusknopf:
Graphs as large as Reactome had vectors so large in the StringIndex that they could not fit on a single page. This pull request adds new Load/DumpUtils classes which provide methods for dumping and loading vectors of arbitrary size which may or may not need to span multiple pages.

### DIFF
--- a/io/fs/Primitives.h
+++ b/io/fs/Primitives.h
@@ -20,6 +20,4 @@ concept TrivialPrimitive = std::is_trivial_v<T>
 template <typename T>
 concept TrivialNonCharPrimitive = TrivialPrimitive<T> && !CharPrimitive<T>;
 
-
-
 }


dump/DumpUtils.h
```diff
@@ -0,0 +1,85 @@
+#pragma once
+
+#include "Primitives.h"
+#include "TuringException.h"
+
+#include "DumpConfig.h"
+#include "DumpResult.h"
+#include "FilePageWriter.h"
+
+
```
rjb32, line 9: Remove extra blank line

dump/DumpUtils.h
```diff
@@ -0,0 +1,85 @@
+#pragma once
+
+#include "Primitives.h"
```
rjb32, line 4: Reorder includes. Your header file DumpUtils.h should be first. Then the most important includes. And then the utilities such as TuringException.h

dump/DumpUtils.cpp
```diff
@@ -0,0 +1,20 @@
+#include "DumpUtils.h"
+
+#include "spdlog/fmt/bundled/format.h"
```
rjb32: use the include <> syntax for spdlog as it is an external library

dump/DumpUtils.h
```diff
@@ -0,0 +1,85 @@
...
+    if (TSize > DumpConfig::PAGE_SIZE) {
+        std::string errMsg = fmt::format(
+            "Attempted to dump object {} with size {}, which exceeds page size of {}.",
+            typeid(WorkingT).name(), TSize, DumpConfig::PAGE_SIZE);
+        throw TuringException("Illegal write: " + errMsg);
+    }
+
+    // Write as many as we can on this page
+    const size_t remainingSpace = wr.buffer().avail();
+    const size_t TsThisPage = remainingSpace / TSize;
```
rjb32: start identifier with lowercase

rjb32: Also change the name because TsThisPage is hard to understand what it means

dump/DumpUtils.h
```diff
@@ -0,0 +1,85 @@
...
+
+#include "Primitives.h"
+#include "TuringException.h"
+
+#include "DumpConfig.h"
+#include "DumpResult.h"
+#include "FilePageWriter.h"
+
+
+namespace db {
```
rjb32, line 19: You are in a cpp file, you can do using namespace

dump/LoadUtils.cpp
```diff
@@ -0,0 +1,38 @@
+#include "LoadUtils.h"
```
rjb32, line 1: Insert a blank line between the include

dump/LoadUtils.h
```diff
@@ -0,0 +1,119 @@
...
+
+    // If we all were on a single page: done
+    if (sz == TsThisPage) {
+        return {};
+    }
+
+    const size_t read = TsThisPage;
+    const size_t remaining = sz - read;
+
+    const size_t TsPerPage = DumpConfig::PAGE_SIZE / TSize;
```
rjb32: Use lowercase for identifier

dump/StringIndexerLoader.cpp
```diff
@@ -1,51 +1,28 @@
 #include "StringIndexerLoader.h"
 
 #include <cstdint>
+#include "ID.h"
+#include "metadata/PropertyType.h"
 #include "spdlog/spdlog.h"
```
rjb32: Order of includes! This is all mixed up with system includes

io/fs/FilePageWriter.h
```diff
@@ -63,6 +63,12 @@ class FilePageWriter {
         writeToCurrentPage({reinterpret_cast<const uint8_t*>(&v), sizeof(decltype(v))});
     }
 
+    template <DumpableID IDT>
```
rjb32: To what extent this specialization overlaps with the capabilities of the TrivialPrimitive?

storage/indexes/StringIndex.h
```diff
@@ -61,6 +64,7 @@ class StringIndex {
         }
 
     private:
+        friend StringIndexerLoader;
```
rjb32: Put the friend directive in the public section of the class

LucLabarriere:
This is a breaking change, can you update the UP_TO_DATE_VERSION constant in DumpConfig? You can run `git show --no-patch --format=%at` to get your latest timestamp (run git update first)

io/fs/FilePageWriter.h
cyrusknopf: Please could you rephrase? I'm not sure I understand the question.

io/fs/FilePageWriter.h
```diff
@@ -63,6 +63,12 @@ class FilePageWriter {
...
+    void writeToCurrentPage(IDT v) {
+        writeToCurrentPage(
+            {reinterpret_cast<const uint8_t*>(&v), sizeof(typename IDT::Type)});
```
cyrusknopf: @LucLabarriere 

I added this specialisation to allow the FilePageWriter::writeToCurrentPage to take ID types (NodeID, EdgeID, etc) explicitly, and dump their value. This is needed for the dump/loadVector methods can take a primitive and dump the type directly, or take an ID type and dump the underlying value type.

At the moment, taking the address of the ID type works, because the only non-static member in ID is the underlying value, so for `ID<T>` this is `T _id`. However if in future we change the ID class this might break.

The alternative is to access the private ID value member in this function, which I think would require making FilePageWriter a friend class of ID, or adding a method which returns a reference/pointer to the ID's underlying value (because currently .getValue() returns by value so you cannot take the address of it). I don't really like either of these solutions.

What do you think?

dump/DumpUtils.h
cyrusknopf: Done

dump/DumpUtils.h
cyrusknopf: Done

dump/DumpUtils.cpp
cyrusknopf: Done

dump/DumpUtils.h
cyrusknopf: Changed to countThisPage

dump/DumpUtils.h
cyrusknopf: This is header

dump/LoadUtils.cpp
cyrusknopf: Done

dump/LoadUtils.h
cyrusknopf: Changed to countPerPage

dump/StringIndexerLoader.cpp
cyrusknopf: Fixed

storage/indexes/StringIndex.h
cyrusknopf: Done

cyrusknopf:
> This is a breaking change, can you update the UP_TO_DATE_VERSION constant in DumpConfig? You can run `git show --no-patch --format=%at` to get your latest timestamp (run git update first)

Thanks for pointing this out, done

io/fs/FilePageWriter.h
rjb32: You created this new concept DumpableID no? And you create a new specialization for this concept. To what extent this was not covered by the existing specializations?

io/fs/FilePageWriter.h
rjb32: Why don't we have a public getter in the ID type and require that every dumpable ID type have some getID() function? or something equivalent

LucLabarriere: Ok so I see the problem now.
The core problem is the fact that the fs library explicitly depends on ID now, which honestly is a bit of a code smell. There are similar cases in the STL and other libraries if we think about it. For example:
- We can overload the operator << to allow us to use std::cout << on our own types
- We can define our own std::to_string overloads
- And we can define our own fmt::formatter
- We can define our own std::hash
A good example of this is in the ID.h header:

```c++
template <typename T, int I>
struct std::hash<db::ID<T, I>> {
    size_t operator()(const db::ID<T, I>& id) const {
        return std::hash<size_t> {}(id.getValue());
    }
};

namespace std {

template <typename T, int I>
inline string to_string(db::ID<T, I> id) {
    return to_string(id.getValue());
}

template <typename T, int I>
ostream& operator<<(ostream& os, db::ID<T, I> id) {
    return os << id.getValue();
}

}

template <typename T, int I>
struct fmt::formatter<db::ID<T, I>> : ostream_formatter {};
```

So I suspect this is what we need actually.

We really shouldn't have the fs library depend on the core library, however, it's probably fine if core depends on fs. So I think we can swap the dependencies around. In the FilePageWriter class, I would keep only two definitions of the `writeToCurrentPage` method:

```c++
    void writeToCurrentPage(std::span<const uint8_t> data);

    void writeToCurrentPage(auto&& v) {
        WriteToCurrentPage(*this, std::forward<decltype(v)>(v));
    }
```

If the user calls writeToCurrentPage with a span of uint8_t, the first method is used. Else it's the second method that is used. That method's sole purpose is to forward the call to a static method: WriteToCurrentPage. At the end of the FilePageWriter file we can add basic definitions (I removed the Dumpable version, we don't need it anymore):
```c++
void WriteToCurrentPage(fs::FilePageWriter& w, fs::TrivialPrimitive auto v) {
    w.writeToCurrentPage({reinterpret_cast<const uint8_t*>(&v), sizeof(decltype(v))});
}

template <TrivialPrimitive T, size_t SpanSizeT>
void WriteToCurrentPage(fs::FilePageWriter& w, std::span<T, SpanSizeT> s) {
    w.writeToCurrentPage({reinterpret_cast<const uint8_t*>(s.data()), s.size() * sizeof(T)});
}

template <CharPrimitive T>
void WriteToCurrentPage(fs::FilePageWriter& w, std::basic_string_view<T> str) {
    w.writeToCurrentPage({reinterpret_cast<const uint8_t*>(str.data()), str.size() * sizeof(T)});
}

template <CharPrimitive T>
void WriteToCurrentPage(fs::FilePageWriter& w, const std::basic_string<T>& str) {
    w.writeToCurrentPage(std::basic_string_view<T> {str});
}
```

Then, in order to get the function to work for ID.h, all we have to do is to define the right overload somewhere in the code base (probably in DumpUtils.h)

```c++
// At the top of DumpUtils.h
namespace fs {

template <typename T, int I>
void WriteToCurrentPage(fs::FilePageWriter& w, db::ID<T, I> id) {
    w.writeToCurrentPage({reinterpret_cast<const uint8_t*>(id.getValue()), sizeof(T)});
}

}
```

(Just tested that and it seems to do the trick)

LucLabarriere: In my opinion this is the best fix because we can extend that to any type that we need, such as CommitHash for example

cyrusknopf: I've integrated this: much cleaner and works perfect, thank you.

Only caveat is the ID template calls the non-span variant of FilePageWriter::writeToCurrentPage:

```c++
// At the top of DumpUtils.h
namespace fs {

template <typename T, int I>
void WriteToCurrentPage(fs::FilePageWriter& w, db::ID<T, I> id) {
    w.writeToCurrentPage(reinterpret_cast<const uint8_t*>(id.getValue())); // -> Pass value, do not construct span
}

}
```

dump/DumpUtils.h
```diff
@@ -1,21 +1,67 @@
...
+template <typename T, bool IsID>
+struct WorkingTypeHelper;
+
+// If an ID type: the underlying type to dump is the ::Type property of the ID
+template <typename T>
+struct WorkingTypeHelper<T, true> { using type = typename T::Type; };
+
+// If not an ID type: use the actual type as the type to dump
+template <typename T>
+struct WorkingTypeHelper<T, false> { using type = T; };
```
LucLabarriere, line 51: @cyrusknopf  See if you can remove some of those now with the change with the static function trick

io/fs/FilePageWriter.h
cyrusknopf: This has been removed, so should be resolved

io/fs/FilePageWriter.h
LucLabarriere: Ah that's a mistake actually, sorry! Here I actually think it writes the pointer address in the file.

You either need to call:
```
w.writeToCurrentPage(id.getValue());
```

or you explicitly construct the span. The function calls will be optimized away anyway here.

dump/DumpUtils.h
cyrusknopf: Currently I believe it's still needed on the Load side, as the get() method of the FilePageReader::iterator doesn't use the static function trick, and so is only supported for TrivialPrimitives (for which an ID is not). Therefore we still need the deduction of the WorkingType being either a primitive or ID::Type. I'm trying the static function trick on the reader iterator now

io/fs/FilePageWriter.h
cyrusknopf: Got it. Is the reinterp cast for the purposes of writing spans? I noticed some other methods use the reinterp cast.

io/fs/Primitives.h
```diff
@@ -20,4 +20,6 @@ concept TrivialPrimitive = std::is_trivial_v<T>
 template <typename T>
 concept TrivialNonCharPrimitive = TrivialPrimitive<T> && !CharPrimitive<T>;
 
+
```
rjb32: Too many blank lines

io/fs/FilePageWriter.h
LucLabarriere: Yeah it should be:
```c++
// either
w.writeToCurrentPage(reinterpret_cast<const uint8_t*>(...), nbytes);

// or
w.writeToCurrentPage(std::span { reinterpret_cast<const uint8_t*>(...), nbytes })
```

to write bytes this way

LucLabarriere:
@rjb32 @cyrusknopf
It's fine by me! The fs library does not depend anymore on storage so that's the most important. The dump library does but that's normal

PR #54, opened 2025-10-02
Implement commit-time node and edge deletions with CommitWriteBuffer

cyrusknopf:
[Specification](https://www.notion.so/turingbio/Updates-Deletes-2623aad664c88034acdfd694e811a304)

### DIFF
--- a/query/parser/parser.y
+++ b/query/parser/parser.y
@@ -204,9 +204,6 @@ static db::YParser::symbol_type yylex(db::YScanner& scanner) {
 
 %type<db::QueryCommand*> call_cmd
 
-%type<db::QueryCommand*> s3connect_cmd
-%type<db::QueryCommand*> s3transfer_cmd
-
 %start query_unit
 
 %%
@@ -650,14 +647,6 @@ change_cmd: CHANGE change_subcmd {
 commit_cmd: COMMIT { $$ = CommitCommand::create(ctxt); }
           ;
 
-s3connect_cmd : S3_CONNECT STRING_CONSTANT STRING_CONSTANT STRING_CONSTANT{ $$ = S3ConnectCommand::create(ctxt,$2,$3,$4); }
-          ;
-
-//                       <LOCAL_DIR>     <S3URL>
-s3transfer_cmd : S3_PUSH STRING_CONSTANT STRING_CONSTANT { $$ = S3TransferCommand::create(ctxt,S3TransferCommand::Dir::PUSH,$3,$2); }
-//                       <S3URL>         <LOCAL_DIR>
-               | S3_PULL STRING_CONSTANT STRING_CONSTANT { $$ = S3TransferCommand::create(ctxt,S3TransferCommand::Dir::PULL,$2,$3); }
-          ;
 %%
 
 
--- a/storage/DataPart.cpp
+++ b/storage/DataPart.cpp
@@ -3,7 +3,6 @@
 #include <memory>
 #include <numeric>
 #include <range/v3/action/sort.hpp>
-#include <range/v3/action/stable_sort.hpp>
 #include <range/v3/view/enumerate.hpp>
 #include <range/v3/view/transform.hpp>
 #include <string>
@@ -69,13 +68,10 @@ bool DataPart::load(const GraphView& view, JobSystem& jobSystem, DataPartBuilder
     std::vector<NodeID> tmpNodeIDs(coreNodeLabelSets.size());
     std::iota(tmpNodeIDs.begin(), tmpNodeIDs.end(), firstTmpNodeID);
 
-    // NOTE: With the below change from @ref rg::sort to @ref rg::stable_sort, the
-    // underlying algorithm is changed from a variant of quicksort (in rg::sort) to a
-    // variant of merge sort (rg::stable_sort). rg::stable_sort allocates a large amount
-    // of memory, even in the case of an already-sorted array. In the case of deletions,
-    // it is always the case that the nodes are already sorted. To avoid this guaranteed
-    // extra memory allocation, we add this is_sorted check to reduce the overhead for
-    // deletions.
+    // NOTE: For deletions, @ref coreNodeLabelSets is always sorted. Preliminary check
+    // prevents sorting the already-sorted label sets, which for deletes causes IDs to be
+    // scattered (not uniformally shifted) due to the non-stable implementation of @ref
+    // ranges::sort. Also, often O(n) to avoid O(nlogn).
     bool sorted = rg::is_sorted(rv::zip(coreNodeLabelSets, tmpNodeIDs),
                                 [](const auto& data1, const auto& data2) {
                                     const LabelSetHandle& lset1 = std::get<0>(data1);
@@ -83,9 +79,6 @@ bool DataPart::load(const GraphView& view, JobSystem& jobSystem, DataPartBuilder
                                     return lset1.getID() < lset2.getID();
                                 });
 
-    // NOTE: stable_sort ensures that nodes are mapped to expected of @ref
-    // DataPartModifier::nodeIDMapping, which provides a much more consistent view of
-    // NodeIDs to the user upon deletion.
     if (!sorted) {
         // Sorting based on the labelset
         rg::sort(rv::zip(coreNodeLabelSets, tmpNodeIDs),
--- a/io/fs/Path.cpp
+++ b/io/fs/Path.cpp
@@ -176,15 +176,3 @@ Result<void> Path::rm() const {
     return {};
 }
 
-bool Path::isSubDirectory(const Path& rootDirectory, const Path& subDirectory) {
-    auto normalizedRoot = std::filesystem::absolute(rootDirectory.c_str()).lexically_normal();
-    auto normalizedSub = std::filesystem::absolute(subDirectory.c_str()).lexically_normal();
-
-    // Check if the full path is within the base directory
-    auto relative = normalizedSub.lexically_relative(normalizedRoot);
-
-    // Safe if:
-    // 1. Relative path is not empty
-    // 2. Doesn't start with ".." (going up directories)
-    return !relative.empty() && relative.begin()->string() != "..";
-}
--- a/pipeline/operations/S3PushStep.cpp
+++ b/pipeline/operations/S3PushStep.cpp
@@ -31,7 +31,6 @@ void S3PushStep::prepare(ExecutionContext* ctxt) {
 void S3PushStep::execute() {
     Profile profile {"S3PushStep::execute"};
     if (!_s3File.empty()) {
-
         auto awsRes = _s3Client->uploadFile(_localPath, std::string(_s3Bucket), std::string(_s3File));
         if (!awsRes) {
             throw PipelineException(awsRes.error().fmtMessage());
--- a/query/AST/DeletedIDs.h
+++ b/query/AST/DeletedIDs.h
@@ -17,7 +17,7 @@ public:
     size_t size() { return _deletedIDs.size(); }
 
     void addID(IDT id);
-    
+
     std::vector<IDT>& getIDs() { return _deletedIDs; }
 
 private:
--- a/storage/EdgeContainer.h
+++ b/storage/EdgeContainer.h
@@ -36,10 +36,11 @@ public:
 
     /**
      * @brief Returns the smallest NodeID which this container holds an edge incident to.
-     * @detail Since @ref _outEdges and @ref _inEdges are sorted in NodeID ascending
-     * order, we can make the following claim:
+     * @detail Since @ref _outEdges are sorted in NodeID ascending order, we can make the
+     * following claim:
      * Let NodeID x = min(src of 1st edge in _outEdges, tgt of 1st edge in _inEdges)
-     * then, x is such that for all other edges, e, in this container, the following hold:
+     * then, x is such that for all other edges, e, in this container, the
+     * following hold:
      * a) x <= e.src
      * b) x <= e.tgt
      */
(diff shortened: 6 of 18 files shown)


io/s3/AwsManager.h
```diff
@@ -1,3 +1,4 @@
+#pragma once
```
rjb32, line 1: Add white line before the first include

pipeline/Executor.cpp
```diff
@@ -2,13 +2,17 @@
...
 #include "CallLabelStep.h"
+#include "DeleteStep.h"
 #include "Pipeline.h"
 #include "PipelineException.h"
 #include "PipelineMacros.h"
 #include "Profiler.h"
 
 using namespace db;
 
+using DeleteNodesStep = DeleteStep<NodeID>;
```
rjb32, line 13: Put these into the header of DeleteStep

pipeline/operations/DeleteStep.h
```diff
@@ -0,0 +1,99 @@
...
+
+class ExecutionContext;
+class CommitBuilder;
+
+template <TypedInternalID IDT>
+class DeleteStep {
+public:
+    struct Tag {};
+
+    explicit DeleteStep(std::vector<IDT>&& deletedIDs)
```
rjb32, line 21: Don't pass the vector by move to the constructor of the step! Use an add method somewhere inside the class

query/AST/ASTContext.h
```diff
@@ -52,6 +57,9 @@ class ASTContext {
     void setError(bool hasError) { _isError = hasError; }
 
 private:
+    using DeletedNodes = DeletedIDs<NodeID>;
```
rjb32, line 60: These type aliases should better be in the DeleteIDs header

query/AST/QueryCommand.h
```diff
@@ -84,6 +89,43 @@ class CreateCommand : public QueryCommand {
     ~CreateCommand() override;
 };
 
+// DELETE command
+template <TypedInternalID IDT>
+class DeleteCommand : public QueryCommand {
+public:
+    static DeleteCommand<IDT>* create(ASTContext* ctxt, DeletedIDs<IDT>* deletedIDs);
+
+    std::vector<IDT>& deletions() const { return _deletedIDs->getIDs(); }
```
rjb32, line 98: It's better to have a type alias for the return type of "deletions"

query/AST/QueryCommand.h
```diff
@@ -84,6 +89,43 @@ class CreateCommand : public QueryCommand {
...
+    Kind getKind() const override {
+        if constexpr (std::is_same_v<IDT, NodeID>) {
+            return QueryCommand::Kind::DELETE_NODES_COMMAND;
+        } else {
+            return QueryCommand::Kind::DELETE_EDGES_COMMAND;
+        }
+    }
+
+private:
+    DeletedIDs<IDT>* _deletedIDs;
```
rjb32, line 109: A pointer must always be default-initialized to nullptr

storage/DataPart.cpp
```diff
@@ -68,13 +68,26 @@ bool DataPart::load(const GraphView& view, JobSystem& jobSystem, DataPartBuilder
...
-             [](const auto& data1, const auto& data2) {
-                 const LabelSetHandle& lset1 = std::get<0>(data1);
-                 const LabelSetHandle& lset2 = std::get<0>(data2);
-                 return lset1.getID() < lset2.getID();
-             });
+    // NOTE: For deletions, @ref coreNodeLabelSets is always sorted. Preliminary check
+    // prevents sorting the already-sorted label sets, which for deletes causes IDs to be
+    // scattered (not uniformally shifted) due to the non-stable implementation of @ref
+    // ranges::sort. Also, often O(n) to avoid O(nlogn).
+    bool sorted = rg::is_sorted(rv::zip(coreNodeLabelSets, tmpNodeIDs),
```
rjb32, line 75: sorted can be const

pipeline/operations/DeleteStep.h
```diff
@@ -0,0 +1,99 @@
...
+
+    if constexpr (std::is_same_v<IDT, NodeID>) {
+        _writeBuffer->addDeletedNodes(_deletions);
+    } else if constexpr (std::is_same_v<IDT, EdgeID>) {
+        _writeBuffer->addDeletedEdges(_deletions);
+    }
+
+    if constexpr (std::is_same_v<IDT, NodeID>) {
+        // TODO: ?
+    }
```
LucLabarriere, line 86: Leftover?

pipeline/operations/S3PullStep.cpp
```diff
@@ -1,13 +1,13 @@
 #include "S3PullStep.h"
 
 #include <sstream>
+#include "spdlog/spdlog.h"
```
LucLabarriere, line 4: surround by <> instead of ""

plan/QueryPlanner.h
```diff
@@ -73,6 +73,9 @@ class QueryPlanner {
     void getMatchingLabelSets(std::vector<LabelSetID>& labelSets,
                               const LabelSet* targetLabelSet);
 
+    using DeleteNodesCommand = DeleteCommand<NodeID>;
+    using DeleteEdgesCommand = DeleteCommand<EdgeID>;
```
LucLabarriere, line 77: Really necessary, maybe the templated version is fine? If necessary, put in the QueryCommand header file maybe

query/AST/QueryCommand.h
```diff
@@ -1,11 +1,14 @@
...
 #include <string>
-#include <vector>
 #include <memory>
+#include <type_traits>
 
 #include "Path.h"
+#include "ID.h"
+#include "TypeConstraint.h"
 #include "metadata/PropertyType.h"
+#include "DeletedIDs.h"
```
LucLabarriere, line 11: Forward declare DeletedIDs:
```c++
template <TypedInternalID IDT>
class DeletedIDs;
```

Then, put the implementations of the DeleteCommand methods in the cpp file. In QueryCommand.cpp, add:
```c++
// DeleteCommand

template <TypedInternalID IDT>
DeleteCommand<IDT>::DeleteCommand(DeletedIDs<IDT>* deletions)
    : _deletedIDs(deletions)
{
}

template <TypedInternalID IDT>
DeleteCommand<IDT>* DeleteCommand<IDT>::create(ASTContext* ctxt,
                                               DeletedIDs<IDT>* deletedIDs) {
    DeleteCommand<IDT>* delCmd = new DeleteCommand<IDT>(deletedIDs);
    delCmd->registerCmd(ctxt);
    return delCmd;
}

template <TypedInternalID IDT>
std::vector<IDT>& DeleteCommand<IDT>::deletions() const { return _deletedIDs->getIDs(); }

namespace db {
template class DeleteCommand<NodeID>;
template class DeleteCommand<EdgeID>;
}
```

The explicit instantiation should speed up compilation

query/parser/parser.y
```diff
@@ -595,15 +647,6 @@ change_cmd: CHANGE change_subcmd {
...
 
-s3connect_cmd : S3_CONNECT STRING_CONSTANT STRING_CONSTANT STRING_CONSTANT{ $$ = S3ConnectCommand::create(ctxt,$2,$3,$4); }
-          ;
-
-//                       <LOCAL_DIR>     <S3URL>
-s3transfer_cmd : S3_PUSH STRING_CONSTANT STRING_CONSTANT { $$ = S3TransferCommand::create(ctxt,S3TransferCommand::Dir::PUSH,$3,$2); }
-//                       <S3URL>         <LOCAL_DIR>
-               | S3_PULL STRING_CONSTANT STRING_CONSTANT { $$ = S3TransferCommand::create(ctxt,S3TransferCommand::Dir::PULL,$2,$3); }
-          ;
-
```
LucLabarriere, line 606: Is this normal?

LucLabarriere:
@rjb32 @cyrusknopf

Guys, are we sure this is the design that we want? Like I don't think I will be able to understand anything that occurs there if there's a bug... We have always had a feeling that handling deletes wouldn't be easy but I feel like we're trying to squeeze the whole architecture to support the deletes.

During our call we realised that deletes would be really slow for large graphs, so are you sure you don't want to consider a very simple tombstone implementation (just an additional filter step at query execution)? I feel like this would be a 500 lines of code PR maximum and if that's 10μs cost at query execution that would be super worth it, especially if deletes don't require to rebuild the entire graph everytime. Like deleting a node in CKG (largest graph that we have) with this design would probably take ~1 min and take ~60 GB of extra ram just to store both versions of CKG in ram.

cyrusknopf:
> @rjb32 @cyrusknopf
> 
> Guys, are we sure this is the design that we want? Like I don't think I will be able to understand anything that occurs there if there's a bug... We have always had a feeling that handling deletes wouldn't be easy but I feel like we're trying to squeeze the whole architecture to support the deletes.
> 
> During our call we realised that deletes would be really slow for large graphs, so are you sure you don't want to consider a very simple tombstone implementation (just an additional filter step at query execution)? I feel like this would be a 500 lines of code PR maximum and if that's 10μs cost at query execution that would be super worth it, especially if deletes don't require to rebuild the entire graph everytime. Like deleting a node in CKG (largest graph that we have) with this design would probably take ~1 min and take ~60 GB of extra ram just to store both versions of CKG in ram.

@LucLabarriere I agree it is getting very complex. And that tombstones would be much easier. I am happy to consider the tombstones option honestly, I don't think we should continue this approach if we think it is not sustainable just because I have spent time on it (basically we should not fall for the sunk cost fallacy). 

The tombstone approach is not completely trivial also however (e.g. where we store the tombstone set, etc), but I'm happy to start thinking about an implementation.

@LucLabarriere is your main concern the complexity of the algorithm needed or just the time and memory impact this approach has?

LucLabarriere:
> > @rjb32 @cyrusknopf
> > Guys, are we sure this is the design that we want? Like I don't think I will be able to understand anything that occurs there if there's a bug... We have always had a feeling that handling deletes wouldn't be easy but I feel like we're trying to squeeze the whole architecture to support the deletes.
> > During our call we realised that deletes would be really slow for large graphs, so are you sure you don't want to consider a very simple tombstone implementation (just an additional filter step at query execution)? I feel like this would be a 500 lines of code PR maximum and if that's 10μs cost at query execution that would be super worth it, especially if deletes don't require to rebuild the entire graph everytime. Like deleting a node in CKG (largest graph that we have) with this design would probably take ~1 min and take ~60 GB of extra ram just to store both versions of CKG in ram.
> 
> @LucLabarriere I agree it is getting very complex. And that tombstones would be much easier. I am happy to consider the tombstones option honestly, I don't think we should continue this approach if we think it is not sustainable just because I have spent time on it (basically we should not fall for the sunk cost fallacy).
> 
> The tombstone approach is not completely trivial also however (e.g. where we store the tombstone set, etc), but I'm happy to start thinking about an implementation.

The thing is that, if your implementation works already, maybe this is fine for now. However, we need to make sure we can switch easily to another implementation if needed in the future because, anyway, we identified key issues for large graphs that will have to be addressed

rjb32:
> > > @rjb32 @cyrusknopf
> > > Guys, are we sure this is the design that we want? Like I don't think I will be able to understand anything that occurs there if there's a bug... We have always had a feeling that handling deletes wouldn't be easy but I feel like we're trying to squeeze the whole architecture to support the deletes.
> > > During our call we realised that deletes would be really slow for large graphs, so are you sure you don't want to consider a very simple tombstone implementation (just an additional filter step at query execution)? I feel like this would be a 500 lines of code PR maximum and if that's 10μs cost at query execution that would be super worth it, especially if deletes don't require to rebuild the entire graph everytime. Like deleting a node in CKG (largest graph that we have) with this design would probably take ~1 min and take ~60 GB of extra ram just to store both versions of CKG in ram.
> > 
> > 
> > @LucLabarriere I agree it is getting very complex. And that tombstones would be much easier. I am happy to consider the tombstones option honestly, I don't think we should continue this approach if we think it is not sustainable just because I have spent time on it (basically we should not fall for the sunk cost fallacy).
> > The tombstone approach is not completely trivial also however (e.g. where we store the tombstone set, etc), but I'm happy to start thinking about an implementation.
> 
> The thing is that, if your implementation works already, maybe this is fine for now. However, we need to make sure we can switch easily to another implementation if needed in the future because, anyway, we identified key issues for large graphs that will have to be addressed

rjb32:
@LucLabarriere 

Hi @cyrusknopf, I think it's worth to spend some time today to write a page on notion on how a tombstone implementation could potentially work. I can see 2 main challenges:
* We want to provide snapshot isolation to customers. So submit into head would involve conflict checking. Does it requires the same kind of journal anyway?
* How does a tombstone implementation works with commits and thus concurrency for us? We will need to make it appear as a new commit anyway. Could it be something like: take these data parts + some "deleted" bitmaps? To avoid reconstructing data parts

LucLabarriere:
> > @rjb32 @cyrusknopf
> > Guys, are we sure this is the design that we want? Like I don't think I will be able to understand anything that occurs there if there's a bug... We have always had a feeling that handling deletes wouldn't be easy but I feel like we're trying to squeeze the whole architecture to support the deletes.
> > During our call we realised that deletes would be really slow for large graphs, so are you sure you don't want to consider a very simple tombstone implementation (just an additional filter step at query execution)? I feel like this would be a 500 lines of code PR maximum and if that's 10μs cost at query execution that would be super worth it, especially if deletes don't require to rebuild the entire graph everytime. Like deleting a node in CKG (largest graph that we have) with this design would probably take ~1 min and take ~60 GB of extra ram just to store both versions of CKG in ram.
> 
> @LucLabarriere I agree it is getting very complex. And that tombstones would be much easier. I am happy to consider the tombstones option honestly, I don't think we should continue this approach if we think it is not sustainable just because I have spent time on it (basically we should not fall for the sunk cost fallacy).
> 
> The tombstone approach is not completely trivial also however (e.g. where we store the tombstone set, etc), but I'm happy to start thinking about an implementation.
> 
> @LucLabarriere is your main concern the complexity of the algorithm needed or just the time and memory impact this approach has?

It's a bit both, if the entire database has to be duplicated to delete a single node (or even just 50%), that's already a massive performance cost in terms of CPU and RAM usage, but let's say that's fine. The problem also is that heavily impacts the whole rebase/change submit mechanism, which makes it hard to switch to another implementation in the future. I would have felt more confident with a dumb/quick/easy fix to support deletes by an add-hoc filter step (even if we store a map in dataparts for tombstones), then remove all of that later once we figure out a better way to do that, than the other way around. Again though, maybe it's fine and it requires just a little bit of clean up to make it easy to revert if needed

PR #61, opened 2025-10-06
Add more meaningful error for undefined variables (!analyzer-error-msg)

cyrusknopf:


interpreter/QueryAnalyzer.cpp
```diff
@@ -221,7 +221,9 @@ void QueryAnalyzer::analyzeMatch(MatchCommand* cmd) {
             // Check that the variable exists in the declContext
             VarDecl* decl = declContext->getDecl(name);
             if (!decl) {
-                throw AnalyzeException("Could not get declation context.");
+                std::string error = fmt::format(
```
rjb32: error can be const

cyrusknopf: Added const qualifier

cyrusknopf: @rjb32 Actually cannot be const as it is moved on the line below haha, I removed the const qualifier.

PR #63, opened 2025-10-09
CREATE command support in the Cypher query Analyzer + Plan graph generator

LucLabarriere:
- Split CypherAnalyzer into ReadStmtAnalyzer/WriteStmtAnalyzer/ExprAnalyzer
- Split PlanGraphGenerator into ReadStmtGenerator/WriteStmtGenerator
- Added the CREATE command
- Properties/Labels/EdgeTypes have to exist for read queries. For write queries, they are created if they don't exist
- Support inputs to CREATE commands, in this query:

```
MATCH (n:Person) CREATE (n)-[e:KNOWS]-(m:Person)
```

`n` in an input to the CREATE command

- Added a file based unit test system and a test generator sample `plan2-test-generator`. Just run the sample with query you want to add unit test for, it generates a string `/// QUERY {query} /// RESULT {result}` that can be put in one of the test folders


v2/analyzer/ExprAnalyzer.cpp
```diff
@@ -0,0 +1,410 @@
...
+}
+
+bool ExprAnalyzer::propTypeCompatible(ValueType vt, EvaluatedType exprType) {
+    switch (exprType) {
+        case EvaluatedType::Null:
+        case EvaluatedType::NodePattern:
+        case EvaluatedType::EdgePattern:
+            return false;
+        case EvaluatedType::Integer:
+            return vt == ValueType::Int64 || vt == ValueType::UInt64 || vt == ValueType::Double;
```
cyrusknopf, line 390: What implications does allowing integers to compare against UInts and Doubles have? Does this only apply to CREATE commands?

samples/plan2/query.txt
```diff
@@ -1,13 +1,16 @@
-// MATCH (b)-->(c), (a { name: c.name })-->(b)
-// RETURN a
+// CREATE (a:Person { name: "Luc" });
```
rjb32: Why are all of these commented?

v2/AST/CypherAST.cpp
```diff
@@ -189,3 +190,20 @@ void CypherAST::addNodePatternData(NodePatternData* data) {
 void CypherAST::addEdgePatternData(EdgePatternData* data) {
     _edgePatternDatas.push_back(data);
 }
+
+std::string CypherAST::createErrorString(std::string_view msg, const void* obj) const {
```
rjb32: Why is this put in CypherAST?

v2/analyzer/AnalyzerVariables.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+AnalyzerVariables::AnalyzerVariables(CypherAST* ast)
+    : _ast(ast)
+{
+}
+
+AnalyzerVariables::~AnalyzerVariables() {
+}
+
+VarDecl* AnalyzerVariables::getOrCreateNamedVariable(EvaluatedType type, std::string_view name) {
```
rjb32: I understand the goal for createUnamedVariable but for getOrCreateNamedVariable, is this essentially just a proxy around the DeclContext?

v2/analyzer/CypherAnalyzer.cpp
```diff
@@ -51,49 +62,35 @@ void CypherAnalyzer::analyze() {
 }
 
 void CypherAnalyzer::analyze(const SinglePartQuery* query) {
-    // Read statements
-    for (const Stmt* statement : query->getReadStmts()->stmts()) {
-        if (const MatchStmt* s = dynamic_cast<const MatchStmt*>(statement)) {
-            analyze(s);
-        } else {
-            throwError("Unsupported statement type", statement);
+    const StmtContainer* readStmts = query->getReadStmts();
```
rjb32, line 65: I like it it's very clean, good refactoring

v2/plan/ReadStmtGenerator.h
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include "views/GraphView.h"
```
rjb32: Fix order of includes

v2/AST/stmt/CypherParser.y
rjb32, line 1: The yacc parser has been copied here

v2/plan/nodes/WriteNode.h
```diff
@@ -0,0 +1,79 @@
...
+    };
+
+    struct PendingEdge {
+        EdgeNeighbour _src;
+        EdgeNeighbour _tgt;
+        const EdgePatternData* _data {nullptr};
+    };
+
+    WriteNode()
+        : PlanGraphNode(PlanGraphOpcode::WRITE) {
```
rjb32: Fix formatting of constructor

v2/plan/nodes/FilterNode.h
```diff
@@ -28,19 +28,19 @@ class FilterNode : public PlanGraphNode {
         return _varNode;
     }
 
-    void addPropertyConstraint(PropertyConstraint* constraint) {
-        _propConstraints.push_back(constraint);
+    void addPropertyConstraint(std::unique_ptr<PropertyConstraint> constraint) {
```
rjb32: Where this pointer was owned before?

v2/plan/nodes/WriteNode.h
```diff
@@ -0,0 +1,79 @@
...
+
+class Expr;
+class VarDecl;
+class NodePatternData;
+class EdgePatternData;
+
+class WriteNode : public PlanGraphNode {
+public:
+    struct PendingNode {
+        const NodePatternData* _data {nullptr};
```
cyrusknopf, line 18: Is it fine to store the pattern data (which contains labels as string views) here? I had the problem where the query string dies before I need to write the node, so the label string is not there when I tried reference the Label string view.

v2/AST/decl/PatternData.h
```diff
@@ -1,23 +1,27 @@
...
+
 class PatternData {
 public:
-    using ExprConstraints = std::vector<std::pair<PropertyType, Expr*>>;
+    using ExprConstraints = std::vector<EntityPropertyConstraint>;
 
     const ExprConstraints& exprConstraints() const { return _exprConstraints; }
 
-    void addExprConstraint(PropertyType propType, Expr* expr);
+    void addExprConstraint(std::string_view typeName, ValueType valueType, Expr* expr);
```
rjb32, line 24: I think we need to discuss the impact of using strings and values for types and labels in the plan graph. I can see that it's practical for creates for types and labels yet to be created. But it will make our life harder for subsequent rewriting passes of the plan graph for optimization? If we need to know the cardinality of some labelset or of some property type?

samples/plan2/query.txt
LucLabarriere: They are just query examples, I'm still not quite sure how to handle those query files, they mess up the git diffs

v2/AST/CypherAST.cpp
LucLabarriere: It's a bit tricky to decide what should have the method, CypherAST has a _debugLocation member (std::unordered_map<uintptr_t, void*>) that stores the position of all its AST nodes. So we need a way, from the parser/analyzer/planner to retrieve the AST node position from CypherAST and generate an error from it. With this function any user of the CypherAST's interface can create error messages from the AST nodes, for example for the analyzer:

```c++
void CypherAnalyzer::throwError(std::string_view msg, const void* obj) const {
    throw AnalyzeException(_ast->createErrorString(msg, obj));
}

// Usage:
throwError("SKIP not supported", returnSt);
```

Output:

```
Analyze error:
-------* Query error
     2 | WHERE a:Person AND a.name = "Luc"
     3 | RETURN a SKIP 5
       | ^^^^^^^^^^^^^^^
-------* SKIP not supported
```

But maybe CypherAST should only return the location and the consumers of the API should construct the error themselves

v2/AST/decl/PatternData.h
LucLabarriere: Yeah, this version with std::string_view everywhere is easier to deal with right now, but we need something better for the plan graph rewrites. This comes from the fact that we have a single CypherAST class that we fill with the necessary information at each stage of the query process pipeline. But then the NodePatterns that appear in MATCH commands are actually different from those that appear in the CREATE command (and they will also have to be different from those in MERGE commands). We could make NodePattern and EdgePattern super complex to be able to store the information for any kind of statement but honestly that would be super ugly

```c++
class NodePattern {
private:
    LabelSet* _labelset {nullptr}; // Used if match command
    std::vector<std::string_view> _labels; // Used if create command or merge
    // Same thing for properties, what do we store
    // PropertyTypeID or std::string_view?

    // Maybe use a variant?
};
```

It really feels like a single CypherAST graph is not enough and we need to attach additional information to it throughout the process. Maybe we could add members to CypherAST for the result of the analysis? If we think about it, we have CypherAST which is the result of the parsing, the PlanGraph is the result of the planner, so it feels like the analyzer is also missing some context holder object

```c++
class CypherAST {
private:
    /* ... */
    AnalyzisContext _analysisCtxt; // Stores the labelset constraints for node/edge
                                   // patterns that appear in MATCH commands, as well
                                   // as PropertyTypeID, maybe even information on
                                   // expressions, etc

    PlannerContext _plannerCtxt;   // Not sure that that's needed since the PlanGraph
                                   // should be enough already
};

// Usage of analysisCtxt
AnalysisContext& ctxt = _ast->getAnalysisContext();

// If planning a MATCH statement:
MatchNodeAnalysis* analysis = ctxt.getAnalysis(nodePattern);
if (!analysis) throwError("Not analyzed", nodePattern);
const LabelSet& nodeLabelset = analysis->getLabelSet();

// If planning a CREATE statement:
CreateNodeAnalysis* analysis = ctxt.getAnalysis(nodePattern);
const bool isInput = analysis->isInput();
```

v2/analyzer/AnalyzerVariables.cpp
LucLabarriere: Yeah that's a bit ugly. Basically those functions need to be common to the ReadStmtAnalyzer and WriteStmtAnalyzer. So maybe I can try to put them in decl context directly?

v2/plan/nodes/WriteNode.h
LucLabarriere: In the whole query processing pipeline, std::string_views directly reference the origin query string, so in my opinion that's fine, the whole design has 0 string allocations throughout the process. I really hope we can keep it this way if possible (maybe we will have to switch to strings at some point, but if we do I think we should have one string owner somewhere, and use std::string_views everywhere else to allow to reuse the memory between query executions

v2/plan/nodes/FilterNode.h
LucLabarriere: It was owned by PlanGraphGenerator

v2/analyzer/ExprAnalyzer.cpp
LucLabarriere: That's a good point and I don't think we know exactly what we want to do with that until we implement expression evaluation in the query execution.

Basically, the idea is that our current Cypher parser v2 has different type semantics than C++. It's as opposed to our v1 parser that differentiates int, uint and double (I think?). In v2 it's just Integer (signed or unsigned) or Double.

That means that technically, you could do WHERE a.age > 14.5 and, if age is a uint in the database, then that query works. To handle that in the execution, it means you either have to generate a filter step where you compare uint64_t to doubles, or that you add a cast step double to uint for the lhs. That's definitely worth discussing at some point.

However, I think we really want to allow people to write these kinds of queries because they are super convenient from a user standpoint (it's like python's implicit handling of integer vs float type management, and javascript's super abstract "Number" type)

v2/AST/decl/PatternData.h
LucLabarriere: Btw I just realized that in the PlanGraph, the information is stored as a LabelSet and a PropertyTypeID in FilterNode, so I actually think it's fine, it's just that the type inferring is delayed to the planner instead of analyzer

v2/plan/nodes/WriteNode.h
cyrusknopf: Yes thats my concern: the string_view is from the query string, but does the point where the PendingNode get used outlive the query string? In my case it did: you write the pending node at Change::commit/submit time, which outlives the CREATE query the string was allocated in, and therefore the string_view is invalid.

I'm not sure if your use case for the PendingNode differs to mine, though.

LucLabarriere: Yeah here it should be fine, we just need to make the query string outlive the pipeline generation (not execution) (and if we don't want that at some point, we can allocate the strings somewhere and keep the string_views everywhere.

Indeed, This WriteNode is a plan node that allows to generate the appropriate WriteStep that will be ran by the query executor. It's the write step that fills the CommitWriteBuffer, so if we need to switch to std::string, we can still do it there, (but I also think we can skip that to the point where execute the query and fill the CommitWriteBuffer

v2/analyzer/ExprAnalyzer.cpp
cyrusknopf: Yeah I agree we should do whatever is most convenient for the user

v2/plan/nodes/WriteNode.h
cyrusknopf: Got it

LucLabarriere:
@rjb32 I have one issue with the latest commit. Moving the getOrCreateNamedVariable method to the DeclContext had the side effect that I cannot use AnalyzerException in the definition (since DeclContext is in the AST library, not anlyzer). And I'm not sure how to fix this. I think the based solution is to derive ParserException, AnalyzerException and PlannerException from a common base class CypherException, and to catch CypherException in all parts of the process instead of catching the individual exceptions:

```c++
try {
    parser.parse(query);
} catch (const CypherException& e) {
    std::cerr << "Syntax error:\n"
              << e.what() << std::endl;
    return;
}

CypherAnalyzer analyzer(&ast, view);
try {
    analyzer.analyze();
} catch (const CypherException& e) {
    std::cerr << "Analyze error:\n"
              << e.what() << std::endl;
    return;
}

PlanGraphGenerator planGen(ast, view);
try {
    planGen.generate(ast.queries().front());
} catch (const CypherException& e) {
    std::cerr << "Plan error:\n"
              << e.what() << std::endl;
    return;
}
```

In fact, it's not particularly useful to have one type of exception in this example since they can't all be thrown at the same time. Having separate exception only make sense if we catch all of them from a single try:

```c++
try {
    interpreter.process(queryStr); // does parsing/analyzing/planning
} catch (const ParserException& e) {
    // parser error
} catch (const AnalyzerException& e) {
    // analyzer error
} catch (const PlannerException& e) {
    // planner error
}
```

The inheritance technic would allow to have both code styles

v2/AST/CypherAST.cpp
rjb32: LLVM people have solved this by recognizing that there is a need in the architecture for a component in charge of this, which is the DiagnosticsManager, that will handle the formatting and creation of error messages. And the SourceManager which is in charge of mapping source locations to AST nodes.

https://www.goldsborough.me/c++/clang/llvm/tools/2017/02/24/00-00-06-emitting_diagnostics_and_fixithints_in_clang_tools/
https://clang.llvm.org/doxygen/classclang_1_1SourceManager.html

It's good to know that this is a need that a larger compiler needs to work on at some point.

For now, I would like to separate the error mapping (which is stored CypherAST) from the formatting of the error.

v2/AST/decl/PatternData.h
rjb32: Yes I agree that we can attach some AnalysisInfo* to relevant CypherAST nodes. And store them in an AnalysisContext residing in CypherAST as you shown earlier.

v2/plan/nodes/FilterNode.h
rjb32: I would like to have a better to manage this unique_ptr. I know I am the one that used this style in the PlanGraph originally but I am somewhat regretting my choice. I wonder if we would just be better by having central ownership by PlanGraph acting as a context

v2/plan/nodes/WriteNode.h
```diff
@@ -0,0 +1,80 @@
...
+        size_t _offset {0};
+    };
+
+    class EdgeNeighbour : public std::variant<const VarDecl*, size_t> {
```
rjb32: Please do not derive from std::variant directly. I find that it's cleaner to have some wrapper class or struct that puts a more meaningful name on what the base class represent.

rjb32:
I think that you use ASTException today? To me it's fine.

But I agree that a CypherException or a CompilerException would be good.

> @rjb32 I have one issue with the latest commit. Moving the getOrCreateNamedVariable method to the DeclContext had the side effect that I cannot use AnalyzerException in the definition (since DeclContext is in the AST library, not anlyzer). And I'm not sure how to fix this. I think the based solution is to derive ParserException, AnalyzerException and PlannerException from a common base class CypherException, and to catch CypherException in all parts of the process instead of catching the individual exceptions:
> 
> ```c++
> try {
>     parser.parse(query);
> } catch (const CypherException& e) {
>     std::cerr << "Syntax error:\n"
>               << e.what() << std::endl;
>     return;
> }
> 
> CypherAnalyzer analyzer(&ast, view);
> try {
>     analyzer.analyze();
> } catch (const CypherException& e) {
>     std::cerr << "Analyze error:\n"
>               << e.what() << std::endl;
>     return;
> }
> 
> PlanGraphGenerator planGen(ast, view);
> try {
>     planGen.generate(ast.queries().front());
> } catch (const CypherException& e) {
>     std::cerr << "Plan error:\n"
>               << e.what() << std::endl;
>     return;
> }
> ```
> 
> In fact, it's not particularly useful to have one type of exception in this example since they can't all be thrown at the same time. Having separate exception only make sense if we catch all of them from a single try:
> 
> ```c++
> try {
>     interpreter.process(queryStr); // does parsing/analyzing/planning
> } catch (const ParserException& e) {
>     // parser error
> } catch (const AnalyzerException& e) {
>     // analyzer error
> } catch (const PlannerException& e) {
>     // planner error
> }
> ```
> 
> The inheritance technic would allow to have both code styles

v2/AST/SourceManager.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <cstdint>
+#include <string_view>
+#include <unordered_map>
+
+#include "SourceLocation.h"
+
+namespace db::v2 {
+
+class SourceManager {
```
rjb32, line 11: I like it, it looks good

v2/plan/PlannerException.h
```diff
@@ -2,14 +2,14 @@
 
 #include <string>
 
-#include "TuringException.h"
+#include "ASTException.h"
 
 namespace db::v2 {
 
-class PlannerException : public TuringException {
+class PlannerException : public ASTException {
```
rjb32: Sorry I think there was a misunderstanding. Can you change the name of ASTException into CompilerException? Because it feels weird to have PlannerException that derives from ASTException

PR #65, opened 2025-10-14
DELETE and DETACH DELETE commands in the query planner V2

LucLabarriere:


v2/AST/expr/ExprChain.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <vector>
+
+namespace db::v2 {
+
+class CypherAST;
+class Expr;
+
+class ExprChain {
```
cyrusknopf, line 10: Does the "Chain" in "ExprChain" mean anything specific? Or can this just be thought of a vector of Exprs?

v2/plan/nodes/WriteNode.h
```diff
@@ -65,31 +65,43 @@ class WriteNode : public PlanGraphNode {
...
+    std::span<const PendingEdge> pendingEdges() const { return _newEdges; }
+    std::span<const VarDecl* const> toDeleteNodes() const { return _toDeleteNodes; }
+    std::span<const VarDecl* const> toDeleteEdges() const { return _toDeleteEdges; }
 
 private:
-    std::vector<PendingNode> _nodes;
-    std::vector<PendingEdge> _edges;
+    std::vector<PendingNode> _newNodes;
+    std::vector<PendingEdge> _newEdges;
+    std::vector<const VarDecl*> _toDeleteNodes;
```
cyrusknopf, line 108: Is there any reason you think we need to store VarDecls here and not just NodeID/EdgeID?

LucLabarriere: Yes! I had a feeling you would notice that ahah, it's particularly tricky for you since you worked on the WriteStep of v1.

Basically, this a "Plan" node that is meant to generate the write step node for the execution. It's not meant to store the result of the match queries or anything like that, it's just saying: "According to the query I parsed, here are the nodes and edges that are meant to be created/deleted". And those "nodes" and "edges" only reference variables in the query, not actual entities in the database.

We basically say: we want to create node(s) 'n' that have these label and property constraints. This may result in creating 1 or more nodes during the pipeline execution (it creates a column of nodeIDs that have been created). For example, this query:

```
MATCH (a:Person) CREATE (a)-[e:LIVES_IN]->(c:City { name: "New York" })
```

Will:
- Match each node (a) Person in the database
For each Person,
- Create an edge from (a) to a new node (c) City

So that plan graph "WriteNode" node just stores that information:
- __node__ (0:City {name (String): "New York" })
- __edge__ (a)-[0:LIVES_IN]->(0)

Meaning:
- For each node Person in the database, create a node City "New York"
- And create an edge between the input node (a) to the created node (at offset 0 = first node in the pending nodes vector)

This query is actually bad, and it should be MERGE written with a MERGE instead:

```
MATCH (a:Person) MERGE (a)-[e:LIVES_IN]->(c:City { name: "New York" })
```

In this case, only one node "New York" is created, and all nodes Person are connected to it

v2/AST/expr/ExprChain.h
LucLabarriere: No it does not mean anything specific, it's just a vector of Exprs. It's mostly used to better represent an actual ASTNode as defined in the grammar file

```
ExprChain
    : Expr
    | ExprChain COMMA Expr
    ;
```

v2/AST/expr/ExprChain.h
```diff
@@ -0,0 +1,29 @@
...
+public:
+    ~ExprChain();
+
+    static ExprChain* create(CypherAST* ast);
+
+    void add(Expr* expr) { _exprs.push_back(expr); }
+
+    const std::vector<Expr*>& getExprs() const { return _exprs; }
+
+    std::vector<Expr*>::const_iterator begin() const { return _exprs.begin(); }
```
rjb32: Just expose a span or just a const std::vector<>& to return a range. Use a type alias with "using" to simplify the return type

v2/AST/expr/ExprChain.h
rjb32: This is a good style to wrap an STL container, std::vector here, into a semantically meaningful class, ExprChain. So that we know as a reader what it means, more than just "vector".
It's probably even better to add a comment in the header file to explain in which grammar rule it is used.

v2/plan/nodes/WriteNode.h
```diff
@@ -65,31 +65,43 @@ class WriteNode : public PlanGraphNode {
...
+    }
+
     const PendingNode& getPendingNode(size_t offset) const {
-        return _nodes.at(offset);
+        return _newNodes.at(offset);
     }
 
-    std::span<const PendingNode> pendingNodes() const { return _nodes; }
-    std::span<const PendingEdge> pendingEdges() const { return _edges; }
+    std::span<const PendingNode> pendingNodes() const { return _newNodes; }
```
rjb32: It's worth defining type aliases for these spans

v2/AST/expr/ExprChain.h
cyrusknopf: Cool

v2/plan/nodes/WriteNode.h
cyrusknopf: Haha yes, I keep getting confused. But here I meant the DeletedNodes. Since something that is deleted needs to exist, it must have an ID. So why the indirection VarDecl* -> _entityID?

But the CREATE/MERGE example makes sense

cyrusknopf: Ah actually I misunderstood what `_entityID` was! I now know it is only used for injecting IDs, never mind my question, I get it now.

LucLabarriere: Oh yeah, notice that the VarDecl in v2 is not the same one as v1 (it doesn't have the ability to inject node IDs yet). Injection of IDs will definitely be an interesting topic too

PR #67, opened 2025-10-15
Add conflict checking and Tombstones at the Commit level to facilitate deletes

cyrusknopf:
Requires #78

- Adds Tombstones to CommitData which  denote deleted nodes and edges
- When creating a new commit, the tombstones are copied over
- DELETE NODES/EDGES commands added, which add to the CommitWriteBuffer
- On CommitWriteBuffer flush, deleted nodes and edges are turned into tombstones

- Adds ChangeConflictChecker which is invoked at Change::rebase to compare the write sets of the commits that have been made on "main" since a change branched
- Rejects in the result of a write conflict

### DIFF
--- a/pipeline/operations/DeleteStep.cpp
+++ b/pipeline/operations/DeleteStep.cpp
@@ -14,9 +14,9 @@ namespace db {
 }
 
 template <TypedInternalID IDT>
-void DeleteStep<IDT>::addDeletions(std::vector<IDT>&& deletedIDs) {
+void DeleteStep<IDT>::addDeletions(const std::vector<IDT>& deletedIDs) {
     bioassert(_deletions.empty());
-    _deletions = std::move(deletedIDs);
+    _deletions = deletedIDs;
 }
 
 template <TypedInternalID IDT>
--- a/pipeline/operations/DeleteStep.h
+++ b/pipeline/operations/DeleteStep.h
@@ -23,7 +23,7 @@ public:
     using DeleteNodesStep = DeleteStep<NodeID>;
     using DeleteEdgesStep = DeleteStep<EdgeID>;
 
-    void addDeletions(std::vector<IDType>&& deletedIDs);
+    void addDeletions(const std::vector<IDType>& deletedIDs);
 
     void prepare(ExecutionContext* ctxt);
 
--- a/query/AST/DeletedIDs.h
+++ b/query/AST/DeletedIDs.h
@@ -19,7 +19,7 @@ public:
 
     void addID(IDT id);
 
-    std::vector<IDT>& getIDs() { return _deletedIDs; }
+    const std::vector<IDT>& getIDs() { return _deletedIDs; }
 
 private:
     friend ASTContext;
--- a/query/AST/QueryCommand.cpp
+++ b/query/AST/QueryCommand.cpp
@@ -66,7 +66,7 @@ DeleteCommand<IDT>* DeleteCommand<IDT>::create(ASTContext* ctxt,
 }
 
 template <TypedInternalID IDT>
-DeleteCommand<IDT>::Deletions& DeleteCommand<IDT>::deletions() const {
+const DeleteCommand<IDT>::Deletions& DeleteCommand<IDT>::deletions() const {
     return _deletedIDs->getIDs();
 }
 
--- a/query/AST/QueryCommand.h
+++ b/query/AST/QueryCommand.h
@@ -97,7 +97,7 @@ public:
     using Deletions = std::vector<IDT>;
     static DeleteCommand<IDT>* create(ASTContext* ctxt, DeletedIDs<IDT>* deletedIDs);
 
-    Deletions& deletions() const;
+    const Deletions& deletions() const;
 
     Kind getKind() const override;
 
--- a/plan/QueryPlanner.cpp
+++ b/plan/QueryPlanner.cpp
@@ -205,14 +205,14 @@ bool QueryPlanner::planCreate(const CreateCommand* createCmd) {
 }
 
 bool QueryPlanner::planDeleteNodes(const DeleteCommand<NodeID>* delCmd) {
-    std::vector<NodeID>& deletions = delCmd->deletions();
+    const DeleteCommand<NodeID>::Deletions& deletions = delCmd->deletions();
 
     _pipeline->add<StopStep>();
 
     DeleteStep<NodeID>& step =
         _pipeline->add<DeleteStep<NodeID>>().get<DeleteStep<NodeID>>();
 
-    step.addDeletions(std::move(deletions));
+    step.addDeletions(deletions);
 
     _pipeline->add<EndStep>();
 
@@ -220,14 +220,14 @@ bool QueryPlanner::planDeleteNodes(const DeleteCommand<NodeID>* delCmd) {
 }
 
 bool QueryPlanner::planDeleteEdges(const DeleteCommand<EdgeID>* delCmd) {
-    std::vector<EdgeID>& deletions = delCmd->deletions();
+    const DeleteCommand<EdgeID>::Deletions& deletions = delCmd->deletions();
 
     _pipeline->add<StopStep>();
 
     DeleteStep<EdgeID>& step =
         _pipeline->add<DeleteStep<EdgeID>>().get<DeleteStep<EdgeID>>();
 
-    step.addDeletions(std::move(deletions));
+    step.addDeletions(deletions);
 
     _pipeline->add<EndStep>();
 


pipeline/Executor.cpp
```diff
@@ -2,13 +2,17 @@
...
 #include "CallLabelStep.h"
+#include "DeleteStep.h"
 #include "Pipeline.h"
 #include "PipelineException.h"
 #include "PipelineMacros.h"
 #include "Profiler.h"
 
 using namespace db;
 
+using DeleteNodesStep = DeleteStep<NodeID>::DeleteNodesStep;
```
cyrusknopf, line 13: I have to redefine this alias here because the macro preprocessor can't handle <> or ::, so I cannot use the templated version or the alias in the DeleteStep namespace in the ACTIVATE_PTR, etc. macros

query/AST/ASTContext.h
```diff
@@ -62,6 +68,8 @@ class ASTContext {
     std::vector<EntityPattern*> _entityPatterns;
     std::vector<TypeConstraint*> _typeConstraints;
     std::vector<InjectedIDs*> _injectedIDs;
+    std::vector<DeletedIDs<NodeID>*> _deletedNodeIDs;
```
cyrusknopf, line 71: I cannot use the alias DeletedNodes or DeletedEdges here without explicitly including DeletedIDs instead of forward declaring it

storage/versioning/Commit.h
```diff
@@ -14,6 +16,9 @@ class FrozenCommitTx;
 
 class Commit {
 public:
+    using CommitVector = std::vector<std::unique_ptr<Commit>>;
```
cyrusknopf, line 19: I moved these definitions here because I needed them in more places than just VersionController (where they were previously) and did not want to include VersionController everywhere

pipeline/operations/DeleteStep.h
```diff
@@ -0,0 +1,45 @@
...
+    struct Tag {};
+
+    explicit DeleteStep()
+    {
+    }
+
+    using DeleteNodesStep = DeleteStep<NodeID>;
+    using DeleteEdgesStep = DeleteStep<EdgeID>;
+
+    void addDeletions(std::vector<IDT>&& deletedIDs);
```
rjb32: Use something more meaningful for IDT like IDType

pipeline/operations/DeleteStep.cpp
```diff
@@ -0,0 +1,77 @@
...
+using namespace db;
+
+namespace db {
+    template class DeleteStep<NodeID>;
+    template class DeleteStep<EdgeID>;
+}
+
+template <TypedInternalID IDT>
+void DeleteStep<IDT>::addDeletions(std::vector<IDT>&& deletedIDs) {
+    bioassert(_deletions.empty());
```
rjb32, line 18: Is it still necessary? If not you can put in header

pipeline/operations/DeleteStep.cpp
```diff
@@ -0,0 +1,77 @@
...
+    _commitBuilder = tx.commitBuilder();
+    _writeBuffer = &_commitBuilder->writeBuffer();
+}
+
+template <TypedInternalID IDT>
+void DeleteStep<IDT>::execute() {
+    GraphReader reader = _commitBuilder->readGraph();
+
+    for (IDT id : _deletions) {
+        bool has {false};
```
rjb32: Choose a more meaningful variable name for has

query/AST/DeletedIDs.h
```diff
@@ -0,0 +1,48 @@
...
+
+    size_t size() { return _deletedIDs.size(); }
+
+    void addID(IDT id);
+
+    std::vector<IDT>& getIDs() { return _deletedIDs; }
+
+private:
+    friend ASTContext;
+    std::vector<IDT> _deletedIDs;
```
rjb32, line 26: Define a DeletedIDs type alias

query/AST/QueryCommand.h
```diff
@@ -84,6 +90,23 @@ class CreateCommand : public QueryCommand {
     ~CreateCommand() override;
 };
 
+// DELETE command
+template <TypedInternalID IDT>
+class DeleteCommand : public QueryCommand {
+public:
+    static DeleteCommand<IDT>* create(ASTContext* ctxt, DeletedIDs<IDT>* deletedIDs);
+
+    std::vector<IDT>& deletions() const;
```
rjb32: Define type alias to make it more elegant

storage/versioning/Change.cpp
```diff
@@ -92,7 +104,17 @@ CommitResult<void> Change::rebase([[maybe_unused]] JobSystem& jobsystem) {
...
-    // For each of the commits to build...
+    // Get the commits that occured on main since branch time
+    const Commit::CommitSpan commitsSinceBranch =
+        _versionController->getCommitsSinceCommitHash(baseHash());
+
+    // Check the write buffer for each commit to be made for write conflicts
+    rebaser.checkConflicts(commitsSinceBranch);
+
+    // If the above call to @ref ChangeRebaser::checkConflicts does not throw, then there
+    // are no write conflicts, and we may safely proceed with rebasing - involving
```
rjb32, line 115: Good explaination it's quite clear

storage/versioning/ChangeConflictChecker.cpp
```diff
@@ -0,0 +1,173 @@
...
+    : _change(change),
+    _entityIDRebaser(entityRebaser),
+    _commitsSinceBranch(commits)
+{
+}
+
+void ChangeConflictChecker::getWritesSinceCommit(ConflictCheckSets& writes) {
+    for (const std::unique_ptr<Commit>& commit : _commitsSinceBranch) {
+        const CommitHistory& history = commit->history();
+        const CommitJournal& journal = history.journal();
```
rjb32, line 31: There is something a bit unclear here: if we say that each CommitHistory is the history of all predecessors of a commit since beginning of time, what does history.journal contains for each commit? Are we sure that we are not iterating over more history that we need to?

storage/versioning/ChangeConflictChecker.cpp
```diff
@@ -0,0 +1,173 @@
...
+        bioassert(&journal);
+        WriteSet<NodeID>::setUnion(writes.writtenNodes, journal.nodeWriteSet());
+        WriteSet<EdgeID>::setUnion(writes.writtenEdges, journal.edgeWriteSet());
+    }
+}
+
+void ChangeConflictChecker::checkConflicts() {
```
rjb32, line 38: In principle the high level algorithmic structure is clear

pipeline/operations/DeleteStep.cpp
cyrusknopf: Yep this is still necessary

pipeline/operations/DeleteStep.cpp
cyrusknopf: Changed to `existsAndNotDeleted`

query/AST/DeletedIDs.h
cyrusknopf: `DeletedIDs` is the name of the wrapper class of this vector. Do you think its still necessary to type alias the underlying container?

query/AST/QueryCommand.h
cyrusknopf: Added `using Deletions = std::vector<IDT>`

storage/versioning/ChangeConflictChecker.cpp
cyrusknopf: The journal in commit X contains the diff (created edges/nodes, deleted edges/nodes) between commit X-1 and commit X. So in order to check all the diffs, we have to iterate over all commits since branch time to see all the diffs since brand time.

pipeline/operations/DeleteStep.h
cyrusknopf: Changed to IDType

pipeline/operations/DeleteStep.cpp
```diff
@@ -0,0 +1,77 @@
...
+    if constexpr (std::is_same_v<IDT, NodeID>) {
+        _writeBuffer->addDeletedNodes(_deletions);
+    } else if constexpr (std::is_same_v<IDT, EdgeID>) {
+        _writeBuffer->addDeletedEdges(_deletions);
+    }
+
+    if constexpr(std::is_same_v<IDT, NodeID> && DETACH) {
+        _writeBuffer->addHangingEdges(reader.getView());
+    }
+}
```
LucLabarriere, line 67: I think it's really nice, it should be straight forward to convert for the query pipeline v2

pipeline/operations/DeleteStep.h
```diff
@@ -0,0 +1,45 @@
...
+    void reset() {};
+
+    bool isFinished() const { return true; };
+
+    void execute();
+
+    void describe(std::string& descr) const;
+
+private:
+    std::vector<IDType> _deletions;
```
LucLabarriere, line 39: Quick question (no change required, just for my curiosity), is there a reason this is a vector and not a Column? In v2, the delete column will be able to read columns from the current block (MATCH (n) DELETE n) so it will have to take columns as input

query/AST/QueryCommand.cpp
```diff
@@ -43,6 +45,46 @@ CreateCommand* CreateCommand::create(ASTContext* ctxt, CreateTargets* targets) {
     return cmd;
 }
 
+// DeleteCommand
+template <TypedInternalID IDT>
+DeleteCommand<IDT>::DeleteCommand(DeletedIDs<IDT>* deletions)
+    : _deletedIDs(deletions) {
```
LucLabarriere: braces on next line

query/AST/QueryCommand.cpp
```diff
@@ -43,6 +45,46 @@ CreateCommand* CreateCommand::create(ASTContext* ctxt, CreateTargets* targets) {
...
+        return QueryCommand::Kind::DELETE_EDGES_COMMAND;
+    }
+}
+
+namespace db {
+    template class DeleteCommand<NodeID>;
+    template class DeleteCommand<EdgeID>;
+    template class DeletedIDs<NodeID>;
+    template class DeletedIDs<EdgeID>;
+}
```
LucLabarriere, line 87: Indention, I think namespace contents are not indented

storage/reader/GraphReader.h
```diff
@@ -56,12 +56,14 @@ class GraphReader {
...
-    template <SupportedType T>
-    [[nodiscard]] const T::Primitive* tryGetNodeProperty(PropertyTypeID ptID, NodeID nodeID) const;
+    template <SupportedType T> [[nodiscard]] const T::Primitive* tryGetNodeProperty(PropertyTypeID ptID,
+                                                                                    NodeID nodeID) const;
 
-    template <SupportedType T>
-    [[nodiscard]] ScanNodePropertiesRange<T> scanNodeProperties(PropertyTypeID ptID) const {
+    template <SupportedType T> [[nodiscard]] ScanNodePropertiesRange<T> scanNodeProperties(PropertyTypeID ptID) const {
         return ScanNodePropertiesRange<T>(_view, ptID);
     }
```
LucLabarriere: template <> on previous line

pipeline/operations/DeleteStep.h
cyrusknopf: No specific reason, should be trivial to convert it into a column

query/AST/QueryCommand.cpp
cyrusknopf: Thanks, fixed

query/AST/QueryCommand.cpp
cyrusknopf: Yep you're right, fixed

storage/reader/GraphReader.h
cyrusknopf: Oops, fixed

query/AST/DeletedIDs.h
rjb32: No indeed, sorry my bad I was looking very quickly.

rjb32: Are you sure this is best to return an std::vector as a non-const reference?

cyrusknopf:
```cpp
bool QueryPlanner::planDeleteNodes(const DeleteCommand<NodeID>* delCmd) {
    std::vector<NodeID>& deletions = delCmd->deletions(); // This calls getIDs, which returns the non-const vector reference

    _pipeline->add<StopStep>();

    DeleteStep<NodeID>& step =
        _pipeline->add<DeleteStep<NodeID>>().get<DeleteStep<NodeID>>();

    step.addDeletions(std::move(deletions));

    _pipeline->add<EndStep>();

    return true;
}
```

It is used here, to move the deleted IDs from the DeleteCommand to the step. I thought this was better from copying from the Command to the Step, since the IDs are never used by the Command after this.

If you prefer I can make it return a const reference and then copy the IDs into the DeleteStep instead of moving?

PR #69, opened 2025-10-15
Fixed Neo4j import deadlock

LucLabarriere:
Fixed a not too obvious deadlock that would occur when import graphs from neo4j dumps. Was caused by this stack:
- fromDumpFileToJsonDir()
- lock(_mutex)
- fromUrlToJsonDir()
- lock(_mutex) // deadlock


rjb32:
Good find!

PR #70, opened 2025-10-16
Fixed type constraints, Renamed NodeLabelExpr to EntityTypeExpr, allows to apply additional type constraints on edges

LucLabarriere:


v2/AST/expr/EntityTypeExpr.h
```diff
@@ -0,0 +1,37 @@
...
+class EntityTypeExpr : public Expr {
+public:
+    using Types = std::vector<Symbol*>;
+
+    Symbol* getSymbol() const { return _symbol; }
+    const Types& getTypes() const { return _types; }
+
+    static EntityTypeExpr* create(CypherAST* ast, 
+                                 Symbol* symbol,
+                                 Types&& types);
```
rjb32: I would prefer to not rely on move semantics as a matter of style. Here I would say it's ok because it prevents a copy. But I don't want it to proliferate in the codebase.

v2/AST/SymbolChain.h
```diff
@@ -0,0 +1,40 @@
...
+namespace db::v2 {
+
+class CypherAST;
+class Symbol;
+
+class SymbolChain {
+public:
+    using SymbolVector = std::vector<Symbol*>;
+
+    ~SymbolChain();
```
rjb32: The destructor is now public? I think we used the friend technique for the other AST nodes.
That allows you to also take out the copy constructors versions if both constructor & destructor are private

v2/AST/EdgePattern.h
```diff
@@ -24,19 +21,19 @@ class EdgePattern : public EntityPattern {
...
 
-    const EdgeTypes& types() const { return _types; }
+    const SymbolChain* types() const { return _types; }
 
     EdgePatternData* getData() const { return _data; }
 
     void setDirection(Direction direction) { _direction = direction; }
 
-    void setTypes(EdgeTypes&& types) { _types = std::move(types); }
+    void setTypes(SymbolChain* types) { _types = types; }
```
rjb32, line 30: That looks much better

PR #73, opened 2025-10-20
Generalise dumpVector to dumpRange (!dump-load-range)

cyrusknopf:
- Dump function to dump a range instead of vector.
    - Load function is still specialised to container

Verified with

`./build/samples/graph-loader-dumper/run.sh` (including Reactome)

### DIFF
--- a/storage/dump/DumpUtils.h
+++ b/storage/dump/DumpUtils.h
@@ -1,13 +1,13 @@
 #pragma once
 
+#include <ranges>
+
 #include "ID.h"
 #include "Primitives.h"
-#include "DumpConfig.h"
-#include "DumpResult.h"
 #include "FilePageWriter.h"
+#include "DumpResult.h"
+#include "DumpConfig.h"
 #include "TuringException.h"
-#include <concepts>
-#include <ranges>
 
 namespace fs {
 


rjb32:
It looks good to me. Any reason to want to generalize to any dumpable range?

rjb32:
Makes sense in the context of dumping the CommitJournal.

rjb32:
@LucLabarriere Please take a look

cyrusknopf:
> It looks good to me. Any reason to want to generalize to any dumpable range?
> Makes sense in the context of dumping the CommitJournal.

Yep, it's because dumping any container is pretty generalisable: iterate through it and dump each element. I need to be able to dump other containers other than vectors for Tombstones, CommitJournal, etc.

LucLabarriere:
Looks good to me, it's a bit sad that the whole dump library now includes `<ranges>` though. But I think it makes sense for convenience. If someway we realise that it's used by 3-4 containers at most then it might make sense to just write down each function version instead of relying on templates with ranges (especially for when we rework the dump/load format).

rjb32:
Thinking about it indeed, it makes sense in the context of the dump library to specialize for each container so that the available specializations are stated explicitly. 

We don't really want on too much smart mechanisms in the dump library that would somewhat hide all the constructs that it has to effectively dump.

rjb32:
This is actually just a PR for a dumping utility in DumpUtils.h for a function which dumps any class with a range interface with begin() and end(). And @cyrusknopf uses only the <range> library to check that the template parameter type of the Range satisfies indeed the requirements of input range. 
And the Range template parameter is enforced to only be a range of IDs or trivially dumpable types, as per the dumpable concept defined earlier in DumpUtils.h
So to me it makes sense given that WriteSets are dumped explicitly in CommitJournalDumper and thus there is a well-identified place in the code where this is dumped.

PR #74, opened 2025-10-20
Add serialisation of the CommitJournal

cyrusknopf:
Depends on #73 

- Adds serialisation of CommitJournal:
    - Journal is dumped per-commit
    - Journal is loaded per-commit
- Adds CommitView comparators to GraphComparator to ensure CommitJournals are the same
- Adds two CommitJournalSerialisation tests to check the expected journal is present after dump and load

### DIFF
--- a/storage/dump/DumpConfig.h
+++ b/storage/dump/DumpConfig.h
@@ -11,7 +11,7 @@ class DumpConfig {
 public:
     static constexpr uint32_t ONE_BAD_CAFE = 0x1BADCAFE;
     static constexpr uint64_t VERSION = HEAD_COMMIT_TIMESTAMP;
-    static constexpr uint64_t UP_TO_DATE_VERSION = 1758794298;
+    static constexpr uint64_t UP_TO_DATE_VERSION = 1761128034;
     static constexpr uint64_t PAGE_SIZE = fs::DEFAULT_PAGE_SIZE;
 
     static constexpr size_t SIZEOF_ONE_BAD_CAFE = sizeof(decltype(ONE_BAD_CAFE));


storage/versioning/CommitJournal.h
```diff
@@ -32,14 +32,22 @@ class CommitJournal {
     auto& nodeWriteSet() { return _nodeWriteSet; }
     auto& edgeWriteSet() { return _edgeWriteSet; }
 
+    const auto& nodeWriteSet() const { return _nodeWriteSet; }
```
rjb32: Use WriteSet<NodeID> instead of auto for the return type

storage/versioning/WriteSet.h
```diff
@@ -36,11 +42,15 @@ class WriteSet {
                                          const WriteSet<IDT>& set2) = delete;
 
 private:
+    friend class CommitJournal;
+
     std::vector<IDT> _set;
+
+    auto& getRaw() { return _set; }
```
rjb32: Don't use auto for return type

storage/versioning/CommitJournal.h
```diff
@@ -32,14 +32,22 @@ class CommitJournal {
     auto& nodeWriteSet() { return _nodeWriteSet; }
```
rjb32: Don't use auto for return type

storage/versioning/CommitJournal.h
```diff
@@ -32,14 +32,22 @@ class CommitJournal {
...
+    friend class CommitJournalLoader;
+
     bool _initialised {false};
 
     WriteSet<NodeID> _nodeWriteSet;
     WriteSet<EdgeID> _edgeWriteSet;
 
     CommitJournal() = default;
+
+    auto& rawNodeWriteSet() { return _nodeWriteSet.getRaw(); }
```
rjb32: Are the raw versions of the methods really necessary at this level?

storage/dump/CommitJournalDumper.h
```diff
@@ -0,0 +1,23 @@
...
+#include "FilePageWriter.h"
+#include "dump/DumpResult.h"
+
+namespace db {
+
+class CommitJournal;
+
+class CommitJournalDumper {
+public:
+    explicit CommitJournalDumper(fs::FilePageWriter& writer)
```
rjb32, line 15: FilePageWriter is only used by reference in the dumper header

storage/dump/CommitJournalLoader.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "FilePageReader.h"
```
rjb32: FilePageReader is only used by reference is the header file

storage/dump/CommitDumper.cpp
```diff
@@ -95,6 +101,23 @@ DumpResult<void> CommitDumper::dump(const Commit& commit, const fs::Path& path)
         }
     }
 
+    // Dumping Journal
+    {
+        Profile profile {"CommitDumper::dump <journal>"};
+        const fs::Path journalPath = path / "journal";
+
+        auto writerRes = fs::FilePageWriter::open(journalPath, DumpConfig::PAGE_SIZE);
```
rjb32, line 109: writerRes can be const

storage/dump/CommitJournalDumper.cpp
```diff
@@ -0,0 +1,35 @@
...
+DumpResult<void> CommitJournalDumper::dump(const CommitJournal& journal) {
+
+    GraphDumpHelper::writeFileHeader(_writer);
+
+    { // Dump node write set
+        const WriteSet<NodeID>& writtenNodes = journal.nodeWriteSet();
+        const size_t nodesSize = writtenNodes.size();
+
+        DumpUtils::ensureDumpSpace(sizeof(nodesSize), _writer);
+        _writer.writeToCurrentPage(writtenNodes.size());
```
rjb32: Should you reuse use nodesSize here?

storage/dump/CommitJournalDumper.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+        DumpUtils::dumpRange(writtenNodes, _writer);
+    }
+
+    { // Dump edge write set
+        const WriteSet<EdgeID>& writtenEdges = journal.edgeWriteSet();
+        const size_t edgesSize = writtenEdges.size();
+
+        DumpUtils::ensureDumpSpace(sizeof(edgesSize), _writer);
+        _writer.writeToCurrentPage(writtenEdges.size());
```
rjb32: Should you reuse edgesSize here?

storage/dump/CommitDumper.cpp
cyrusknopf: That makes the FilePageWriter value const which is not valid for the constructor for the Dumper

storage/dump/CommitJournalDumper.cpp
cyrusknopf: Yep thanks!

storage/versioning/CommitJournal.h
cyrusknopf: Fixed

storage/versioning/WriteSet.h
cyrusknopf: Fixed

storage/versioning/CommitJournal.h
cyrusknopf: Fixed

storage/versioning/CommitJournal.h
cyrusknopf: My logic was this allows CommitJournalLoader to be a friend of CommitJournal, instead of a friend of WriteSet, which I thought was cleaner

storage/dump/CommitJournalDumper.h
cyrusknopf: Fixed with forward declare

storage/dump/CommitJournalLoader.h
cyrusknopf: Fixed with forward declare

storage/dump/CommitJournalDumper.cpp
cyrusknopf: Fixed

storage/versioning/CommitJournal.h
rjb32: Indeed it is cleaner. Does CommitJournalLoader really need to do getRaw() on the writeSet instead of just getting a range via begin and end methods?

storage/comparators/CommitViewComparator.cpp
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        const WriteSet<EdgeID> edgeWriteSetA = journalA.edgeWriteSet();
+        const WriteSet<EdgeID> edgeWriteSetB = journalB.edgeWriteSet();
+
+        if (!WriteSetComparator<EdgeID>::same(edgeWriteSetA, edgeWriteSetB)) {
+            spdlog::error("CommitJournal Edge WriteSets are not the same.");
+            return false;
+        }
+    }
```
LucLabarriere, line 30: Why the outer braces?

storage/comparators/WriteSetComparator.cpp
```diff
@@ -0,0 +1,39 @@
...
+namespace db {
+    template class WriteSetComparator<NodeID>;
+    template class WriteSetComparator<EdgeID>;
+}
+
+template <TypedInternalID IDT>
+bool WriteSetComparator<IDT>::same(const WriteSet<IDT> setA , const WriteSet<IDT> setB) {
+    if (setA.size() != setB.size()) {
+        spdlog::error(
+            "Node WriteSets in CommitJournal have sizes {} and {}, respecitvely.",
```
LucLabarriere: respectively

storage/comparators/WriteSetComparator.cpp
```diff
@@ -0,0 +1,39 @@
...
+        return false;
+    }
+
+    auto itA = setA.begin();
+    auto itB = setB.begin();
+    size_t index = 0;
+    while (itA != setA.end() && itB != setB.end()) {
+        if (*itA != *itB) {
+            spdlog::error(
+                "Node WriteSets have values {} and {}, respecitvely, at index {}", *itA,
```
LucLabarriere: respectively

storage/comparators/WriteSetComparator.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "ID.h"
+#include "versioning/WriteSet.h"
+namespace db {
```
LucLabarriere, line 6: blank line above namespace maybe

storage/versioning/CommitJournal.h
```diff
@@ -3,9 +3,7 @@
 #include "versioning/WriteSet.h"
 #include "ID.h"
 
-namespace db {
-
-class CommitWriteBuffer;
+namespace db { class CommitWriteBuffer;
```
LucLabarriere: Is it written like this:

```c++
namespace { class CommitWriteBuffer;
```

Or is my diffs view that is messed up?

storage/versioning/CommitJournal.h
```diff
@@ -15,6 +13,9 @@ class CommitWriteBuffer;
  */
 class CommitJournal {
 public:
+    using NodeWriteSet = WriteSet<NodeID>;
```
LucLabarriere, line 18: @rjb32 Is this a good use case for type aliases here? It echoes what Suhas said with the fact that `using` hides information to the user just like what "auto" does. I fill like `WriteSet<NodeID>` is clearer than `NodeWriteSet`, but maybe that's just me

test/serialisation/CommitJournalSerialisationTest.cpp
```diff
@@ -0,0 +1,231 @@
...
+        _config.setSyncedOnDisk(false);
+        _workingPath = fs::Path {_outDir + "/testfile"};
+        _env = TuringTestEnv::create(_workingPath);
+        _config.setTuringDirectory(_workingPath);
+        _db = std::make_unique<TuringDB>(&_config);
+        _db->run();
+
+        SystemManager& sysMan = _db->getSystemManager();
+        _builtGraph = sysMan.createGraph("simple");
+        SimpleGraph::createSimpleGraph(_builtGraph);
```
LucLabarriere, line 32: Do you need the _db member? _env should have everything you need

test/serialisation/CommitJournalSerialisationTest.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+        // XXX: Need to remove the directory created in TuringTest.h SetUp() has
+        // GraphDumper requires the directory does not exist
+        if (FileUtils::exists(_workingPath.filename())) {
+            FileUtils::removeDirectory(_workingPath.filename());
+        }
```
LucLabarriere: If you inherit from TuringTest, it should do the cleanups automatically normally

LucLabarriere:
The changes look good to me, the only part that I'm not a fan of is the test/serialization part. I think the raw dump tests should be in test/storage/dump, they should only link storage and examples, and test only The individual Dumper and Loaders. Then if you want to have a more general test/serialization library, that should not test the individual dump/load components but just test the dump/load features:

- Create a test env with TuringTestEnv::createSyncedOnDisk()
- Create graphs, make new commits etc, then create a new testenv that tries to work with the dumped graph and see if the graphs are in a valid state. (It should not use Dump/LoadGraph, but only use the automatic SystemManager dump mechanism).

What do you think? The test/serialization library was already there so maybe we can keep it as is for now and change later on, in another PR when we have time

storage/versioning/CommitJournal.h
rjb32: It depends on the usage and it's true that it hides a bit of information. And it boils down to esthetic preference in the end. In this case this is acceptable because NodeWriteSet is exactly the name of the template specialization of "WriteSet<NodeID>" in words.

storage/comparators/CommitViewComparator.cpp
cyrusknopf: Removed

storage/comparators/WriteSetComparator.cpp
cyrusknopf: haha thanks fixed

storage/comparators/WriteSetComparator.cpp
cyrusknopf: fixed

storage/comparators/WriteSetComparator.h
cyrusknopf: Fixed

storage/versioning/CommitJournal.h
cyrusknopf: Nope you are right, fixed

storage/versioning/CommitJournal.h
cyrusknopf: Yep I thought since the type of NodeWriteSet is obvious to anyone who knows what a WriteSet is, that it was clearer without losing any information

cyrusknopf:
> The changes look good to me, the only part that I'm not a fan of is the test/serialization part. I think the raw dump tests should be in test/storage/dump, they should only link storage and examples, and test only The individual Dumper and Loaders. Then if you want to have a more general test/serialization library, that should not test the individual dump/load components but just test the dump/load features:
> 
> * Create a test env with TuringTestEnv::createSyncedOnDisk()
> * Create graphs, make new commits etc, then create a new testenv that tries to work with the dumped graph and see if the graphs are in a valid state. (It should not use Dump/LoadGraph, but only use the automatic SystemManager dump mechanism).
> 
> What do you think? The test/serialization library was already there so maybe we can keep it as is for now and change later on, in another PR when we have time

I agree it would be nice to split. The problem I had was that I need to do both:

1. make queries to the DB, because I need to do CREATE NODES etc. to test out the creation of the journal via the CommitWriteBuffer (for context: GraphWriter/anything other than a create query does not use a write buffer and therefore does not create a journal entry)
2. inspect the internal state of the CommitJournal, which is not exposed to the user at all (meaning I cannot access it with a query, and have to have access to the DB object itself)

Can this be done just by linking to testenv?

test/serialisation/CommitJournalSerialisationTest.cpp
cyrusknopf: Good catch, removed it and use the _env DB member

test/serialisation/CommitJournalSerialisationTest.cpp
cyrusknopf: Thanks, removed this check

LucLabarriere:
> > The changes look good to me, the only part that I'm not a fan of is the test/serialization part. I think the raw dump tests should be in test/storage/dump, they should only link storage and examples, and test only The individual Dumper and Loaders. Then if you want to have a more general test/serialization library, that should not test the individual dump/load components but just test the dump/load features:
> > 
> > * Create a test env with TuringTestEnv::createSyncedOnDisk()
> > * Create graphs, make new commits etc, then create a new testenv that tries to work with the dumped graph and see if the graphs are in a valid state. (It should not use Dump/LoadGraph, but only use the automatic SystemManager dump mechanism).
> > 
> > What do you think? The test/serialization library was already there so maybe we can keep it as is for now and change later on, in another PR when we have time
> 
> I agree it would be nice to split. The problem I had was that I need to do both:
> 
> 1. make queries to the DB, because I need to do CREATE NODES etc. to test out the creation of the journal via the CommitWriteBuffer (for context: GraphWriter/anything other than a create query does not use a write buffer and therefore does not create a journal entry)
> 2. inspect the internal state of the CommitJournal, which is not exposed to the user at all (meaning I cannot access it with a query, and have to have access to the DB object itself)
> 
> Can this be done just by linking to testenv?

Yes the testenv library provides everything needed to run queries, but honestly don't worry about it. It was more a general comment on our unit tests organization. Unit testing dump/load is super tricky at the moment anyway

cyrusknopf:
> > > The changes look good to me, the only part that I'm not a fan of is the test/serialization part. I think the raw dump tests should be in test/storage/dump, they should only link storage and examples, and test only The individual Dumper and Loaders. Then if you want to have a more general test/serialization library, that should not test the individual dump/load components but just test the dump/load features:
> > > 
> > > * Create a test env with TuringTestEnv::createSyncedOnDisk()
> > > * Create graphs, make new commits etc, then create a new testenv that tries to work with the dumped graph and see if the graphs are in a valid state. (It should not use Dump/LoadGraph, but only use the automatic SystemManager dump mechanism).
> > > 
> > > What do you think? The test/serialization library was already there so maybe we can keep it as is for now and change later on, in another PR when we have time
> > 
> > 
> > I agree it would be nice to split. The problem I had was that I need to do both:
> > 
> > 1. make queries to the DB, because I need to do CREATE NODES etc. to test out the creation of the journal via the CommitWriteBuffer (for context: GraphWriter/anything other than a create query does not use a write buffer and therefore does not create a journal entry)
> > 2. inspect the internal state of the CommitJournal, which is not exposed to the user at all (meaning I cannot access it with a query, and have to have access to the DB object itself)
> > 
> > Can this be done just by linking to testenv?
> 
> Yes the testenv library provides everything needed to run queries, but honestly don't worry about it. It was more a general comment on our unit tests organization. Unit testing dump/load is super tricky at the moment anyway

Okay no problem, I'll add a Jira to look at reorganising this whole test/serialisation suite with your dependency concerns in mind

storage/dump/CommitJournalDumper.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "GraphDumpHelper.h"
+#include "DumpUtils.h"
+#include "versioning/CommitJournal.h"
+#include "versioning/WriteSet.h"
+
+using namespace db;
+
+DumpResult<void> CommitJournalDumper::dump(const CommitJournal& journal) {
+
```
rjb32: Spurious blank line

storage/comparators/WriteSetComparator.cpp
```diff
@@ -0,0 +1,39 @@
+#include "WriteSetComparator.h"
+
+#include <spdlog/spdlog.h>
+
+#include "ID.h"
+
+using namespace db;
+
+namespace db {
+    template class WriteSetComparator<NodeID>;
```
rjb32: No indentation inside namespace

storage/versioning/WriteSet.h
rjb32: Not fixed

PR #76, opened 2025-10-21
Turing 851: Fixed joins on filter nodes. Also fixed loops that appeared when placing constraints

LucLabarriere:
This PR fixes two issues:
- Merging more than two branches on a node `(a)-->(x), (b)-->(x), (c)-->(x)`
This now properly generates two joins, (a)->(J1), (b)->(J1)->(J2), (c)->(J2)->(x)

- Loops with WHERE predicates and property constraints. Examples:
```
MATCH (a)-->(x), (b)-->(x) WHERE b.name = a.name
// And
MATCH (a)-->(x), (b { name: a.name })-->(x)
```

The new algorithm now finds a common successor between all dependencies, and places the constraint there (in this example, on `x`)

I'm not a huge fan on how this is handled right now. The tricky thing is that PropertyExpr constraints `(a { name: "sth" })` and WHERE predicates, should be strictly equivalent, but right now they are handled slightly differently. Indeed, a WHERE predicate is an expression that has to be evaluated to a boolean, whereas a PropertyConstraint is composed of two parts: the property type, and an expression that has to be evaluated, then compared to the property value of the input. This is a side effect of the current implementation and we could convert the PropertyConstraint to a regular WHERE predicate (they mean exactly the same thing), but that would mean creating dummy ASTNodes, which should not be done.

For the moment, I cleaned up the code as much as possible to highlight the fact they are both treated by similar algorithms. Once we work on expression evaluates in the pipeline, I think it will be a good time to revisit this and unify the way we treat them


v2/plan/ExprDependencies.h
```diff
@@ -79,6 +85,22 @@ class ExprDependencies {
         }
     }
 
+    const VarNode* findCommonSuccessor(const VarNode* var) const {
+        for (const auto& dep : _dependencies) {
+            const auto* successor = PlanGraphTopology::findCommonSuccessor(var, dep._var);
```
rjb32: The common successor should be cached for variables, it can potentially be invoked for several predicates with the same variables

v2/plan/ExprDependencies.h
```diff
@@ -79,6 +85,22 @@ class ExprDependencies {
         }
     }
 
+    const VarNode* findCommonSuccessor(const VarNode* var) const {
+        for (const auto& dep : _dependencies) {
```
rjb32: Don't use auto here and use the explicit type for dep and successor

v2/plan/ExprDependencies.h
```diff
@@ -79,6 +85,22 @@ class ExprDependencies {
...
+            const auto* successor = PlanGraphTopology::findCommonSuccessor(var, dep._var);
+
+            if (successor) {
+                var = PlanGraphTopology::findNextVar(successor);
```
rjb32: What does this mean exactly?

v2/plan/PlanGraphTopology.cpp
```diff
@@ -11,6 +13,16 @@ PlanGraphTopology::PathToDependency PlanGraphTopology::getShortestPath(const Pla
...
+    // 2. Explore the graph breadth-first from all encountered nodes, going downward
+    //
+    // If target is found in phase 1: BackwardPath
+    // Else if target is found in phase 2: UndirectedPath
+    // Else: NoPath
+    //
+    // TODO: Store the PlanGraphTopology as member, to allow to reuse the queues and hash maps
+
     std::queue<const PlanGraphNode*> phase1;
     std::queue<const PlanGraphNode*> phase2;
```
rjb32: Do we really need to do it with 2 queues? We could do it in one phase and pass if we go up/down/both when we add the node to the queue?

v2/plan/ReadStmtGenerator.cpp
```diff
@@ -268,7 +264,6 @@ VarNode* ReadStmtGenerator::generatePatternElementTarget(VarNode* prevNode,
         std::tie(var, filter) = _variables->createVarNodeAndFilter(decl);
         currentNode->connectOut(filter);
     } else {
-        incrementDeclOrders(prevNode->getDeclOrder(), filter);
         currentNode->connectOut(filter);
 
         // Detect loops
```
rjb32, line 270: Why do it only in one specific place here? I would think it's most efficient to do it afterward once the entire PlanGraph is generated

v2/plan/PlanGraphTopology.cpp
```diff
@@ -103,6 +123,13 @@ PlanGraphNode* PlanGraphTopology::getBranchTip(PlanGraphNode* origin) {
...
 bool PlanGraphTopology::detectLoops(const PlanGraphNode* origin) {
+    // Algorithm:
+    // 1. Explore the graph breadth-first from the origin node, going upwards
+    // 2. If we encounter origin again, we have a loop 
+    //
+    // Note: The algorithm works because is not added to the visited set at the
+    //       beginning of the algorithm
+
     std::unordered_set<PlanGraphNode*> visited;
     std::queue<PlanGraphNode*> q;
```
rjb32, line 105: This is detecting only loops starting at origin. This is a bit confusion for a function called detectLoops. 
Loops are most typically detected using DFS.

v2/plan/ReadStmtGenerator.cpp
LucLabarriere: That's for a later implementation, the detectLoops algorithm (renamed detectLoopsFrom), is an algorithm that starts at an origin node and goes backward to see if we see the origin again (if we do, it's a loop).

The idea is that, later on, we can detect the loop immediately and do loop unrolling here. The detection is only performed for path patterns, when we connect to an already existing var node. Example:

```
(a)-->(b)-->(a)
```

when we connect to (a) the second time around, we see that it already exists so we check if we just formed a loop, if we did, we can apply loop unrolling (in a later PR):

```
(a)-->(b)-->(a') WHERE a == a'.
```

Doing it right away like this allows for everything that comes after in during the PlanGraphGeneration step to know that there are no loops (algorithms can be adapted and simplified based on that knowledge)

v2/plan/PlanGraphTopology.cpp
LucLabarriere: Renamed detectLoopsFrom

v2/plan/ExprDependencies.h
LucLabarriere: That's part of the design that I don't like for now, the find common successor finds a connexion point later on in the PlanGraph where we know that the two variables are defined (they both have a column that define them in the current block). However, we need to put the evaluation of the predicate/constraint on the next FilterNode that we find. So we find the next VarNode after the successor, we get its filter, and put the constraint/predicate evaluation here. This is clearly not great and I think it should be cleaned up/reworked once we settle on how we generate the processors that evaluate the expressions

v2/plan/PlanGraphTopology.cpp
LucLabarriere: With an `std::queue<std::pair<const PlanGraphNode*, Direction>>` queue?

rjb32: Yes exactly

v2/plan/PlanGraphTopology.cpp
rjb32: Ok sounds good

v2/plan/ReadStmtGenerator.cpp
rjb32: Ok sounds good it will be handled later

PR #77, opened 2025-10-21
Turing-848,849: Relative paths for import graph and s3 operations, all must be within the .turing/data folder

LucLabarriere:
- Re-add @suhassn-turing-bio changes that were previously reverted on S3 operations (mostly isSubdirectory check). Remaining missing features will be added by the S3_BACKUP and S3_RESTORE PR

### DIFF
--- a/io/fs/Path.cpp
+++ b/io/fs/Path.cpp
@@ -188,7 +188,7 @@ Result<void> Path::mkdir() const {
     try {
         std::filesystem::create_directories(_path);
     } catch (const std::filesystem::filesystem_error& e) {
-        return Error::result(ErrorType::CANNOT_MKDIR);
+        return Error::result(ErrorType::CANNOT_MKDIR, e.code().value());
     }
 
     return {};
--- a/io/fs/Path.h
+++ b/io/fs/Path.h
@@ -35,7 +35,6 @@ public:
     [[nodiscard]] bool isSubDirectory(const Path& root) const;
     [[nodiscard]] Result<void> toCanonical() ;
 
-
     friend Path operator/(const Path& lhs, const Path& rhs) {
         std::string p = lhs._path + "/";
         p += rhs._path;
--- a/test/io/ByteBufferIteratorTest.cpp
+++ b/test/io/ByteBufferIteratorTest.cpp
@@ -65,3 +65,8 @@ TEST_F(ByteBufferIteratorTest, General) {
     ASSERT_TRUE(it.get<char>(11) == "Hello world"sv);
 }
 
+int main(int argc, char** argv) {
+    return turing::test::turingTestMain(argc, argv, [] {
+        testing::GTEST_FLAG(repeat) = 4;
+    });
+}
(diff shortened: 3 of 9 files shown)


io/fs/Path.cpp
```diff
@@ -83,6 +83,38 @@ Result<FileInfo> Path::getFileInfo() const {
...
+        const auto relative = canonicalSub.lexically_relative(canonicalRoot);
+
+        return !relative.empty() && relative.begin()->string() != "..";
+    } catch (const std::filesystem::filesystem_error& e) {
+        return false;
+    }
+}
+
+Path Path::parent() const {
+    const auto pos = _path.find_last_of('/');
```
rjb32: Could you use something like https://en.cppreference.com/w/cpp/filesystem/path/parent_path.html

pipeline/operations/S3PullStep.cpp
```diff
@@ -32,14 +32,24 @@ void S3PullStep::prepare(ExecutionContext* ctxt) {
...
+
+    const fs::Path& dataDir = _sysMan->getConfig()->getDataDir();
+    const fs::Path absolute = dataDir / _localPath;
+
+    if (!absolute.isSubDirectory(dataDir)) {
+        throw PipelineException("Invalid file path.");
+    }
+
+    const fs::Path parent = absolute.parent();
+    parent.mkdir();
```
rjb32, line 44: You could have an already exists error no?

query/AST/QueryCommand.h
```diff
@@ -313,17 +313,17 @@ class S3TransferCommand : public QueryCommand {
         return _s3File;
     }
 
-    static S3TransferCommand* create(ASTContext* ctx, Dir _transferDir, const std::string& s3URL, const std::string& localDir);
+    static S3TransferCommand* create(ASTContext* ctx, Direction _transferDir, const std::string& s3URL, const std::string& localDir);
```
rjb32: _transferDir argument starts with underscore

system/SystemManager.cpp
```diff
@@ -145,31 +145,42 @@ void SystemManager::listGraphs(std::vector<std::string_view>& names) {
 bool SystemManager::importGraph(const std::string& graphName, const fs::Path& filePath, JobSystem& jobSystem) {
     const fs::Path graphPath = _config->getGraphsDir() / filePath;
 
-    // Check if graph was already loaded || is already loading
+    // Step 1. Check if graph was already loaded || is already loading
```
rjb32, line 148: I like the steps comments it's quite clear

pipeline/operations/S3PullStep.cpp
LucLabarriere: Yes, in this case mkdir returns an error which I thought could be ignored. However, I just realized that it does not create the whole folder tree:
```python3
turingdb.transfer("file.csv", "turingdb://file.csv") # Works
turingdb.transfer("file.csv", "turingdb://subfolder/file.csv") # Works too.
                                                               # Creates subfolder if
                                                               # it does not exist
turingdb.transfer("file.csv", "turingdb://sub/sub/folder/file.csv") # Does not work
```

What do you think should be the behaviour? Similar to `mkdir -p`? Throw if the directory does not exist?

rjb32: I think it's interesting to create the folder hierarchy.
Also, can we transfer directories recursively?

LucLabarriere: That's a good point, I'll check

PR #78, opened 2025-10-22
Add goToPart for Iterators

cyrusknopf:
- Rename `PartIterator::isValid` to `PartIterator::isNotEnd`, as there is conflated use of the word "valid" in iterators; where a `PartIterator` is "valid" if it is not equal to `end()`, whilst a `ScanNodesIterator` (for example) is "valid" if its current `PartIterator` is pointing to a DataPart which has nodes (for example: equivalent case for `ScanEdges`, `GetOutEdges`, etc.)

- Add `advanceToPart` and `goToPart` to `GetInEdgesIterator` and `GetOutEdgesIterator` to allow for much more efficient conflict checking of tombstones

- Add tests for `goToPart` for both the above iterators

### DIFF
--- a/storage/iterators/GetInEdgesIterator.cpp
+++ b/storage/iterators/GetInEdgesIterator.cpp
@@ -56,7 +56,7 @@ void GetInEdgesIterator::goToPart(size_t partIdx) {
 
 void GetInEdgesIterator::advancePartIterator(size_t n) {
     // Advance n dataparts forward
-    for (; n > 0 && _partIt.isNotEnd(); n--) {
+    for (size_t i = 0; i < n && _partIt.isNotEnd(); i++) {
         _partIt.next();
     }
     // If we have not reached the end, update the _node members
--- a/storage/iterators/GetOutEdgesIterator.cpp
+++ b/storage/iterators/GetOutEdgesIterator.cpp
@@ -56,7 +56,7 @@ void GetOutEdgesIterator::goToPart(size_t partIdx) {
 
 void GetOutEdgesIterator::advancePartIterator(size_t n) {
     // Advance n dataparts forward
-    for (; n > 0 && _partIt.isNotEnd(); n--) {
+    for (size_t i = 0; i < n && _partIt.isNotEnd(); i++) {
         _partIt.next();
     }
     // If we have not reached the end, update the _node members


storage/iterators/GetOutEdgesIterator.cpp
```diff
@@ -49,6 +49,32 @@ void GetOutEdgesIterator::next() {
...
 }
 
+void GetOutEdgesIterator::goToPart(size_t partIdx) {
+    Iterator::reset();
+    advancePartIterator(partIdx);
+}
+
+void GetOutEdgesIterator::advancePartIterator(size_t n) {
+    // Advance n dataparts forward
+    for (; n > 0 && _partIt.isNotEnd(); n--) {
```
rjb32: Style: define an integer variable i for the for loop instead of modifying the argument

storage/iterators/GetInEdgesIterator.cpp
```diff
@@ -49,6 +49,32 @@ void GetInEdgesIterator::next() {
...
 }
 
+void GetInEdgesIterator::goToPart(size_t partIdx) {
+    Iterator::reset();
+    advancePartIterator(partIdx);
+}
+
+void GetInEdgesIterator::advancePartIterator(size_t n) {
+    // Advance n dataparts forward
+    for (; n > 0 && _partIt.isNotEnd(); n--) {
```
rjb32: Style: don't modify the function argument but use a loop index i

rjb32:
> * Rename `PartIterator::isValid` to `PartIterator::isNotEnd`, as there is conflated use of the word "valid" in iterators; where a `PartIterator` is "valid" if it is not equal to `end()`, whilst a `ScanNodesIterator` (for example) is "valid" if its current `PartIterator` is pointing to a DataPart which has nodes (for example: equivalent case for `ScanEdges`, `GetOutEdges`, etc.)
> * Add `advanceToPart` and `goToPart` to `GetInEdgesIterator` and `GetOutEdgesIterator` to allow for much more efficient conflict checking of tombstones
> * Add tests for `goToPart` for both the above iterators

I am not sure the rename of isValid into isNotEnd is warranted here because is it not the same concept but applied to another kind of iterator? Because in general isValid means can I dereference the iterator and does it point to valid data. So for PartIterator it should be when we have no more DataPart to iterate on. For ScanNodesIterator it should be when there is no more nodes to iterate on.

cyrusknopf:
> > * Rename `PartIterator::isValid` to `PartIterator::isNotEnd`, as there is conflated use of the word "valid" in iterators; where a `PartIterator` is "valid" if it is not equal to `end()`, whilst a `ScanNodesIterator` (for example) is "valid" if its current `PartIterator` is pointing to a DataPart which has nodes (for example: equivalent case for `ScanEdges`, `GetOutEdges`, etc.)
> > * Add `advanceToPart` and `goToPart` to `GetInEdgesIterator` and `GetOutEdgesIterator` to allow for much more efficient conflict checking of tombstones
> > * Add tests for `goToPart` for both the above iterators
> 
> I am not sure the rename of isValid into isNotEnd is warranted here because is it not the same concept but applied to another kind of iterator? Because in general isValid means can I dereference the iterator and does it point to valid data. So for PartIterator it should be when we have no more DataPart to iterate on. For ScanNodesIterator it should be when there is no more nodes to iterate on.

Makes sense, if you would like to keep the existing name I can revert the name change. Although when you  say
"For ScanNodesIterator it should be when there is no more nodes to iterate on."
do you mean no more nodes in the current datapart, or in the entire graph?

rjb32:
The rename for isValid() is relative to the fact that isValid() is about checking if PartIterator is at the end, where it conflicts with the action of the nextValid() function in higher level iterators such as GetOutEdgesIterator where GetOutEdgesIterator::nextValid() moves if there are no more out edges to get in the current DataPart, not necessarily if we are at the end of data parts.

storage/iterators/GetOutEdgesIterator.cpp
cyrusknopf: Fixed

storage/iterators/GetInEdgesIterator.cpp
cyrusknopf: Fixed

storage/iterators/GetOutEdgesIterator.h
```diff
@@ -49,6 +56,13 @@ class GetOutEdgesIterator : public Iterator {
 
     void init();
     void nextValid();
+
+    /**
+     * @brief Advances @ref _partIt by at least @param n steps, or precisely
+     * d = distance(_partIt._it, _partIt._itEnd) steps if @ref n > d. Always leaves @ref
+     * _partIt pointing to a valid datapart, or pointing to the end iterator.
+     */
+    void advancePartIterator(size_t n);
```
rjb32, line 65: Please clarify why these changes are in these iterators specifically

cyrusknopf: These are the only iterators needed to check for the unique conflict check which covers cases such as

```
Change 1                 Change 2
new change            new change

delete nodes 1        ....

....                            create (n @ 1)-[e:NEWEDGE]-(m:NEWNODE)

....                            change submit

change submit      .....
```

Where Change 1 should be rejected because it would delete NEWEDGE which was not existing when it branched: write conflict.

I use the methods added in this PR to call GetOut/InEdges on the nodes which have been deleted, to see if any new edges have been made from them. So I advance the GetOut/InEdges iterator to the first datapart that did not exist when Change1 branched, and see if there are any more edges created from the deleted nodes. If there are, then write conflict and reject.  But this check is done in PR #67

LucLabarriere: Seems good to me. The weirdness of having only GetInEdges and GetOutEdges modified comes from the fact that iterators are not really used anymore. Even the interface is weird and comes from ancient time (they take Column* as inputs)

PR #79, opened 2025-10-23
DataPart Merge

ghost:
This PR contains the code for merging a span of data parts into 1 single datapart. 

It does not include the integration of the merger into the change system because there are many different concerns for that case - which will be addressed separately in the next PR.

### DIFF
--- a/storage/mergers/DataPartMerger.cpp
+++ b/storage/mergers/DataPartMerger.cpp
@@ -42,7 +42,7 @@ DataPartMerger::DataPartMerger(CommitData* commitData,
 {
 }
 
-std::unique_ptr<DataPartBuilder> DataPartMerger::merge(DataPartSpan dataParts, JobSystem& jobSystem) const {
+std::unique_ptr<DataPartBuilder> DataPartMerger::merge(DataPartSpan dataParts) const {
     const auto& graphReader = _graphView.read();
     const size_t nodeCount = graphReader.getNodeCount();
     const size_t edgeCount = graphReader.getEdgeCount();
--- a/test/storage/versioning/DatapartMergerTest.cpp
+++ b/test/storage/versioning/DatapartMergerTest.cpp
@@ -370,7 +370,7 @@ TEST_F(DataPartMergerTest, labels) {
     auto* commitBuilder = writer.openWriteTransaction().commitBuilder();
 
     DataPartMerger merger(&commitBuilder->commitData(),commitBuilder->metadata());
-    std::unique_ptr<DataPartBuilder> dataPartBuilder = merger.merge(commitBuilder->commitData().allDataparts(), *_jobSystem);
+    std::unique_ptr<DataPartBuilder> dataPartBuilder = merger.merge(commitBuilder->commitData().allDataparts());
 
     DataPart mergedDataPart = DataPart(0, 0);
 
--- a/storage/mergers/DataPartMerger.h
+++ b/storage/mergers/DataPartMerger.h
@@ -9,15 +9,13 @@ namespace db {
 class MetadataBuilder;
 class GraphView;
 class DataPartBuilder;
-class JobSystem;
 
 class DataPartMerger {
 public:
     DataPartMerger(CommitData* commitData,
                             MetadataBuilder& metadataBuilder);
 
-    std::unique_ptr<DataPartBuilder> merge(DataPartSpan dataParts,
-                                           JobSystem& jobsystem) const;
+    std::unique_ptr<DataPartBuilder> merge(DataPartSpan dataParts) const;
 
 private:
     GraphView _graphView;


storage/mergers/DataPartMerger.cpp
```diff
@@ -0,0 +1,200 @@
...
+    size_t edgeOffset = 0;
+    std::vector<EdgeRecord>& edges = datapartBuilder->edges();
+    edges.resize(edgeCount);
+
+    PropertyManager& nodePropertyManager = *datapartBuilder->nodeProperties();
+    PropertyManager& edgePropertyManager = *datapartBuilder->edgeProperties();
+
+    const auto& graphMetadata = _commitBuilder->commitData().metadata();
+
+    std::unordered_map<PropertyTypeID, size_t> _nodePropertiesCount;
```
rjb32, line 65: Add comments to delimit the phases of this function and explain the steps

storage/mergers/DataPartMerger.cpp
```diff
@@ -0,0 +1,200 @@
...
+            _nodePropertiesCount[it.first] += it.second->size();
+        }
+
+        for (const auto& it : dataPart->edgeProperties()) {
+            _edgePropertiesCount[it.first] += it.second->size();
+        }
+    }
+
+    for (const auto& dataPart : dataParts) {
+        for (const auto& record : dataPart->nodes().records()) {
```
rjb32: Don't use auto here but say explicitly NodeRecord

storage/mergers/DataPartMerger.cpp
```diff
@@ -0,0 +1,200 @@
...
+            labelSets.emplace_back(*graphMetadata.labelsets().getValue(record._labelset.getID()));
+        }
+
+        std::memcpy(edges.data() + edgeOffset, dataPart->edges()._outEdges.data(), (dataPart->edges().size() * sizeof(EdgeRecord)));
+        edgeOffset += dataPart->edges().size();
+
+        for (const auto& it : dataPart->nodeProperties()) {
```
rjb32: It's a bit weird to see a "for auto it" when "it" is probably not an iterator

storage/mergers/DataPartMerger.cpp
```diff
@@ -0,0 +1,200 @@
...
+                                oldContainer.ids().data(),
+                                (oldContainer.ids().size() * sizeof(EntityID)));
+
+                    _nodePropertiesCount[it.first] += oldContainer.ids().size();
+                }
+            };
+            PropertyTypeDispatcher {it.second->getValueType()}.execute(process);
+        }
+
+        for (const auto& it : dataPart->edgeProperties()) {
```
rjb32: Same here what is "it"?

storage/writers/DataPartBuilder.cpp
```diff
@@ -12,15 +10,14 @@ using namespace db;
...
 
 std::unique_ptr<DataPartBuilder> DataPartBuilder::prepare(MetadataBuilder& metadata,
-                                                          const GraphView& view,
+                                                          const size_t nodeCount,
+                                                          const size_t edgeCount,
                                                           size_t partIndex) {
-    const auto reader = view.read();
     auto* ptr = new DataPartBuilder();
 
-    ptr->_view = view;
```
rjb32, line 20: The view attribute is still in the class and not set anymore?

storage/versioning/CommitBuilder.cpp
```diff
@@ -44,7 +44,10 @@ DataPartBuilder& CommitBuilder::newBuilder() {
     std::scoped_lock lock {_mutex};
     GraphView view {*_commitData};
     const size_t partIndex = view.dataparts().size() + _builders.size();
-    auto& builder = _builders.emplace_back(DataPartBuilder::prepare(*_metadataBuilder, view, partIndex));
+    auto& builder = _builders.emplace_back(DataPartBuilder::prepare(*_metadataBuilder,
+                                                                    view.read().getNodeCount(),
```
rjb32, line 48: So actually you pass now the node count and the edge count that we were getting from the view in order to get an empty DataPartBuilder and pass zero in DataPartMerger.
I am not sure this is the best approach. Could a prepareEmpty function be better? 
There is also to what extent an existing GraphView is really necessary in DataPartBuilder at a high level?

rjb32:
I think DataPartMerge is rather readable given the task

storage/mergers/DataPartMerger.cpp
ghost: It is a PropertyContainerMap::Iterator 😅, maybe I can rename it to propertyIt.

LucLabarriere: I don't think so actually, isn't it a pair <PropertyTypeID, PropertyContainer> ?, if so you can extract the pair: `const auto& [propTypeID, propContainer] : ...`

test/storage/versioning/DatapartMergerTest.cpp
```diff
@@ -0,0 +1,503 @@
...
+        int64_t _propertyID {};
+        EntityType _oldID;
+        EntityType _newID;
+
+        EntityMapping() = delete;
+
+        explicit EntityMapping(int64_t propertyID)
+        :_propertyID(propertyID)
+        {
+        }
```
LucLabarriere, line 37: Formatting (indention and space after the :)

storage/mergers/DataPartMerger.cpp
ghost: Ah yes this would be cleaner!

storage/versioning/CommitBuilder.cpp
ghost: I don't think datapart builder really needs a view of the current graph? It's solely based on adding nodes edges and into the datapart builder - which are stand alone. And it is probably the responsibility of the datapart object to look into the existing graphView to verify info.

ghost:
> Seems good to me! Only remarks in terms of dependency/organization, do you think you could pass to the DataPartMerger a MetadataBuilder and GraphView instead of the commitBuilder directly (I think it's the only things you use). It would decouple a bit the DatapartMerger from the versioning which will change a lot in the future. If you can do that, then DataPartMerger can live in storage/ directly, next to DataPartBuilder, and the unit tests also don't have to be in the versioning/ subfolder.
> 
> I think this organization would be better because it puts emphases a bit more on the fact that it's a raw/core tool and that it's not thread safe (Instead if you pass CommitBuilder, it might give the impression that you use the CommitBuilder's mutex, which you don't)

@LucLabarriere that makes sense! I was not too sure of using commitBuilder as I did especially as I did see it as something that should exist on it's own.

storage/mergers/DataPartMerger.cpp
```diff
@@ -0,0 +1,214 @@
...
+};
+
+DataPartMerger::DataPartMerger(CommitData* commitData,
+                               MetadataBuilder& metadataBuilder)
+    : _graphView(*commitData),
+    _metadataBuilder(metadataBuilder)
+{
+}
+
+std::unique_ptr<DataPartBuilder> DataPartMerger::merge(DataPartSpan dataParts, JobSystem& jobSystem) const {
```
LucLabarriere: Do you need the job system by the way?

ghost: Good Catch that was a holdover from when I was building the datapart within the merge itself.

PR #80, opened 2025-10-23
Add serialisation for Tombstones

cyrusknopf:
Requires #67 

Adds serialisation for Tombstones

### DIFF
--- a/test/serialisation/TombstoneSerialisationTest.cpp
+++ b/test/serialisation/TombstoneSerialisationTest.cpp
@@ -1,12 +1,18 @@
 #include <gtest/gtest.h>
 
-#include "Panic.h"
-#include "SystemManager.h"
+#include <memory>
+
 #include "TuringTest.h"
 #include "TuringTestEnv.h"
+
+#include "SystemManager.h"
+#include "versioning/Tombstones.h"
+#include "Graph.h"
+#include "dump/GraphLoader.h"
 #include "columns/Block.h"
 #include "versioning/Change.h"
-#include "Graph.h"
+#include "versioning/Transaction.h"
+#include "Panic.h"
 
 using namespace db;
 using namespace turing::test;
@@ -18,18 +24,13 @@ public:
 
         _builtGraph = _env->getSystemManager().createGraph(_workingGraphName);
 
-        populate();
+        populateAndDump();
+        applyDeletesAndDump();
     }
 
-    
-
-    void populate() {
+    void populateAndDump() {
         auto& db = _env->getDB();
         auto& sysMan = _env->getSystemManager();
-
-        auto _callback = [this](const Block& block) {
-            this->_outBlock.assignFrom(block);
-        };
 
         auto res = sysMan.newChange(_workingGraphName);
         if (!res) {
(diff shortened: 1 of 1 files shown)


rjb32:
Looks good to me in principle. @LucLabarriere feel free to comment further on the dumping

storage/dump/TombstonesLoader.cpp
```diff
@@ -0,0 +1,58 @@
...
+        // Load into vector first
+        std::vector<NodeID> tempNodeVec;
+        tempNodeVec.reserve(nodeCount);
+        if (auto res = LoadUtils::loadVector(tempNodeVec, nodeCount, _reader, it); !res) {
+            return res;
+        }
+
+        // Insert all elements of vector into the tombstones set in one go
+        tombstones.edgeTombstones().reserve(nodeCount);
+        tombstones.nodeTombstones().insert(tempNodeVec);
```
LucLabarriere, line 38: - Why do you need to load to a vector first?

- there is typo line 37, you're reserving space for the edgeTombostones container instead of nodeTombstones

- @rjb32 and @cyrusknopf, what do you guys think of dumping the tombstones in a file like that (it's an unordered_set), should we be concerned with the fact that it's unordered ? I would assume that the order completely changes once we load back the set from disk. Worst case scenario we can always sort the ids before dumping, but I'm not sure that that's worth it

storage/dump/TombstonesDumper.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+    { // Dump edge tombstones
+        const Tombstones::EdgeTombstones& edgeTombstones = tombstones.edgeTombstones();
+        const size_t edgesSize = edgeTombstones.size();
+
+        DumpUtils::ensureDumpSpace(sizeof(edgesSize), _writer);
+        _writer.writeToCurrentPage(edgesSize);
+
+        DumpUtils::dumpRange(edgeTombstones, _writer);
+    }
```
LucLabarriere, line 31: I don't think this works, have you tested with many deletes? I think it needs to span multiple pages

cyrusknopf: Do you mean if there are so many tombstones to delete that it cannot fit on one page? If so, `dumpRange` is an abstraction that handles that

LucLabarriere: Oh ok sorry then, great

storage/dump/TombstonesLoader.cpp
cyrusknopf: - I was debating whether to load into vector first vs create an explicit `LoadUtils::loadUnorderedSet` method, but I assumed that loading into a vector would be faster than loading into an unordered set, but if we would rather have an explicit `loadUnorderedSet` method I can add it
- Oops good catch thanks
- Do we not already do this with  other unordered containers like `unordered_maps` in `LabelMap`? How would sorting help with determinism in this case?

LucLabarriere: All true yeah, I think we agreed with Remy that using unordered containers was a bit mandatory for better performances so it's only the comparators that need to not depend on orders (which you properly handled).

I think it's fine to keep the temporary vector for now, we don't know what the performances impacts are, and it's definitely not the bottleneck here

So all good. Nice PR

PR #81, opened 2025-10-24
Add filtering in ChunkWriters/Iterators using Tombstones

cyrusknopf:
- Implements a `TombstoneFilter` class for filtering the output of `ChunkWriters` based on the tombstones of the provided `GraphView`
- Adds additional checks in `QueryPlanner` to attempt to ensure that the columns required for correct filtering are always allocated

### DIFF
--- a/test/serialisation/TombstoneSerialisationTest.cpp
+++ b/test/serialisation/TombstoneSerialisationTest.cpp
@@ -1,12 +1,18 @@
 #include <gtest/gtest.h>
 
-#include "Panic.h"
-#include "SystemManager.h"
+#include <memory>
+
 #include "TuringTest.h"
 #include "TuringTestEnv.h"
+
+#include "SystemManager.h"
+#include "versioning/Tombstones.h"
+#include "Graph.h"
+#include "dump/GraphLoader.h"
 #include "columns/Block.h"
 #include "versioning/Change.h"
-#include "Graph.h"
+#include "versioning/Transaction.h"
+#include "Panic.h"
 
 using namespace db;
 using namespace turing::test;
@@ -18,18 +24,13 @@ public:
 
         _builtGraph = _env->getSystemManager().createGraph(_workingGraphName);
 
-        populate();
+        populateAndDump();
+        applyDeletesAndDump();
     }
 
-    
-
-    void populate() {
+    void populateAndDump() {
         auto& db = _env->getDB();
         auto& sysMan = _env->getSystemManager();
-
-        auto _callback = [this](const Block& block) {
-            this->_outBlock.assignFrom(block);
-        };
 
         auto res = sysMan.newChange(_workingGraphName);
         if (!res) {
@@ -39,27 +40,55 @@ public:
 
         // populate the graph
         for (size_t i = 0; i < NUM_EDGES; i++) {
-            db.query("create (n:Person)-[e:FRIENDSWITH]-(m:Person)",
-                     _workingGraphName,
-                     &_env->getMem(),
-                     CommitHash::head(),
-                     change->id());
+            ASSERT_TRUE(db.query("create (n:Person)-[e:FRIENDSWITH]-(m:Person)",
+                                 _workingGraphName, &_env->getMem(), CommitHash::head(),
+                                 change->id()));
         }
         spdlog::info("Ran create queries");
 
-        db.query("change submit",
-                 _workingGraphName,
-                 &_env->getMem(),
-                 CommitHash::head(),
-                 change->id());
+        // implicit dump on change submit
+        ASSERT_TRUE(db.query("change submit", _workingGraphName, &_env->getMem(),
+                             CommitHash::head(), change->id()));
         spdlog::info("Submitted change");
 
-        db.query("match (n) return n", _workingGraphName, &_env->getMem(), _callback,
-                 CommitHash::head(), ChangeID::head());
-        if (_outBlock.empty()) {
-            panic("Failed to populate graph.");
-        }
+        const auto VERIFY = [](const Block& block) {
+            if (block.empty()) {
+                panic("Failed to populate graph.");
+            }
+            if (block.columns().front()->size() != NUM_NODES) {
+                panic("Failed to populate graph.");
+            }
+        };
+
+        ASSERT_TRUE(db.query("match (n) return n", _workingGraphName, &_env->getMem(),
+                             VERIFY, CommitHash::head(), ChangeID::head()));
+
         spdlog::info("Successfully populated graph");
+    }
+
+    void applyDeletesAndDump() {
+        auto& db = _env->getDB();
+        auto& sysMan = _env->getSystemManager();
+
+        auto delRes = sysMan.newChange(_workingGraphName);
+        if (!delRes) {
+            panic("Failed to make change in populate().");
+        }
+        Change* delChange = delRes.value();
+
+        for (size_t node : DELETED_NODES) {
+            db.query("delete nodes " + std::to_string(node), _workingGraphName,
+                     &_env->getMem(), CommitHash::head(), delChange->id());
+        }
+        for (size_t node : DELETED_EDGES) {
+            db.query("delete edges " + std::to_string(node), _workingGraphName,
+                     &_env->getMem(), CommitHash::head(), delChange->id());
+        }
+        // implicit dump on change submit
+        ASSERT_TRUE(db.query("change submit", _workingGraphName, &_env->getMem(),
+                             CommitHash::head(), delChange->id()));
+
+        spdlog::info("Submitted deletions change");
     }
 
 protected:
@@ -73,8 +102,39 @@ protected:
 
     static constexpr size_t NUM_EDGES = 10;
     static constexpr size_t NUM_NODES = 2 * NUM_EDGES;
+
+    static constexpr std::array<size_t, 3> DELETED_NODES = {0, NUM_NODES - 1, 2};
+    static constexpr std::array<size_t, 4> DELETED_EDGES = {0, NUM_EDGES - 1, 1, 4};
 };
 
-TEST_F(TombstoneSerialisationTest, deleteNodesAfterLoad) {
-    ASSERT_TRUE(false);
+TEST_F(TombstoneSerialisationTest, deleteNodesThenLoad) {
+    _loadedGraph = Graph::create();
+    auto res = GraphLoader::load(
+        _loadedGraph.get(),
+        _env->getSystemManager().getGraph(_workingGraphName)->getPath());
+    ASSERT_TRUE(res);
+
+
+    const Tombstones& tombstones =
+        _loadedGraph->openTransaction().viewGraph().commits().back().tombstones();
+
+    const Tombstones::NodeTombstones& nodeTombstones = tombstones.nodeTombstones();
+    const Tombstones::EdgeTombstones& edgeTombstones = tombstones.edgeTombstones();
+
+    ASSERT_EQ(DELETED_NODES.size(), nodeTombstones.size());
+    ASSERT_EQ(DELETED_EDGES.size(), edgeTombstones.size());
+
+    for (const NodeID deletedNode : DELETED_NODES) {
+        ASSERT_TRUE(nodeTombstones.contains(deletedNode));
+    }
+
+    for (const EdgeID deletedEdge : DELETED_EDGES) {
+        ASSERT_TRUE(edgeTombstones.contains(deletedEdge));
+    }
+}
+
+int main(int argc, char** argv) {
+    return turingTestMain(argc, argv, [] {
+        testing::GTEST_FLAG(repeat) = 1;
+    });
 }


storage/iterators/TombstoneFilter.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace db {
+template void TombstoneFilter::populateDeletedIndices<NodeID>(const ColumnVector<NodeID>& column);
+template void TombstoneFilter::populateDeletedIndices<EdgeID>(const ColumnVector<EdgeID>& column);
+
+template void TombstoneFilter::onePassApplyFilter<NodeID>(ColumnVector<NodeID>& column);
+template void TombstoneFilter::onePassApplyFilter<EdgeID>(ColumnVector<EdgeID>& column);
+}
+
+TombstoneFilter::TombstoneFilter(const Tombstones& tombstones)
+        : _tombstones(tombstones)
```
LucLabarriere: indention

storage/iterators/TombstoneFilter.h
```diff
@@ -0,0 +1,160 @@
...
+        this->applyDeletedIndices(*col);
+    };
+
+    (populateIfID(columns), ...); // Populate @ref _deletedIndices for ID columns
+    if (_deletedIndices.empty()) { // Nothing to filter -> exit early
+        return;
+    }
+
+    (applyFilter(columns), ...);  // Apply filter to all columns
+}
```
LucLabarriere: I'm honestly not a fan of the templated approach, and I'm not even sure I understand what's going on. I don't see the onePassApplyFilter function being used anywhere. Can you try without templates? Something like this (can probably be improved, but at least it doesn't rely on templates):

```c++
inline void TombstoneFilter::filter(std::initializer_list<Column*> columns) {
    bool hasNodes = _tombstones.hasNodes();
    bool hasEdges = _tombstones.hasEdges();

    // Step 1. Populate @ref _deletedIndices with the indices of rows to be deleted
    for (Column* col : columns) {
        if (!col) {
            continue;
        }

        if (auto* c = dynamic_cast<ColumnVector<NodeID>*>(col)) {
            if (!hasNodes) {
                continue;
            }

            this->populateDeletedIndices(*c);
        } else if (auto* c = dynamic_cast<ColumnVector<EdgeID>*>(col)) {
            if (!hasEdges) {
                continue;
            }

            this->populateDeletedIndices(*c);
        } else [[unlikely]] {
            throw std::runtime_error("TombstoneFilter::filter: column is not a NodeID or EdgeID column");
        }
    }

    if (_deletedIndices.empty()) {
        // Nothing to filter -> exit early
        return;
    }

    // Step 2. Apply the deletions to all non-null columns
    for (Column* col : columns) {
        if (!col) {
            continue;
        }

        if (auto* c = dynamic_cast<ColumnVector<NodeID>*>(col)) {
            if (!hasNodes) {
                continue;
            }

            this->applyDeletedIndices(*c);
        } else if (auto* c = dynamic_cast<ColumnVector<EdgeID>*>(col)) {
            if (!hasEdges) {
                continue;
            }

            this->applyDeletedIndices(*c);
        } else [[unlikely]] {
            throw std::runtime_error("TombstoneFilter::filter: column is not a NodeID or EdgeID column");
        }
    }
}
```

I mean there are only two cases, so you might as well write both versions for all methods:
```c++
void TombstoneFilter::populateDeletedIndices(const ColumnVector<NodeID>& column) {
    for (size_t i = 0; i < column.size(); i++) {
        const NodeID id = column.getRaw()[i];

        if (_tombstones.containsNode(id)) {
            _deletedIndices.insert(i);
        }
    }
}

void TombstoneFilter::populateDeletedIndices(const ColumnVector<EdgeID>& column) {
    for (size_t i = 0; i < column.size(); i++) {
        const EdgeID id = column.getRaw()[i];

        if (_tombstones.containsEdge(id)) {
            _deletedIndices.insert(i);
        }
    }
}
```

LucLabarriere: Even better, you could do the dynamic_cast thing + hasNodes and hasEdges check directly in the populateDeletedIndices and applyDeletedIndices methods directly so that the filter is simply:
```c++
inline void TombstoneFilter::filter(std::initializer_list<Column*> columns) {
    // Step 1. Populate @ref _deletedIndices with the indices of rows to be deleted
    for (Column* col : columns) {
        if (col) {
            this->populateDeletedIndices(c);
        }
    }

    if (_deletedIndices.empty()) {
        // Nothing to filter -> exit early
        return;
    }

    // Step 2. Apply the deletions to all non-null columns
    for (Column* col : columns) {
        if (col) {
            this->applyDeletedIndices(*c);
        }
    }
}
```

cyrusknopf: I can try a non-templated approach, sure.

Just to clear a misunderstanding up, though, I wanted the filter method to take in an arbitrary number of columns since each ChunkWriter has a different number of columns to produce, whilst also those columns being of any internal type. For example some chunk writers produce `ColumnVec<NodeID>, ColumnVec<SupportedTypeT>`, etc. whilst others produce `ColumnVec<EdgeID>, ColumnVec<NodeID>,...` etc etc.

The two steps of the filter:
1. populating deleted indices
2. applying deletions

require operating on different types of columns. Step 1 is only concerned with finding indices in a column which contain an edge/node ID which is deleted. Hence step 1 has nothing to do with columns such as `ColumnVec<types::String>`, because it doesn't make sense to filter based on a property, only on IDs which are deleted. This is why the `populateDeletedIndices` function does nothing to non-ID type columns.

Step 2. removes the indices which were populated as deleted based on step 1 to _all_ columns, regardless of their type, and so that is why there is no conditional behaviour based on the internal type of the column.

All that being said, I can definitely achieve a similar behaviour with non-templates, and I suppose the fact that what the current templated implementation does being not clear is a sign it is overly complex

LucLabarriere: Yeah ok I got the algorithm right but I completely forgot about more complex columns like `ColumnVec<types::String>`, it's true that that might need a templated approach.

If you go with what I mentioned in the meeting where you use a base column as the reference to check for tombstones and then apply the filter to each column, then maybe the template function that applies the filter can be implemented in the header file directly (and kept really simple, no concepts, nothing, just take a `ColumnVector<T>` and filter it based on `ColumnVector<NodeID/EdgeID>`)

cyrusknopf:
```cpp
void TombstoneFilter::filterGetOutEdges(ColumnEdgeIDs* edgeIDs, ColumnNodeIDs* tgtIDs,
                                        ColumnEdgeTypes* edgeTypes, ColumnIndices* indices) {
    // Filter based on deleted edges: if a target has been deleted, any edges incident to
    // it would also be marked deleted
    ColumnEdgeIDs* const baseColumn = edgeIDs;
    bioassert(baseColumn); // XXX Should be exception?

    size_t initialSize = baseColumn->size();

    size_t writePtr = 0;
    for (size_t readPtr = 0; readPtr < initialSize; readPtr++) {
        EdgeID edge = (*baseColumn)[readPtr];
        if (_tombstones.containsEdge(edge)) {
            continue;
        }
        (*baseColumn)[writePtr] = (*baseColumn)[readPtr];
        if (tgtIDs) {
            (*tgtIDs)[writePtr] = (*tgtIDs)[readPtr];
        }
        if (edgeTypes) {
            (*edgeTypes)[writePtr] = (*edgeTypes)[readPtr];
        }
        if (indices) {
            (*indices)[writePtr] = (*indices)[readPtr];
        }
        writePtr++;
    }
    baseColumn->resize(writePtr);
    if (tgtIDs) {
        tgtIDs->resize(writePtr);
    }
    if (edgeTypes) {
        edgeTypes->resize(writePtr);
    }
    if (indices) {
        indices->resize(writePtr);
    }
}
```

@LucLabarriere @rjb32 @suhassn-turing-bio  Here is my first draft of the "base column" approach for filtering GetOutEdges. The remaining iterators will look similar. Some things to note:
1. There is not auxiliary data store of indices: not a set nor a vector
2. The functions for each chunk writer are not templated (as we preferred), but they need to be explicitly defined for each chunk writer
3. In the inner loop there are a lot more conditionals, as we need to explicitly check what columns are provided. In contrast, my current implementation has a much tighter loop, at the cost of an additional auxiliary data structure for lookups. It is hard to quantify how this affects cache compared to updating each column at a time


How do we feel now considering the different options?
- Do we think this new approach is preferred, given that an explicit specialisation need be defined for each chunk writer?
- Do we think that this larger, less tight loop is preferred? I think it looks worse than it is, because all the internal branches for nullptr are extremely predictable and therefore should be negligible once the branch predictor is saturated
- Do we think that the cost of an external vector (not u_set as my initial implementation) is worth the potential speedup that we can gain from performing a memcpy instead of the explicit for loop (the loop in my original and the above implementations)

Interested in hearing your guys' thoughts

rjb32:
> ```c++
> void TombstoneFilter::filterGetOutEdges(ColumnEdgeIDs* edgeIDs, ColumnNodeIDs* tgtIDs,
>                                         ColumnEdgeTypes* edgeTypes, ColumnIndices* indices) {
>     // Filter based on deleted edges: if a target has been deleted, any edges incident to
>     // it would also be marked deleted
>     ColumnEdgeIDs* const baseColumn = edgeIDs;
>     bioassert(baseColumn); // XXX Should be exception?
> 
>     size_t initialSize = baseColumn->size();
> 
>     size_t writePtr = 0;
>     for (size_t readPtr = 0; readPtr < initialSize; readPtr++) {
>         EdgeID edge = (*baseColumn)[readPtr];
>         if (_tombstones.containsEdge(edge)) {
>             continue;
>         }
>         (*baseColumn)[writePtr] = (*baseColumn)[readPtr];
>         if (tgtIDs) {
>             (*tgtIDs)[writePtr] = (*tgtIDs)[readPtr];
>         }
>         if (edgeTypes) {
>             (*edgeTypes)[writePtr] = (*edgeTypes)[readPtr];
>         }
>         if (indices) {
>             (*indices)[writePtr] = (*indices)[readPtr];
>         }
>         writePtr++;
>     }
>     baseColumn->resize(writePtr);
>     if (tgtIDs) {
>         tgtIDs->resize(writePtr);
>     }
>     if (edgeTypes) {
>         edgeTypes->resize(writePtr);
>     }
>     if (indices) {
>         indices->resize(writePtr);
>     }
> }
> ```
> 
> @LucLabarriere @rjb32 @suhassn-turing-bio Here is my first draft of the "base column" approach for filtering GetOutEdges. The remaining iterators will look similar. Some things to note:
> 
> 1. There is not auxiliary data store of indices: not a set nor a vector
> 2. The functions for each chunk writer are not templated (as we preferred), but they need to be explicitly defined for each chunk writer
> 3. In the inner loop there are a lot more conditionals, as we need to explicitly check what columns are provided. In contrast, my current implementation has a much tighter loop, at the cost of an additional auxiliary data structure for lookups. It is hard to quantify how this affects cache compared to updating each column at a time
> 
> How do we feel now considering the different options?
> 
> * Do we think this new approach is preferred, given that an explicit specialisation need be defined for each chunk writer?
> * Do we think that this larger, less tight loop is preferred? I think it looks worse than it is, because all the internal branches for nullptr are extremely predictable and therefore should be negligible once the branch predictor is saturated
> * Do we think that the cost of an external vector (not u_set as my initial implementation) is worth the potential speedup that we can gain from performing a memcpy instead of the explicit for loop (the loop in my original and the above implementations)
> 
> Interested in hearing your guys' thoughts

There is indeed now if statements inside the loop. The templates were a way to generate implicitly each possible specialization of this loop

cyrusknopf:
> ```c++
> void TombstoneFilter::filterGetOutEdges(ColumnEdgeIDs* edgeIDs, ColumnNodeIDs* tgtIDs,
>                                         ColumnEdgeTypes* edgeTypes, ColumnIndices* indices) {
>     // Filter based on deleted edges: if a target has been deleted, any edges incident to
>     // it would also be marked deleted
>     ColumnEdgeIDs* const baseColumn = edgeIDs;
>     bioassert(baseColumn); // XXX Should be exception?
> 
>     size_t initialSize = baseColumn->size();
> 
>     size_t writePtr = 0;
>     for (size_t readPtr = 0; readPtr < initialSize; readPtr++) {
>         EdgeID edge = (*baseColumn)[readPtr];
>         if (_tombstones.containsEdge(edge)) {
>             continue;
>         }
>         (*baseColumn)[writePtr] = (*baseColumn)[readPtr];
>         if (tgtIDs) {
>             (*tgtIDs)[writePtr] = (*tgtIDs)[readPtr];
>         }
>         if (edgeTypes) {
>             (*edgeTypes)[writePtr] = (*edgeTypes)[readPtr];
>         }
>         if (indices) {
>             (*indices)[writePtr] = (*indices)[readPtr];
>         }
>         writePtr++;
>     }
>     baseColumn->resize(writePtr);
>     if (tgtIDs) {
>         tgtIDs->resize(writePtr);
>     }
>     if (edgeTypes) {
>         edgeTypes->resize(writePtr);
>     }
>     if (indices) {
>         indices->resize(writePtr);
>     }
> }
> ```
> 
> @LucLabarriere @rjb32 @suhassn-turing-bio Here is my first draft of the "base column" approach for filtering GetOutEdges. The remaining iterators will look similar. Some things to note:
> 
> 1. There is not auxiliary data store of indices: not a set nor a vector
> 2. The functions for each chunk writer are not templated (as we preferred), but they need to be explicitly defined for each chunk writer
> 3. In the inner loop there are a lot more conditionals, as we need to explicitly check what columns are provided. In contrast, my current implementation has a much tighter loop, at the cost of an additional auxiliary data structure for lookups. It is hard to quantify how this affects cache compared to updating each column at a time
> 
> How do we feel now considering the different options?
> 
> * Do we think this new approach is preferred, given that an explicit specialisation need be defined for each chunk writer?
> * Do we think that this larger, less tight loop is preferred? I think it looks worse than it is, because all the internal branches for nullptr are extremely predictable and therefore should be negligible once the branch predictor is saturated
> * Do we think that the cost of an external vector (not u_set as my initial implementation) is worth the potential speedup that we can gain from performing a memcpy instead of the explicit for loop (the loop in my original and the above implementations)
> 
> Interested in hearing your guys' thoughts

For context, here is what a first implementation of Remy's memmove option could look like

```cpp
template <typename T>
void TombstoneFilter::applyDeletedVec(ColumnVector<T>& column) {
    bioassert(!_delVec.empty());

    if (column.empty()) {
        return;
    }

    // We should be able to omit these if we only generated filtered indices on a "base
    // column" (e.g. the edge column in GetOutEdges)
    std::ranges::sort(_delVec);
    auto [newEnd, oldEnd] = std::ranges::unique(_delVec);
    _delVec.erase(newEnd, oldEnd);

    // Unique and sorted
    bioassert(std::ranges::adjacent_find(_delVec) == _delVec.end());
    bioassert(std::ranges::is_sorted(_delVec));

    std::vector<T>& raw = column.getRaw();
    size_t initialSize = raw.size();

    // Track the index in the column we are reading from and writing to
    size_t readPtr {0};
    size_t writePtr {0};
    for (size_t i = 0; i < _delVec.size();) {
        // Compute the range [start, end] that is to be deleted
        size_t deletedRangeStart = _delVec[i];
        size_t deletedRangeEnd = deletedRangeStart;

        // Find a contiguous block of deleted indices
        while (i + 1 < _delVec.size() && _delVec[i + 1] == _delVec[i] + 1) {
            i++;
            deletedRangeEnd++;
        }

        // Calculate the size of the block to be deleted based on the indexes of the
        // column
        size_t deletedBlockSize = deletedRangeStart - readPtr;

        std::memmove(raw.data() + writePtr, // Write to the current point in the column
                     raw.data() + readPtr,  // From where we are reading the in the col
                     deletedBlockSize * sizeof(T));

        // Increment so the next place we write is after the block we just moved
        writePtr += deletedBlockSize;

        // Do not read the deleted block that we just skipped
        readPtr = deletedRangeEnd + 1;
        i++;
    }

    // There may be an interval [_delVec.back(), column.size()]. Move that down
    if (readPtr < initialSize) {
        size_t remainingSize = initialSize - readPtr;
        std::memmove(raw.data() + writePtr,
                     raw.data() + readPtr,
                     remainingSize * sizeof(T));
        writePtr += remainingSize;
    }

    // We potentially shifted elements down, so the end of the vector needs to be
    // truncated
    raw.resize(writePtr);
}
```

LucLabarriere:
I was thinking something much simpler (Didn't look too much the memmove approach, will look at it):
```c++
class TombstoneFilter {
public:
    explicit TombstoneFilter(const Tombstones& tombstones);
    ~TombstoneFilter() = default;

    TombstoneFilter(const TombstoneFilter&) = delete;
    TombstoneFilter(TombstoneFilter&&) = delete;
    TombstoneFilter& operator=(const TombstoneFilter&) = delete;
    TombstoneFilter& operator=(TombstoneFilter&&) = delete;

    template <typename T>
    void filter(const ColumnVector<NodeID>& baseCol, ColumnVector<T>* col) {
        if (!col) {
            return;
        }

        if (!_tombstones.hasNodes()) [[unlikely]] {
            throw std::runtime_error("Tombstones: Cannot filter using a nullptr base column");
        }

        if (baseCol.size() != col->size()) [[unlikely]] {
            throw std::runtime_error("Tombstones: NodeIDs and to be filtered column must have the same size");
        }

        ColumnVector<T>& filtered = *col;

        size_t writePtr = 0;

        for (size_t i = 0; i < baseCol.size(); i++) {
            if (_tombstones.containsNode(baseCol[i])) {
                continue;
            }

            filtered[writePtr++] = filtered[i];
        }

        filtered.resize(writePtr);
    }

    /* Same thing for a ColumnVector<EdgeID> base col, or use a template approach */

private:
    const Tombstones& _tombstones;
};

// For example in GetOutEdgesChunkWriter::fill(...)
    if (_view.tombstones().hasEdges()) {
        TombstoneFilter filter(_view.tombstones());

        // Must be set by the caller (e.g. Query planner)
        if (!_edgeIDs) [[unlikely]] {
            throw std::runtime_error("GetOutEdges: Commit has tombstones, but no edge IDs were provided");
        }

        filter.filter(*_edgeIDs, _indices);
        filter.filter(*_edgeIDs, _edgeIDs);
        filter.filter(*_edgeIDs, _tgts);
        filter.filter(*_edgeIDs, _types);
    }
```

Issues with this approach:
- Code duplication for Node/Edge ids. Templated approach could be good here:
```c++
class TombstoneFilter {
public:
    explicit TombstoneFilter(const Tombstones& tombstones);
    ~TombstoneFilter() = default;

    TombstoneFilter(const TombstoneFilter&) = delete;
    TombstoneFilter(TombstoneFilter&&) = delete;
    TombstoneFilter& operator=(const TombstoneFilter&) = delete;
    TombstoneFilter& operator=(TombstoneFilter&&) = delete;

    template <TypedInternalID ID>
    bool shouldFilter() const {
        return std::is_same_v<ID, NodeID>
                 ? _tombstones.hasNodes()
                 : _tombstones.hasEdges();
    }

    template <TypedInternalID ID>
    bool hasTombstone(ID id) const {
        return std::is_same_v<ID, NodeID>
                 ? _tombstones.containsNode(id)
                 : _tombstones.containsEdge(id);
    }

    template <TypedInternalID ID, typename T>
    void filter(const ColumnVector<ID>& baseCol, ColumnVector<T>* col) {
        if (!col) {
            return;
        }

        if (!shouldFilter<ID>()) [[unlikely]] {
            throw std::runtime_error("Tombstones: Cannot filter using a nullptr base column");
        }

        if (baseCol.size() != col->size()) [[unlikely]] {
            throw std::runtime_error("Tombstones: Base column and to be filtered column must have the same size");
        }

        ColumnVector<T>& filtered = *col;

        size_t writePtr = 0;

        for (size_t i = 0; i < baseCol.size(); i++) {
            if (hasTombstone(baseCol[i])) {
                continue;
            }

            filtered[writePtr++] = filtered[i];
        }

        filtered.resize(writePtr);
    }

private:
    const Tombstones& _tombstones;
};
```

- We perform the hash map look up for the tombstone for each column. To profile but maybe your approach of building a vector<size_t> first on the base column might be faster. Can be stored in a TombstoneFilter member, and built with some kind of init function: 

```c++
if (_view.tombstones().hasEdges()) {
        TombstoneFilter filter(_view.tombstones());

        // Must be set by the caller (e.g. Query planner)
        if (!_edgeIDs) [[unlikely]] {
            throw std::runtime_error("GetOutEdges: Commit has tombstones, but no edge IDs were provided");
        }

       filter.init(*_edgeIDs); // Builds the indices

        filter.filter(_indices);
        filter.filter(_edgeIDs);
        filter.filter(_tgts);
        filter.filter(_types);
    }
```

LucLabarriere:
the memmove approach seems interesting but I think we should profile because it creates a temporary to hold the block that is copied, if there is a single row deleted in the middle of a 10k lines column, we need to allocate a temporary 5k lines array to hold the data. (By the way, maybe a memcpy approach would work here?)

cyrusknopf:
> the memmove approach seems interesting but I think we should profile because it creates a temporary to hold the block that is copied, if there is a single row deleted in the middle of a 10k lines column, we need to allocate a temporary 5k lines array to hold the data. (By the way, maybe a memcpy approach would work here?)

Memcpy is not possible I don't think because the dst and src are guaranteed to overlap which leads to UB i believe. This is why I used memmove. In any case, memmove has an internal check to see if it can use memcpy (in the case that intervals don't overlap/ overlap in a way that can be partly memcpyd). Also I read that any sensible implementation of memmove does not actually create the extra buffer. So yeah, to profile.

LucLabarriere:
> > the memmove approach seems interesting but I think we should profile because it creates a temporary to hold the block that is copied, if there is a single row deleted in the middle of a 10k lines column, we need to allocate a temporary 5k lines array to hold the data. (By the way, maybe a memcpy approach would work here?)
> 
> Memcpy is not possible I don't think because the dst and src are guaranteed to overlap which leads to UB i believe. This is why I used memmove. In any case, memmove has an internal check to see if it can use memcpy (in the case that intervals don't overlap/ overlap in a way that can be partly memcpyd). Also I read that any sensible implementation of memmove does not actually create the extra buffer. So yeah, to profile.

ok interesting

storage/iterators/TombstoneFilter.cpp
cyrusknopf: Fixed

plan/QueryPlanner.cpp
```diff
@@ -969,16 +969,22 @@ void QueryPlanner::planScanEdges(const EntityPattern* source,
...
+            edges = _mem->alloc<ColumnEdgeIDs>();
             edgeWriteInfo._edges = edges;
             _transformData->addColumn(edges, edgeDecl);
         }
     }
+    // ... or if we have tombstones and need to filter chunk writer outputs
+    if (!edges && (_view.tombstones().hasEdges() || _view.tombstones().hasNodes())) {
+        edges = _mem->alloc<ColumnEdgeIDs>();
+        edgeWriteInfo._edges = edges;
+    }
```
LucLabarriere, line 1005: Are you sure that just checking _view.tombstones().hasEdges() wouldn't be enough? If there are deleted nodes but no deleted edges, it means we deleted isolated nodes which shouldn't impact edgeIDs nor their source/target

samples/v2/main.cpp
```diff
@@ -79,8 +82,9 @@ int main(int argc, char** argv) {
         dumper.dump(std::cout);
     }
 
+    PlanGraphGenerator planGen(ast, view);
+    const PlanGraph& planGraph = planGen.getPlanGraph();
```
LucLabarriere, line 86: I noticed that you sometimes had other people's commits in your PRs, I'm not sure if we also have this in ours. How do you pull the changes of main in your branch. Do you run `git update`? (It does `git pull origin main --rebase`)

storage/iterators/GetInEdgesIterator.cpp
```diff
@@ -106,6 +107,29 @@ GetInEdgesChunkWriter::GetInEdgesChunkWriter(const GraphView& view,
 {
 }
 
+void GetInEdgesChunkWriter::filterTombstones() {
+    // Base column of this ChunkWriter is _edgeIDs
+    bioassert(_edgeIDs);
```
LucLabarriere, line 112: I think I would prefer to throw an exception for this check, because it's hard to know what code path could yield to this point and we prefer to be able to catch an exception (even if it's a bug), than run this and segfault the whole server.

storage/iterators/GetInEdgesIterator.cpp
```diff
@@ -114,7 +138,7 @@ void GetInEdgesChunkWriter::fill(size_t maxCount) {
     static constexpr auto bools = generateArray<NColumns, NCombinations>();
     static constexpr auto masks = generateBitmasks<NColumns, NCombinations>();
 
-    _indices->clear();
+     _indices->clear();
```
LucLabarriere: There's a new space at the beginning of this line

storage/iterators/GetInEdgesIterator.cpp
```diff
@@ -186,6 +210,10 @@ void GetInEdgesChunkWriter::fill(size_t maxCount) {
         CASE(6);
         CASE(7);
     }
+
+    if (_view.tombstones().hasEdges() || _view.tombstones().hasNodes()) {
+        filterTombstones();
+    }
```
LucLabarriere: Shouldn't just checking for _view.tombstones().hasEdges() be enough?

storage/iterators/ScanEdgesIterator.cpp
```diff
@@ -62,6 +63,30 @@ ScanEdgesChunkWriter::ScanEdgesChunkWriter(const GraphView& view)
...
+    }
+    if (_tgts) {
+        filter.filter(_tgts);
+        bioassert(_tgts->size() == newSize);
+    }
+    if (_types) {
+        filter.filter(_types);
+        bioassert(_types->size() == newSize);
+    }
+}
```
LucLabarriere, line 86: Would it be better to put the if statements in the FilterTombstones::filter method directly ? Early return if the column is not set (you can't have the bioassert if you do that though)

storage/iterators/ScanNodePropertiesIterator.cpp
```diff
@@ -86,6 +88,23 @@ ScanNodePropertiesChunkWriter<T>::ScanNodePropertiesChunkWriter(const GraphView&
...
+
+    TombstoneFilter filter(this->_view.tombstones());
+    filter.populateRanges(_nodeIDs);
+
+    filter.filter(_nodeIDs);
+
+    if (_properties) {
+        filter.filter(_properties);
+        size_t newSize = _nodeIDs->size();
+        bioassert(_properties->size() == newSize);
```
LucLabarriere: This is typically why I don't like asserts, in this code snippet, `newSize` can become "unused" depending on the implementation of the assert. I'm not sure how it compiles in Release mode here

storage/iterators/GetInEdgesIterator.cpp
LucLabarriere: I would think we want to assert things that should never happen in restricted parts of the code base, like within a single algorithm. Example: some algorithm is ran in two stages, first part is supposed to find 1 valid entry to some object. Second stage uses it. It makes sense here to assert that the pointer is valid before going to the next stage (Although I'm not even sure tbh because it's risky to crash the whole server is something wrong happens).

But here, the _edgeIDs ptr is coming from a really complex code path (QueryPlanner, later planner v2, and what if we implement python bindings to build pipelines, or the SQL query language, etc.)

plan/QueryPlanner.cpp
cyrusknopf: Yep, you're right

samples/v2/main.cpp
cyrusknopf: Yeah I run `git update`. I also don't know why this is showing in the diff, because I think main has these changes anyway?

storage/iterators/GetInEdgesIterator.cpp
cyrusknopf: Thinking from a user perspective, if we throw an exception on this case it means that we have made a mistake: we have not allocated the _edgeID column in the Planner when we should have. So this would then leave the DB in a strange state for the user as whatever query triggers the code path can never be executed, as it will always trigger an exception. I agree that is better than a segfault, but it still feels a bit weird.

In general I see exceptions as catching erroneous behaviour as the result of a user's mistake, which is not the case here, but if we prefer an exception here I can add it. If I were to, should I add a new exception type? or use the PipelineException (I don't think we throw exceptions in any of the chunk writers/iterators so far), and I will have to check what will catch that exception.

storage/iterators/GetInEdgesIterator.cpp
cyrusknopf: Thanks

storage/iterators/GetInEdgesIterator.cpp
cyrusknopf: Yeah good catch thanks, that was a left over I forgot to fix

storage/iterators/ScanEdgesIterator.cpp
cyrusknopf: If the check is in the iterator, we can predict the branch not-taken and avoid the function invocation which would always result in a jump, even if it returns immediately. I think.

storage/iterators/ScanNodePropertiesIterator.cpp
cyrusknopf: Actually yeah I was assuming that newSize would just be optimised out in release, but it might not even compile. I will just replace newSize with an invocation of _nodeIDs->size() everywhere

storage/iterators/GetInEdgesIterator.cpp
LucLabarriere: Yeah I agree 100%, the issue is that it's better from the client's perspective to get a weird EXEC_ERROR, than to have something like "Incomplete chunks received" from their http client and then see that the server is now unreachable (it crashed), and have to restart the server from the console. I'll have to do a bit of research to check what others do in other complex server implementations like ours.

@rjb32 What do you think? I think it would be interesting to have a custom FatalException defined somewhere (maybe common), that would inherit from std::logic_error (exception for things that should never happen like out of bounds errors). 

You would also simply throw std::logic_error directly since we already catch errors in the interpreter:

```c++
    // Execute
    ExecutionContext execCtxt(_sysMan, _jobSystem, view, graphName, commitHash, changeID, &txRes.value());
    try {
        Executor executor;
        executor.run(&execCtxt, planner.getPipeline());
    } catch (const PipelineException& e) {
        return QueryStatus(QueryStatus::Status::EXEC_ERROR, e.what());
    } catch (const std::exception& e) {
        return QueryStatus(QueryStatus::Status::EXEC_ERROR,
                           std::string("Unexpected exception: ") + e.what());
    } catch (...) {
        return QueryStatus(QueryStatus::Status::EXEC_ERROR,
                           "Unknown exception occurred");
    }
```

samples/v2/main.cpp
LucLabarriere: Yeah maybe, and maybe it happens for all PRs and I'm just noticing it now ahah. Let's forget about it

storage/iterators/GetInEdgesIterator.cpp
cyrusknopf: Ah yeah I really like the option of a special FatalException type

storage/versioning/Tombstones.h
```diff
@@ -1,7 +1,9 @@
 #pragma once
 
 #include <ranges>
+#include <type_traits>
```
LucLabarriere: You don't need type_traits anymore or do you?

storage/iterators/GetInEdgesIterator.cpp
rjb32: Yes agree, a FatalException could be good. I do use PipelineException myself in the pipeline processors when I am doing a dynamic cast and I check if it's the right column type. If it happens this is a bug  of the pipeline generator and not the fault of the user. But it's nice indeed to get an exception acting a bit like an assert doing a "soft" abort

storage/iterators/GetOutEdgesIterator.cpp
```diff
@@ -106,6 +108,28 @@ GetOutEdgesChunkWriter::GetOutEdgesChunkWriter(const GraphView& view,
 {
 }
 
+void GetOutEdgesChunkWriter::filterTombstones() {
+    // Base column of this ChunkWriter is _edgeIDs
+    bioassert(_edgeIDs);
+
+    TombstoneFilter filter(_view.tombstones());
```
rjb32: The TombstoneFilter is created from scratch at each call of the fill() method, which notably recreates the std::vector inside TombstoneFilter. Let's try to make TombstoneFilter a member of the chunk writer and reset its state each time so that we can reuse

storage/iterators/TombstoneFilter.h
```diff
@@ -59,17 +66,22 @@ class TombstoneFilter {
     };
 
     const Tombstones& _tombstones;
-    NonDeletedRanges _nonDeletedRanges;
+
+    // Pointer indirection to vector: only allocate vector if the owning ChunkWriter needs
+    // to be filter
+    std::unique_ptr<NonDeletedRanges> _nonDeletedRanges {nullptr};
```
rjb32: No need to default initialize to nullptr for a unique_ptr, it's done automatically by the default constructor of unique_ptr

storage/iterators/TombstoneFilter.cpp
```diff
@@ -23,9 +26,18 @@ TombstoneFilter::TombstoneFilter(const Tombstones& tombstones)
 template <TypedInternalID IDT>
 void TombstoneFilter::populateRanges(const ColumnVector<IDT>* baseCol) {
     bioassert(baseCol);
-    bioassert(_nonDeletedRanges.empty());
+    // TODO: Throw FatalException if !baseCol
+
+    // We use pointer indirection to avoid heap allocating a vector in each chunk writer
+    // if the filter is never needed. On first invocation of this function, create the
+    // vector.
+    if (!_nonDeletedRanges) {
```
rjb32, line 39: In the end you do an if on _nonDeletedRanges so the initialise flag is not really needed

storage/iterators/TombstoneFilter.h
cyrusknopf: Removed explicit initialisation

storage/iterators/TombstoneFilter.cpp
cyrusknopf: Removed `initialised` flag

storage/iterators/TombstoneFilter.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    // Failsafe
+    _nonDeletedRanges->clear();
+
+    const ColumnVector<IDT>& col = *baseCol;
+
+    size_t i = 0;
+    while (i < col.size()) { // Scan each element in the base column
+        // Skip until we find a non-deleted entry
+        bool deleted =  _tombstones.contains(col[i]);
```
rjb32: deleted can be const

pipeline/operations/ScanNodesStringApproxStep.cpp
```diff
@@ -49,6 +50,13 @@ void Step::execute() {
     // Fill _nodes with the matches of all datapart's indexes
     try {
         StringIndexUtils::getMatches<NodeID>(_nodes->getRaw(), _view, _pId, _strQuery);
+        if (_view.tombstones().hasNodes()) {
+            bioassert(_nodes);
```
LucLabarriere: Still considering using exceptions for this?

plan/QueryPlanner.cpp
```diff
@@ -380,6 +381,7 @@ void QueryPlanner::planScanNodes(const EntityPattern* entity) {
     const TypeConstraint* typeConstr = entity->getTypeConstraint();
     const ExprConstraint* exprConstr = entity->getExprConstraint();
 
+    // TOMBSTONES: base column @ref nodes always allocd on function entry
```
LucLabarriere, line 384: Are those comments necessary? I think they make sense for GetOutEdges etc, but if it's allocated anyway isn't it just noise?

plan/QueryPlanner.cpp
```diff
@@ -1444,13 +1477,16 @@ void QueryPlanner::planExpandEdgeWithTargetConstraint(const EntityPattern* edge,
     const ExprConstraint* edgeExprConstr = edge->getExprConstraint();
     const VarExpr* edgeVar = edge->getVar();
     VarDecl* edgeDecl = edgeVar ? edgeVar->getDecl() : nullptr;
-    const bool mustWriteEdges = edgeDecl && edgeDecl->isUsed();
+
+    const bool needFilter = _view.tombstones().hasEdges() || _view.tombstones().hasNodes();
```
LucLabarriere: Is it necessary to check if there are tombstones on the nodes here?

storage/iterators/ScanEdgesIterator.cpp
```diff
@@ -160,6 +183,10 @@ void ScanEdgesChunkWriter::fill(size_t maxCount) {
         CASE(14);
         CASE(15);
     }
+
+    if (_view.tombstones().hasEdges() || _view.tombstones().hasNodes()) {
+        filterTombstones();
+    }
```
LucLabarriere: Just checking hasEdges should be enough?

storage/iterators/ScanOutEdgesByLabelIterator.cpp
```diff
@@ -189,6 +213,10 @@ void ScanOutEdgesByLabelChunkWriter::fill(size_t maxCount) {
         CASE(14);
         CASE(15);
     }
+
+    if (_view.tombstones().hasNodes() || _view.tombstones().hasEdges()) {
+        filterTombstones();
+    }
```
LucLabarriere, line 219: Only check hasEdges?

pipeline/operations/ScanNodesStringApproxStep.cpp
cyrusknopf: There is an exception thrown at the start of populateRanges with the same check, so I'll just remove the assert here

plan/QueryPlanner.cpp
cyrusknopf: I added them for consistency: if some steps have documentation as to if their base column is allocated, then a step not having that documentation could be due to it being overlooked. So the explicit comment ensures it is clear that it has not been overlooked.

@rjb32 what do you think? is it too noisy?

rjb32: This one is good because this is to say that this is allocated before entering the function

storage/versioning/Tombstones.h
cyrusknopf: Good catch, removed

plan/QueryPlanner.cpp
cyrusknopf: Fixed

storage/iterators/ScanEdgesIterator.cpp
cyrusknopf: Fixed

storage/iterators/ScanOutEdgesByLabelIterator.cpp
cyrusknopf: Fixed

PR #82, opened 2025-10-27
Implement simple dataframes to be used in v2 Pipeline

rjb32:
Simple Dataframe implementation to be used as replacement for Block in v2 pipeline.

A Dataframe can contain a sequence of columns of possibly different length, with a ColumnName associated to each column. Dataframe becomes the owner of NamedColumn objects that are light wrappers around a Column* and a ColumnName.

ColumnName can be either a string or an "anonymous" name which is a numerical tag. Because we will probably have a lot of recurring names in queries, such as "a", "e", "p"..etc, strings used in ColumnName are deduplicated and managed by a ColumnNameManager to have shared strings in the same thread and hopefully recycle strings between queries.

### DIFF
--- a/storage/dataframe/Dataframe.h
+++ b/storage/dataframe/Dataframe.h
@@ -9,6 +9,11 @@ namespace db {
 
 class NamedColumn;
 
+// A basic Dataframe class where columns are indexed by numerical ColumnTag.
+// Columns can have optional names as string_views, but only tags can be used
+// to uniquely retrieve a column.
+// A Dataframe can possibly have several columns with the same name strings.
+// Uniqueness is only enforced on ColumnTags.
 class Dataframe {
 public:
     using NamedColumns = std::vector<NamedColumn*>;


storage/CMakeLists.txt
```diff
@@ -34,6 +34,10 @@ set(storage_sources
 
         columns/Block.cpp
         columns/Column.cpp
+        columns/ColumnNameManager.cpp
+        columns/Dataframe.cpp
+        columns/DataframeHeaderMap.cpp
+        columns/NamedColumn.cpp
```
LucLabarriere: Is there a plan to use those DataFrame/NamedColumns somewhere in storage? (Like in iterators and chunk writers directly for example)

storage/columns/ColumnName.h
```diff
@@ -0,0 +1,90 @@
...
+
+    ~ColumnName() = default;
+
+    bool operator==(const ColumnName& other) const {
+        if (_sharedStr) {
+            return _sharedStr == other._sharedStr;
+        } else {
+            return _tag == other._tag;
+        }
+    }
```
LucLabarriere: Would it possible to make ColumnName just a mandatory Tag + an optional string (instead of either a tag or a string). This way we would check only the tag in operator==, and the std::hash implementation would just return the tag

storage/CMakeLists.txt
rjb32: Not at the moment. Iterators for now should continue to work on primitive column objects I think

storage/columns/ColumnNameManager.cpp
```diff
@@ -0,0 +1,28 @@
...
+    const std::string sstr(str);
+    const auto sharedStrIt = _strMap.find(sstr);
+    if (sharedStrIt != _strMap.end()) {
+        return sharedStrIt->second;
+    }
+
+    ColumnNameSharedString* sharedStr = new ColumnNameSharedString(std::string(str));
+    _strings.push_back(sharedStr);
+    _strMap[std::move(sstr)] = sharedStr;
+    return sharedStr;
```
LucLabarriere: So we make a copy of the string to compare? Maybe a custom Hash + Predicate impl would allow to compare the string_view directly.

storage/columns/ColumnNameManager.h
```diff
@@ -0,0 +1,30 @@
...
+    ColumnName getName(std::string_view str) {
+        return ColumnName(getString(str));
+    }
+
+    ColumnNameSharedString* getString(std::string_view str);
+
+private:
+    std::vector<ColumnNameSharedString*> _strings;
+    std::unordered_map<std::string, ColumnNameSharedString*> _strMap;
+};
```
LucLabarriere: I'm a bit worried of the cost of this, as you say strings might be small strings reused over and over again, but if they're really small, they will be stack allocated due to small string optimization (up to 15 characters I believe).

storage/columns/Dataframe.h
```diff
@@ -0,0 +1,47 @@
...
+public:
+    using NamedColumns = std::vector<NamedColumn*>;
+    friend NamedColumn;
+
+    Dataframe();
+
+    Dataframe(Dataframe&& other)
+        : _cols(std::move(other._cols))
+    {
+    }
```
LucLabarriere, line 28: headerMap is not moved?

storage/dataframe/Dataframe.h
```diff
@@ -0,0 +1,47 @@
...
+
+    Dataframe();
+
+    Dataframe(Dataframe&& other)
+        : _cols(std::move(other._cols))
+    {
+    }
+
+    Dataframe& operator=(Dataframe&& other) {
+        _cols = std::move(other._cols);
```
LucLabarriere, line 32: Same thing, _headerMap not moved?

storage/columns/Dataframe.h
rjb32: Good one thank you

storage/dataframe/Dataframe.h
rjb32: Good one thanks

storage/columns/ColumnNameManager.cpp
rjb32: So the issue is that we need to store strings as keys in the std::unordered_map that we are sure will stay alive, possibly for longer that a given query lifetime. Because in case of bucket collision in the unordered_map the hash table implementation will potentially need to do key comparisons. So this is why I can't use std::string_view as keys because the views are not alive across queries in general.

storage/columns/ColumnNameManager.h
rjb32: Yes indeed, string deduplication is a debate among performance people for these reasons and the worthyness for small strings. Some people do it, some people don't bother with it and trust and/or rely on std::string implementation for short strings.

rjb32: If we put an std::string directly as a member of each NamedColumn, the NamedColumn will be allocated in the heap  and so any local small array in the string will be inside the heap anyway

LucLabarriere: I was surprised because I was thinking simpler (just a uint64_t id in the column, + a map name -> id), but I trust you on the design

PR #84, opened 2025-10-28
Added support for the SET statement in the query language

LucLabarriere:
For now, only `SET n.name = "..."` expressions are supported, more complex set commands are not:
- `SET n:Person`: add a label `Person` to node `n` (we don't support adding labels for now)
- `SET n += { name: "...", age: 20 }`: Updating node properties dynamically, should not be too hard to handle later

### DIFF
--- a/v2/AST/stmt/SetItem.h
+++ b/v2/AST/stmt/SetItem.h
@@ -37,10 +37,6 @@ public:
     Variant& item() { return _item; }
     const Variant& item() const { return _item; }
 
-
-    template<class... Ts>
-    struct Overloaded : Ts... { using Ts::operator()...; };
-
 private:
     friend CypherAST;
 
--- a/v2/analyzer/WriteStmtAnalyzer.cpp
+++ b/v2/analyzer/WriteStmtAnalyzer.cpp
@@ -5,6 +5,7 @@
 #include "AnalyzeException.h"
 #include "CypherAST.h"
 #include "DiagnosticsManager.h"
+#include "Overloaded.h"
 
 #include "Literal.h"
 #include "EdgePattern.h"
@@ -249,7 +250,7 @@ void WriteStmtAnalyzer::analyze(EdgePattern* edgePattern) {
 }
 
 void WriteStmtAnalyzer::analyze(SetItem* item) {
-    const auto visitor = SetItem::Overloaded {
+    const auto visitor = Overloaded {
         // PropertyExprAssign case
         [this, item](const SetItem::PropertyExprAssign& v) {
             // Analyzing rhs
--- a/v2/plan/WriteStmtGenerator.cpp
+++ b/v2/plan/WriteStmtGenerator.cpp
@@ -3,6 +3,7 @@
 #include <spdlog/fmt/bundled/format.h>
 
 #include "CypherAST.h"
+#include "Overloaded.h"
 #include "DiagnosticsManager.h"
 #include "Pattern.h"
 #include "PatternElement.h"
@@ -77,7 +78,7 @@ void WriteStmtGenerator::generateSetStmt(const SetStmt* stmt, PlanGraphNode* pre
     prepareWriteNode(prevNode);
 
     for (const SetItem* item : stmt->getItems()) {
-        const auto visitor = SetItem::Overloaded {
+        const auto visitor = Overloaded {
             // PropertyExprAssign case
             [this](const SetItem::PropertyExprAssign& v) {
                 const VarDecl* decl = v.propTypeExpr->getDecl();


cyrusknopf:
Is there any plan on how to support this? I don't see how it's possible without recreating dataparts as with the original design of the deletes

LucLabarriere:
> Is there any plan on how to support this? I don't see how it's possible without recreating dataparts as with the original design of the deletes

Yeah exactly, that's a big question, I think it echoes what you said the other day, we can always delete the node and recreate one with the updated prop? At least it's there in the Parser/Planner and we can disable the execution for now (throw Not Implemented)

cyrusknopf:
> > Is there any plan on how to support this? I don't see how it's possible without recreating dataparts as with the original design of the deletes
> 
> Yeah exactly, that's a big question, I think it echoes what you said the other day, we can always delete the node and recreate one with the updated prop? At least it's there in the Parser/Planner and we can disable the execution for now (throw Not Implemented)

Ah yep, I completely forgot about deleting and recreating, I think that is the only feasible idea - creating a whole new datapart is far too expensive as we saw. We would also have to recreate any edges incident to the modified node, though.

LucLabarriere:
> > > Is there any plan on how to support this? I don't see how it's possible without recreating dataparts as with the original design of the deletes
> > 
> > 
> > Yeah exactly, that's a big question, I think it echoes what you said the other day, we can always delete the node and recreate one with the updated prop? At least it's there in the Parser/Planner and we can disable the execution for now (throw Not Implemented)
> 
> Ah yep, I completely forgot about deleting and recreating, I think that is the only feasible idea - creating a whole new datapart is far too expensive as we saw. We would also have to recreate any edges incident to the modified node, though.

Oh yeah that last point will be really tricky

v2/AST/stmt/SetItem.h
```diff
@@ -0,0 +1,55 @@
...
+class CypherAST;
+class PropertyExpr;
+class Expr;
+class Symbol;
+class EntityTypeExpr;
+
+class SetItem {
+public:
+    struct PropertyExprAssign {
+        static constexpr std::size_t index = 0;
```
rjb32: How this index static constexpr being used?

v2/analyzer/ExprAnalyzer.cpp
```diff
@@ -285,12 +285,18 @@ void ExprAnalyzer::analyze(PropertyExpr* expr) {
         // Property does not exist yet
 
         std::string_view name = propName->getName();
```
rjb32: name can be const

v2/analyzer/ExprAnalyzer.cpp
```diff
@@ -372,6 +380,28 @@ void ExprAnalyzer::analyze(PathExpr* expr) {
     throwError("Path expressions not supported", expr);
 }
 
+void ExprAnalyzer::addToBeCreatedType(std::string_view name, ValueType type, const void* obj) {
+    auto it = _toBeCreatedTypes.find(name);
```
rjb32: it can be const

v2/plan/nodes/WriteNode.h
```diff
@@ -24,13 +34,11 @@ class WriteNode : public PlanGraphNode {
         EdgeNeighbour() = default;
 
         explicit EdgeNeighbour(const VarDecl* var)
-            : _data(var)
-        {
+            : _data(var) {
         }
 
         explicit EdgeNeighbour(size_t offset)
```
rjb32: Please comment on what the different constructors are to be used for

v2/AST/stmt/SetItem.h
LucLabarriere: This is not great, it's used in a switch on the type contained in the variant, I can see how to handle that better

LucLabarriere:
@rjb32 fixed some of the issues
- Removed the static constexpr size_t index in the structs of the SetItem class, they were used to do a switch on the type in the variant, but that's a bit ugly. I tried an approach with the visitor pattern but it may look a bit obscure. I got the approach from here: [ccppreference](https://en.cppreference.com/w/cpp/utility/variant/visit):
```c++
const auto visitor = overloads
    {
        [](int i){ [std::print](https://en.cppreference.com/w/cpp/io/print.html)("int = {}\n", i); },
        []([std::string_view](https://en.cppreference.com/w/cpp/string/basic_string_view.html) s){ [std::println](https://en.cppreference.com/w/cpp/io/println.html)("string = “{}”", s); },
        [](const Base&){ [std::println](https://en.cppreference.com/w/cpp/io/println.html)("base"); }
    };
```

If it's too ugly, the last resort solution is an if {} else if {} else if {} chain on the type of the SetItem, I guess?

- Cleaned up the WriteNode class, it should be simpler now (also fixed a bug)

v2/AST/stmt/SetItem.h
```diff
@@ -0,0 +1,53 @@
...
+                                 SymbolAddAssign,
+                                 SymbolEntityTypes>;
+
+    static SetItem* create(CypherAST* ast, PropertyExpr* expr, Expr* value);
+    static SetItem* create(CypherAST* ast, EntityTypeExpr* value);
+    static SetItem* create(CypherAST* ast, Symbol* symbol, Expr* value);
+
+    Variant& item() { return _item; }
+    const Variant& item() const { return _item; }
+
```
rjb32, line 39: Too many blank lines

v2/AST/stmt/SetItem.h
```diff
@@ -0,0 +1,53 @@
...
+
+    template<class... Ts>
+    struct Overloaded : Ts... { using Ts::operator()...; };
```
rjb32: Define in a common header

v2/analyzer/WriteStmtAnalyzer.cpp
```diff
@@ -238,6 +248,41 @@ void WriteStmtAnalyzer::analyze(EdgePattern* edgePattern) {
     }
 }
 
+void WriteStmtAnalyzer::analyze(SetItem* item) {
+    const auto visitor = SetItem::Overloaded {
+        // PropertyExprAssign case
+        [this, item](const SetItem::PropertyExprAssign& v) {
```
rjb32, line 255: This is looking clean I like it

PR #85, opened 2025-10-28
Refactor Serialisation tests to use TuringTestEnv

cyrusknopf:


cyrusknopf:
Hi @LucLabarriere I'm looking at migrating the few serialisation tests to use the TuringTestEnv and have fewer dependencies. When you get a change could you look at the test I've changed and see if it is what you had in mind? (I think you mentioned you wanted to see a refactor of the serialisation tests in a PR I had for serialising the CommitJournal). If not let me know and I can remove other dependencies/make any other changes.

This is not urgent so only when you have a free moment.

test/serialisation/StringIndexSerialisationTest.cpp
```diff
@@ -1,67 +1,68 @@
...
 #include "SystemManager.h"
-#include "indexers/StringPropertyIndexer.h"
-#include "reader/GraphReader.h"
-#include "versioning/Transaction.h"
-#include "views/GraphView.h"
 #include "dump/GraphLoader.h"
 #include "TuringConfig.h"
+#include "TuringTestEnv.h"
 #include "SimpleGraph.h"
+#include "indexers/StringPropertyIndexer.h"
```
LucLabarriere, line 15: I thihnk this header is unused

test/serialisation/StringIndexSerialisationTest.cpp
```diff
@@ -1,67 +1,68 @@
...
         }
 
-        _loadedGraph = Graph::create();
-        const auto loadRes = GraphLoader::load(_loadedGraph.get(), _workingPath);
+        Graph* loadedGraph = sysMan.createGraph(_loadedGraphName);
+        const auto loadRes = GraphLoader::load(loadedGraph, _workingPath);
         if (!loadRes) {
             throw TuringException("Failed to dump graph:\n" + res.error().fmtMessage());
         }
     }
```
LucLabarriere, line 58: See if you can use SystemManager::dumpGraph instead of the GraphDump/GraphLoader classes directly instead

test/serialisation/StringIndexSerialisationTest.cpp
cyrusknopf: Used on line 82 to call a member function of StringIndexer

LucLabarriere: Oh I see, this is part of the things I wan't a fan of, I think testing the dump of specific parts of a graph like this should be done in storage/dump and that higher level unit test like serialization should just test with the higher level API (TuringDB + query language). But it's fine for this PR

test/serialisation/StringIndexSerialisationTest.cpp
cyrusknopf: @LucLabarriere I'm not sure I can at the moment: I can create a graph with name "simple" with SysMan, then dump it with SysMan, but that dumps it to a `simple/` directory. So then if I try and load it I have to specify I want to load "simple", but since that graph already exists, it fails. (I think that is what is going on). One way around this would be to create another test env with a different system manager that doesn't see the original "simple" graph, but that seems hacky

LucLabarriere: Ok yeah I see, honestly don't worry about it for now

PR #87, opened 2025-10-30
Implement The Merges In Change System

ghost:
1. Add DATA_PART_MERGE query
2. Allow A Change To Create A Commit With 1 Merged Data Part
3. Handle A Merged Commit When Dumping And Loading A File

### DIFF
--- a/storage/CMakeLists.txt
+++ b/storage/CMakeLists.txt
@@ -80,6 +80,7 @@ set(storage_sources
         versioning/WriteSet.cpp
         versioning/TombstoneSet.cpp
         versioning/VersionControlException.cpp
+        versioning/TombstoneRanges.cpp
 
         comparators/DataPartComparator.cpp
         comparators/StringIndexerComparator.cpp
--- a/storage/DataPart.cpp
+++ b/storage/DataPart.cpp
@@ -65,8 +65,11 @@ bool DataPart::load(const GraphView& view, JobSystem& jobSystem, DataPartBuilder
     }
 
     // Storing temp node IDs to keep track of the sorting mechanism
-    std::vector<NodeID> tmpNodeIDs(coreNodeLabelSets.size());
-    std::iota(tmpNodeIDs.begin(), tmpNodeIDs.end(), firstTmpNodeID);
+    std::vector<NodeID>& tmpNodeIDs = builder.getTmpNodeIDs();
+    if (tmpNodeIDs.empty()) {
+        tmpNodeIDs.resize(coreNodeLabelSets.size());
+        std::iota(tmpNodeIDs.begin(), tmpNodeIDs.end(), firstTmpNodeID);
+    }
 
     // Sorting based on the labelset
     rg::sort(rv::zip(coreNodeLabelSets, tmpNodeIDs),
--- a/storage/iterators/TombstoneFilter.h
+++ b/storage/iterators/TombstoneFilter.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "versioning/Tombstones.h"
+#include "versioning/NonDeletedRanges.h"
 #include "columns/ColumnVector.h"
 #include "ID.h"
 
@@ -57,14 +58,6 @@ public:
     }
 
 private:
-    struct NonDeletedRange;
-    using NonDeletedRanges = std::vector<NonDeletedRange>;
-
-    struct NonDeletedRange {
-        size_t start;
-        size_t size;
-    };
-
     const Tombstones& _tombstones;
 
     // Pointer indirection to vector: only allocate vector if the owning ChunkWriter needs
--- a/storage/mergers/DataPartMerger.cpp
+++ b/storage/mergers/DataPartMerger.cpp
@@ -1,16 +1,21 @@
 #include "DataPartMerger.h"
 
+#include <range/v3/view/enumerate.hpp>
+
 #include "DataPart.h"
 #include "writers/DataPartBuilder.h"
 #include "versioning/CommitBuilder.h"
 #include "EdgeContainer.h"
 #include "NodeContainer.h"
+#include "versioning/TombstoneRanges.h"
 #include "properties/PropertyManager.h"
-#include "ID.h"
 #include "reader/GraphReader.h"
-#include "versioning/VersionController.h"
+
+#include "ID.h"
 
 using namespace db;
+namespace rg = ranges;
+namespace rv = rg::views;
 
 struct PropertyTypeDispatcher {
     ValueType _valueType;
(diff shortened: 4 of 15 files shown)


storage/writers/MetadataBuilder.cpp
```diff
@@ -40,7 +40,6 @@ std::unique_ptr<MetadataBuilder> MetadataBuilder::create(const GraphMetadata& pr
 
     auto* ptr = new MetadataBuilder;
     ptr->_metadata = metadata;
-    *ptr->_metadata = prevMetadata;
```
ghost, line 43: I noticed that we are copying the metadata twice.

```c++
std::unique_ptr<Commit> Commit::createNextCommit(VersionController* controller,
                                                 const WeakArc<CommitData>& data,
                                                 const CommitView& prevCommit) {
    auto* ptr = new Commit {controller, data};

    // Copy previous commit history and metadata
    ptr->_data->_history.newCommitHistoryFromPrevious(prevCommit.history());
    ptr->_data->_history.pushCommit(ptr->view());
    ptr->_data->_metadata = prevCommit.metadata();

    return std::unique_ptr<Commit> {ptr};
}
```
Which is called from [here.](https://github.com/turing-db/turingdb/blob/e329ac5ef80cf085bedd98e98647d2feef22c42c/storage/versioning/CommitBuilder.cpp#L162)

storage/versioning/Commit.cpp
```diff
@@ -20,6 +20,13 @@ bool Commit::isHead() const {
     return _controller->getHeadHash() == _hash;
 }
 
+bool Commit::isMergeCommit() const {
```
cyrusknopf: I don't think this is correct:
```
turing:default> history
Query executed in 0.197529 ms.
+----------------------------------+
| Commit: 5d71f3edd5579df6 (HEAD)  |
+----------------------------------+
turing:default> change new
Query executed in 0.10918 ms.
+---+
| 0 |
+---+
turing:default> checkout change-0
turing:default@0> change submit
[2025-10-30 18:50:31] [info] Dumping graph default
Change '0' does not exist anymore, switching back to head
Query executed in 87.2709 ms.

turing:default> history
Query executed in 0.07058 ms.
+----------------------------------+
| Commit: 5d71f3edd5579df6         |
+----------------------------------+
| Commit: 6fc5f5442e0b5f03 (HEAD)  |
+----------------------------------+
turing:default>
```

Not first commit, but no edges or nodes, so first IDs are both 0

storage/versioning/CommitBuilder.cpp
```diff
@@ -130,6 +135,11 @@ CommitBuilder::CommitBuilder(VersionController& controller, Change* change, cons
 {
 }
 
+void CommitBuilder::removeDataparts() {
+    // TODO: Locking
+    _commitData->history()._allDataparts = {};
```
cyrusknopf: Is `.clear()` more purposeful?

system/SystemManager.cpp
```diff
@@ -417,6 +417,15 @@ ChangeResult<Change*> SystemManager::newChange(const std::string& graphName, Com
     return _changes->storeChange(graph, std::move(change));
 }
 
+void SystemManager::mergeDataParts(Graph* graph, JobSystem& jobSystem) {
+    // We do not currently perform a merge if there are active changes.
+    graph->lockMainState();
```
cyrusknopf: When do we release this lock?

storage/dump/CommitDumper.cpp
```diff
@@ -119,6 +119,19 @@ DumpResult<void> CommitDumper::dump(const Commit& commit, const fs::Path& path)
...
+        Profile profile {"CommitDumper::dump <merge>"};
+        if (commit.isMergeCommit()) {
+            const fs::Path mergePath = path / "merge";
+            auto writerRes = fs::FilePageWriter::open(mergePath, DumpConfig::PAGE_SIZE);
+            if (!writerRes) {
+                return DumpError::result(DumpErrorType::CANNOT_OPEN_JOURNAL, writerRes.error());
+            }
+        }
+    }
+
```
LucLabarriere, line 154: I think that's a pretty good way to handle this for now (the simplest I think?)

Check the error `CANNOT_OPEN_JOURNAL` though, I think you forgot to change it?

storage/versioning/Commit.cpp
LucLabarriere: What do you guys think of adding metadata to the Commit class directly? Originally I was planning on having a timestamp, possibly author, description, etc. Not sure we still want this, but we could still have a flag "isMerge" or something like that.

storage/versioning/VersionController.cpp
```diff
@@ -36,6 +38,41 @@ void VersionController::createFirstCommit() {
...
+
+    auto buildRes = newTip->build(jobSystem);
+    if (!buildRes) {
+        return buildRes.get_unexpected();
+    }
+    addCommit(std::move(buildRes.value()));
+
+    return {};
+}
+
```
LucLabarriere: I feel like the poor API of CommitBuilder kind of gets in your way there, what about making your own MergeCommitBuilder? You could start clean and if we rework the CommitBuilder, it won't impact your changes

system/SystemManager.cpp
LucLabarriere: Here the lock is released immediately. The signature is:

```c++
std::unique_lock<std::mutex> lockMainState();
```

At the very least, it needs to be:
```c++
[[nodiscard]] std::unique_lock<std::mutex> lockMainState();
```

But here we must find a better way, we can't rely on exposing the mutex like this, we need more restrictive mechanisms

system/SystemManager.cpp
```diff
@@ -417,6 +417,15 @@ ChangeResult<Change*> SystemManager::newChange(const std::string& graphName, Com
...
+    if (!_changes->isEmpty()) {
+        return;
+    }
+    return graph->mergeDataParts(jobSystem);
```
LucLabarriere: This is a race condition.

To solve all these issues, I think the whole function should be in the ChangeManager class directly

cyrusknopf: Out of interest what is the race condition on? I assumed if the "graph lock" was taken correctly then the merge would be the only worker on the graph.

LucLabarriere: If everything's perfect I think it would be fine but here the lock is taken on the graph itself, and then we check the ChangeManager which are not protected by the same locks. If we can create changes from other places than the ChangeManager or w/e, then it's a race condition.

Also I'm just realizing that if there are changes on other graphs, they forbid to merge on any graph?

LucLabarriere: What about putting all of this in a single method in ChangeManager, you lock the ChangeManager, check if there are any changes on the graph right now, return early if there are. Else, try to create a merge data part by calling graph->mergeDataparts(), which calls versionController->mergeDataparts(), which locks the graph then do the merge. It's super simple and maybe not what we want later on but at least it works (hopefully) and is safe for the release?

cyrusknopf: Ah yes you are right. We would have to check the GraphChangePairs to check if there are any Changes on the current graph, right?

LucLabarriere: Yeah I think so... Hopefully this crap goes away soon ahah

storage/versioning/VersionController.cpp
LucLabarriere: In the end you don't need the MergeCommitBuilder to have a vector of pending dataparts, you don't need it to be tied to a change, etc.

storage/versioning/Commit.cpp
cyrusknopf: Yeah could fit into CommitData or a new CommitMetaData

system/SystemManager.cpp
ghost: ah yes that's a good catch with the race there. I think this will be simplified a lot if we move change manager into Graph.

system/SystemManager.cpp
rjb32: This is not very good to expose the mutex like this. Why is this lockMainState() even necessary?

storage/versioning/VersionController.cpp
ghost: I didn't go for a separate merge commit builder class in the end, because we use a lot of private members of other classes directly through a friend membership. I didn't want to simply add more friend relations, so I added prepareMerge and initialiseMerge functions.

storage/versioning/CommitBuilder.cpp
ghost: I've removed this function mostly, I felt it would be better to have our own createMergedCommit() function.

system/ChangeManager.cpp
```diff
@@ -20,6 +20,26 @@ Change* ChangeManager::storeChange(const Graph* graph, std::unique_ptr<Change> c
 
     return ptr;
 }
+Change* ChangeManager::createChange(Graph* graph, CommitHash baseHash) {
```
rjb32, line 15: I like this for now it's good

storage/versioning/CommitBuilder.cpp
```diff
@@ -159,3 +172,33 @@ void CommitBuilder::initialize() {
     // Copy tombstones from previous commit
     _commitData->_tombstones = prevCommit.tombstones();
 }
+
+void CommitBuilder::initializeMerge() {
+    Profile profile {"CommitBuilder::initialize"};
+
+    auto reader = _view.read();
+
+    // The first ID of this commit will be one more than the max ID in the graph
```
cyrusknopf: These comments (and the ones below) are left over and don't make sense here I don't think?

storage/versioning/CommitHistory.cpp
```diff
@@ -19,3 +19,8 @@ void CommitHistory::newCommitHistoryFromPrevious(const CommitHistory& previous)
     _commitDataparts = {};
 }
 
+void CommitHistory::newCommitHistoryFromPreviousCommits(const CommitHistory& previous) {
```
cyrusknopf: I think this function should be called something else, its way to similar to the one above, whilst doing something very different.

storage/versioning/CommitBuilder.cpp
ghost: Ah yes, will remove it!

storage/versioning/CommitHistory.cpp
ghost: yeah, I was struggling a bit to name it as well, do you have any suggestions (for something that only copies over the commitHistory)

cyrusknopf: Yeah you're right naming this is tricky, maybe something like `newEmptyHistory` or `newMergeCommitHistory` with plenty of comments to explain why we are copying `CommitHistory` but not `allDataParts`?

storage/Graph.h
```diff
@@ -1,12 +1,14 @@
 #pragma once
 
 #include <memory>
+#include <mutex>
```
LucLabarriere: Do you need this header?

storage/Graph.h
```diff
@@ -41,6 +43,7 @@ class Graph {
 
     [[nodiscard]] std::unique_ptr<Change> newChange(CommitHash base = CommitHash::head());
     [[nodiscard]] FrozenCommitTx openTransaction(CommitHash hash = CommitHash::head()) const;
+    DataPartMergeResult<void> mergeDataParts(JobSystem& jobSystem);
```
LucLabarriere: Maybe add [[nodiscard]] ? (still not sure how to feel on whether to add them or not). What do you think?

storage/versioning/Commit.h
```diff
@@ -21,6 +21,7 @@ class Commit {
 
     Commit();
     Commit(VersionController* controller, const WeakArc<CommitData>& data);
+    Commit(VersionController* controller, const WeakArc<CommitData>& data, bool isMergeCommit);
```
LucLabarriere: Maybe just add bool isMergeCommit = false instead of creating a new constructor. Or maybe you don't like this style? Do we prefer overloads or having defaults for arguments?

storage/versioning/CommitBuilder.cpp
```diff
@@ -41,6 +49,11 @@ GraphReader CommitBuilder::readGraph() const {
     return viewGraph().read();
 }
 
+void CommitBuilder::insertBuilder(std::unique_ptr<DataPartBuilder> builder) {
```
LucLabarriere: Maybe rename, here it seems to imply you're adding the DataPartBuilder in the middle of the container

storage/versioning/CommitHistory.cpp
LucLabarriere: It makes sense that it's hard to name because I realise now that dataparts shouldn't really be stored in CommitHistory... CommitHistory should just be list of commits that happened before (imo). Dataparts are data that should be stored somewhere else.

In the meantime, what about:
- fillFromPrevious
- fillMergedFromPrevious

- fillFromPrevious
- newEmptyHistoryForMerge

(not great tbh)

system/ChangeManager.cpp
```diff
@@ -20,6 +20,26 @@ Change* ChangeManager::storeChange(const Graph* graph, std::unique_ptr<Change> c
...
+    std::unique_lock guard(_changesLock);
+
+    auto change = graph->newChange(baseHash);
+    const auto id = change->id();
+    auto* ptr = change.get();
+    _changes.emplace(GraphChangePair {graph, id}, std::move(change));
+
+    return ptr;
```
LucLabarriere, line 23: Yeah this is nicer than the store method I think, will make even more sense once ChangeManager is in Graph directly

system/ChangeManager.h
```diff
@@ -42,12 +43,17 @@ class ChangeManager {
     ChangeManager& operator=(ChangeManager&&) = delete;
 
     Change* storeChange(const Graph* graph, std::unique_ptr<Change> change);
```
LucLabarriere: Still needed?

storage/Graph.h
ghost: ah yep shouldn't need this because we removed the lockMainState function

storage/Graph.h
ghost: That is a good catch, I like the pattern personally - it's good to force users to handle/think about results imo.

storage/versioning/CommitBuilder.cpp
ghost: ah yes, I could do appendBuilder.

system/ChangeManager.h
ghost: Nope, I'll clear it out!

storage/versioning/CommitHistory.cpp
ghost: I did want to avoid using the merge keyword, because I thought having a commit independent of any previous data parts could be something useful in the future.

ghost: but I think it's a good idea to move data parts out of commit history in the future, so I'll just go with the merge naming for now.

pipeline/operations/DataPartMergeStep.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include "ExecutionContext.h"
+#include "PipelineException.h"
+#include "SystemManager.h"
+#include "Graph.h"
+
+using namespace db;
+
+void DataPartMergeStep::prepare(ExecutionContext* ctxt) {
+    _systemManager = ctxt->getSystemManager();
```
cyrusknopf, line 13: Are we having merges require being in a change themselves? Here we are not checking if we are in a write transaction. No sure if we need to or if the merge is a level higher

ghost: Merges can work independently of changes (and thus transactions), so we don't have to check for that.

test/query/MergeQueryTest.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+protected:
+    Graph* _graph;
+    std::unique_ptr<TuringTestEnv> _env;
+    std::unique_ptr<QueryInterpreter> _interp {nullptr};
+};
+
+TEST_F(MergeQueryTest, TestMerge) {
+    QueryTester tester {_env->getMem(), *_interp};
+    GraphWriter oldWriter = GraphWriter {_graph};
```
cyrusknopf: I think you can do
```
const auto transaction = _graph->openTransaction();
const auto reader = transaction.readGraph();
```
so you don't need `GraphWriter`.

test/query/MergeQueryTest.cpp
```diff
@@ -0,0 +1,64 @@
...
+    tester.query("data_part_merge").execute();
+
+    GraphWriter newWriter = GraphWriter {_graph};
+    const GraphReader newReader = newWriter.openWriteTransaction().readGraph();
+    const auto newNodeCount = newReader.getNodeCount();
+    const auto newEdgeCount = newReader.getEdgeCount();
+
+    ASSERT_NE(prevNumDataParts, 1);
+    ASSERT_EQ(prevNumCommits + 1, newReader.commits().size());
+    ASSERT_EQ(oldNodeCount, newNodeCount);
```
cyrusknopf, line 61: I've just realised this will fail with deletes, as `oldNodeCount` will include the tombstoned nodes, whilst `newNodeCount` will not because they will not have been added to the merged DP. 

I will fix this in a PR tomorrow, just letting you know in case your tests start failing.

storage/Graph.h
LucLabarriere: Did you add [[nodiscard]] in the end?

storage/dump/CommitDumper.cpp
```diff
@@ -137,7 +137,19 @@ DumpResult<void> CommitDumper::dump(const Commit& commit, const fs::Path& path)
         if (auto res = dumper.dump(tombstones); !res) {
             return res;
         }
-        
+    }
+
+    // Dumping Merge File
+    // Existence Of File Confirms If We Have A Merge Commit
```
cyrusknopf, line 143: Since you modify dump/load don't forget (like I did) to update `DumpConfig::UP_TO_DATE_VERSION`

LucLabarriere: I was thinking, does he need to update the version here? Because if its an older binary version, there's no merge file so they should still load fine.

This is actually a general question on when we should update the UP_TO_DATE_VERSION. I can see several strategies:

1. When we change anything in the whole project (basically, each new commit invalidates the previous binaries)
2. When we change anything in the dump/load library + any change that we think break the binary format (including changes in storage, like sizeof(PropertyTypeID)
3. When we think older binaries are loadable anymore
4. When we make only the previous release binaries not loadable anymore

What do you guys think? Generally speaking I think option 3 makes more sense, but options 2/3 are safer (because we might assume our changes didn't change anything when actually they did)

If we ever have a binary converter mechanism (I thought of a way to do this), option 1 is possible

cyrusknopf: True, perhaps it is fine in this case however, on your question:

I think options 1 or 2 are the best, because 3 and 4 require us "thinking" older binaries are not loadable anymore - which is prone to human error. For example here I thought that this change would make older binaries not loadable, but I was wrong. It is possible in future we assume that a previous version is compatible, but in fact it is not.

With regards to Option 1, is it not the case that if we make no changes to Load/Dumping, and can prove that all Load/Dump tests using GraphComparators remain passing, then previous binaries will still be loadable?

Option 2 is probably the safest without being overly defensive.

What do we think?

ghost: Can't we (and shouldn't we ) have some sort of test that saves simpledb  (or a bit more complex graph to include deletes/merges/etc) of the latest commit into main and we have CI to test if that can be loaded into the latest turingdb version?

ghost: Other than that I think option 2 is a pretty safe middle ground.

rjb32: Yes Option 2 is the safest today

cyrusknopf: Does `samples/graph-loader-dumper` do what you mean?

LucLabarriere: > True, perhaps it is fine in this case however, on your question:
> 
> I think options 1 or 2 are the best, because 3 and 4 require us "thinking" older binaries are not loadable anymore - which is prone to human error. For example here I thought that this change would make older binaries not loadable, but I was wrong. It is possible in future we assume that a previous version is compatible, but in fact it is not.
> 
> With regards to Option 1, is it not the case that if we make no changes to Load/Dumping, and can prove that all Load/Dump tests using GraphComparators remain passing, then previous binaries will still be loadable?
> 
> Option 2 is probably the safest without being overly defensive.
> 
> What do we think?

The issue is that Load/Dump unit tests would have to check against older versions of binaries, which I do think do and would be hard to maintain



> Can't we (and shouldn't we ) have some sort of test that saves simpledb (or a bit more complex graph to include deletes/merges/etc) of the latest commit into main and we have CI to test if that can be loaded into the latest turingdb version?

Yeah we could but that might be really hard to maintain.

I assume we can start with option 2 for now. Later on we could also have another level of check, we would have:
```c++
UP_TO_DATE_VERSION // Points to the latest version known to be compatible
CURRENT_VERSION    // Points to the current version

if (binaryVersion < UP_TO_DATE_VERSION) {
    // Error, we know that this is incompatible
} else if (binaryVersion > UP_TO_DATE_VERSION) {
    // Warning only, might be incompatible, but try to load anyway because
    // we didn't update UP_TO_DATE_VERSION meaning we think it's still valid
} else {
    // Load normally
}
```

storage/dump/CommitLoader.h
```diff
@@ -50,6 +50,16 @@ class CommitLoader {
         if (prevHistory) {
             commit->_data->_history.newCommitHistoryFromPrevious(*prevHistory);
         }
+
+        auto it = std::ranges::find_if(files.value(),
```
rjb32: Can it be const?

storage/versioning/CommitBuilder.cpp
```diff
@@ -41,6 +49,11 @@ GraphReader CommitBuilder::readGraph() const {
     return viewGraph().read();
 }
 
+void CommitBuilder::appendBuilder(std::unique_ptr<DataPartBuilder> builder) {
+    std::scoped_lock lock {_mutex};
```
rjb32: Why a scoped_lock and not a unique_lock?

PR #91, opened 2025-10-31
BUG: Invalid nodes created in WriteStep

cyrusknopf:
- `WriteStep` would create invalid `pendingNodes`/`pendingEdges` in the case of parse errors, which would not be removed

- To reproduce:
```
turing:default> change new
Query executed in 0.07251 ms.
+---+
| 0 |
+---+
turing:default> checkout change-0
turing:default@0> create (n)
[2025-10-31 16:23:26] [error] EXEC_ERROR: Nodes must have at least one label
turing:default@0> create (n:Person)
Query executed in 0.05951 ms.

turing:default@0> change submit
zsh: segmentation fault (core dumped)  turingdb -cli
```

- Added test to expose bug
- Added fix for bug

### DIFF
--- a/pipeline/operations/WriteStep.cpp
+++ b/pipeline/operations/WriteStep.cpp
@@ -169,7 +169,7 @@ void WriteStep::writeEdge(const ContingentNode src, const ContingentNode tgt,
     // Get the EdgeType for PendingEdge
     const TypeConstraint* patternType = edgePattern->getTypeConstraint();
 
-    EdgeTypeID pendingEdgeType {std::numeric_limits<EdgeTypeID>::max()};
+    EdgeTypeID pendingEdgeType;
 
     if (patternType != nullptr) {
         pendingEdgeType = _metadataBuilder->getOrCreateEdgeType(


pipeline/operations/WriteStep.cpp
```diff
@@ -163,16 +166,23 @@ CommitWriteBuffer::ExistingOrPendingNode WriteStep::getOrWriteNode(const EntityP
 
 void WriteStep::writeEdge(const ContingentNode src, const ContingentNode tgt,
                           const EntityPattern* edgePattern) {
-    auto& pendingEdge = _writeBuffer->newPendingEdge(src, tgt);
     // Get the EdgeType for PendingEdge
     const TypeConstraint* patternType = edgePattern->getTypeConstraint();
+
+    EdgeTypeID pendingEdgeType {std::numeric_limits<EdgeTypeID>::max()};
```
rjb32: All IDs are already default initialised

cyrusknopf: Removed explicit initialisation

PR #92, opened 2025-10-31
Add TypeContraint checking into QueryAnalyzer

cyrusknopf:


interpreter/QueryAnalyzer.cpp
```diff
@@ -370,6 +372,51 @@ void QueryAnalyzer::analyzeBinExprConstraint(const BinExpr* binExpr,
     }
 }
 
+void QueryAnalyzer::ensureValidTypeConstraints(EntityPattern* entity) {
+    bioassert(entity);
+
+    const std::string& varName = entity->getVar() ? entity->getVar()->getName() : "";
```
rjb32: Is that string only used in the case of an error?

interpreter/QueryAnalyzer.cpp
```diff
@@ -370,6 +372,51 @@ void QueryAnalyzer::analyzeBinExprConstraint(const BinExpr* binExpr,
...
+    switch (entity->getKind()) {
+        case DeclKind::NODE_DECL: {
+            if (typeConstraints.empty()) [[unlikely]]{
+                std::string error = "Nodes require at least one label.";
+                if (varName != "") {
+                    error += fmt::format(" Node {} was not provided any label constraints.", varName);
+                }
+                throw AnalyzeException(std::move(error));
+            }
+        break;
```
rjb32: Prefer to put the break after the enclosing brackets { } so that the brackets are sandwiched between the case line and the break line

interpreter/QueryAnalyzer.cpp
cyrusknopf: Yep, moved it to only allocate in case of error

interpreter/QueryAnalyzer.cpp
cyrusknopf: Fixed

PR #96, opened 2025-11-04
Merging tombstones

ghost:
Check TombStones When Merging Dataparts.

### DIFF
--- a/storage/mergers/DataPartMerger.cpp
+++ b/storage/mergers/DataPartMerger.cpp
@@ -1,16 +1,21 @@
 #include "DataPartMerger.h"
 
+#include <range/v3/view/enumerate.hpp>
+
 #include "DataPart.h"
 #include "writers/DataPartBuilder.h"
 #include "versioning/CommitBuilder.h"
 #include "EdgeContainer.h"
 #include "NodeContainer.h"
+#include "versioning/TombstoneRanges.h"
 #include "properties/PropertyManager.h"
-#include "ID.h"
 #include "reader/GraphReader.h"
-#include "versioning/VersionController.h"
+
+#include "ID.h"
 
 using namespace db;
+namespace rg = ranges;
+namespace rv = rg::views;
 
 struct PropertyTypeDispatcher {
     ValueType _valueType;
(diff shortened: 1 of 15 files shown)


storage/mergers/DataPartMerger.cpp
```diff
@@ -61,42 +66,83 @@ std::unique_ptr<DataPartBuilder> DataPartMerger::merge(DataPartSpan dataParts) c
 
     const auto& graphMetadata = _graphView.metadata();
 
-    //Count The Number Of Entries For Each Property
+    // Map to store the count the number of entries for each property
     std::unordered_map<PropertyTypeID, size_t> _nodePropertiesCount;
     std::unordered_map<PropertyTypeID, size_t> _edgePropertiesCount;
 
-    for (const auto& dataPart : dataParts) {
+    // Map to store the ranges of the property container to keep in every datapart
```
rjb32: Maps of maps are ugly but I don't know if you have any other possibility to do what you need to do

storage/versioning/TombstoneRanges.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "versioning/Tombstones.h"
+
+using namespace db;
+
+TombstoneRanges::TombstoneRanges(const Tombstones& tombstones)
+    :_tombstones(tombstones)
+{
+}
+
+void TombstoneRanges::populateRanges(const std::vector<EdgeRecord>& baseVec) {
```
rjb32: Use a more explicit name to show what baseVec is for

storage/versioning/TombstoneRanges.h
```diff
@@ -0,0 +1,38 @@
...
+
+    void populateRanges(const std::vector<EdgeRecord>& baseVec);
+
+    template <TypedInternalID IDT>
+    void populateRanges(const std::vector<EntityID>& baseVec);
+
+    const NonDeletedRanges& getRanges() const { return _nonDeletedRanges; };
+    size_t getNumRemainingEntities() const { return _numRemainingEntities; };
+
+    NonDeletedRanges::const_iterator cbegin() const { return _nonDeletedRanges.cbegin(); };
```
rjb32: Define type alias for iterator type to make it look nicer

rjb32:
@LucLabarriere @cyrusknopf Please inspect

storage/mergers/DataPartMerger.cpp
```diff
@@ -61,42 +66,83 @@ std::unique_ptr<DataPartBuilder> DataPartMerger::merge(DataPartSpan dataParts) c
 
     const auto& graphMetadata = _graphView.metadata();
 
-    //Count The Number Of Entries For Each Property
+    // Map to store the count the number of entries for each property
```
LucLabarriere: fix comment

storage/mergers/DataPartMerger.cpp
LucLabarriere: If those are datapart index, can it be a vector of unordered_maps?

storage/mergers/DataPartMerger.cpp
```diff
@@ -61,42 +66,83 @@ std::unique_ptr<DataPartBuilder> DataPartMerger::merge(DataPartSpan dataParts) c
...
 
         for (const auto& [propertyID, propertyContainer] : dataPart->edgeProperties()) {
-            _edgePropertiesCount[propertyID] += propertyContainer->size();
+            const auto [it, inserted] = edgeRangesPerPart[idx].emplace(propertyID, _graphView.tombstones());
+            auto& tombstoneRanges = it->second;
+
+            tombstoneRanges.populateRanges<EdgeID>(propertyContainer->ids());
+            _edgePropertiesCount[propertyID] += tombstoneRanges.getNumRemainingEntities();
         }
     }
```
LucLabarriere: I'm starting to have the feeling that those ranges should be computed on DataPart::load time, when we build the tombstones object. We could do all of those calculations and retrieve them whenever we want. It would simplify considerably those kinds of algorithms

LucLabarriere: Moreover, I think it would considerably improve the tombstones check in the iterators no? @cyrusknopf

storage/mergers/DataPartMerger.cpp
```diff
@@ -61,42 +66,83 @@ std::unique_ptr<DataPartBuilder> DataPartMerger::merge(DataPartSpan dataParts) c
...
+    for (const auto& [idx, dataPart] : dataParts | rv::enumerate) {
+        // Create labelSets pointing to the current commits metatdata.
+        // As the nodeIDs of the original dataPart span are sorted in order, when we insert the
+        // labelSets into the new dataPart they follow an ascending insertion nodeID order -
+        // this follows the pattern expected by dataPart::load() and will allow us to directly copy
+        // other datapart containers without needing to map the nodeIDs.
         for (const auto& nodeRecord : dataPart->nodes().records()) {
-            labelSets.emplace_back(*graphMetadata.labelsets().getValue(nodeRecord._labelset.getID()));
+            if (!_graphView.tombstones().containsNode(nodeIDVal)) {
+                labelSets.emplace_back(*graphMetadata.labelsets().getValue(nodeRecord._labelset.getID()));
```
LucLabarriere: Can you store LabelSetMap&, PropertyTypeMap&, Tombstones&, etc. once so that we can see a bit more clearly what's going on

storage/writers/GraphWriter.cpp
```diff
@@ -44,6 +44,12 @@ bool GraphWriter::submit() {
         return false;
     }
 
+    if (_graph) {
```
cyrusknopf, line 51: Probably should have `if (!_graph) return false;` above this

storage/mergers/DataPartMerger.cpp
cyrusknopf: I don't think its applicable to the iterators because the ranges are based on arbitrary columns which are produced by the previous steps, whereas here they are over the whole graph. If we are to precompute the ranges then it would have to be at Commit build time, and potentially we could store them in the Tombstone object. In any case I'm not sure it's worth it since the only time we need it is during a merge, and we would slow down (to some extent) COMMIT/SUBMIT TIME if we computed them per commit - and the ranges would never be used if there was never a merge

ghost: I agree, it prevents recomputing the same data any times.

storage/mergers/DataPartMerger.cpp
ghost: what do you mean exactly?

LucLabarriere: Just extract the references once and for all at the beginning:

```c++
const Tombostones& tombstones = _graphView.tombstones();
const LabelSetMap& labelsets = _graphMetadata.labelsets();
```

storage/mergers/DataPartMerger.cpp
LucLabarriere: > I don't think its applicable to the iterators because the ranges are based on arbitrary columns which are produced by the previous steps, whereas here they are over the whole graph. If we are to precompute the ranges then it would have to be at Commit build time, and potentially we could store them in the Tombstone object. In any case I'm not sure it's worth it since the only time we need it is during a merge, and we would slow down (to some extent) COMMIT/SUBMIT TIME if we computed them per commit - and the ranges would never be used if there was never a merge

Oh yeah I forgot that it was computed on the columns.

So I guess this is fine

storage/mergers/DataPartMerger.cpp
```diff
@@ -44,59 +49,100 @@ DataPartMerger::DataPartMerger(const CommitData* commitData,
...
+            const auto [it, inserted] = nodeRangesPerPart[idx].emplace(propertyID, _graphView.tombstones());
+            auto& tombstoneRanges = it->second;
+
+            tombstoneRanges.populateRanges<NodeID>(propertyContainer->ids());
+            _nodePropertiesCount[propertyID] += tombstoneRanges.getNumRemainingEntities();
         }
 
         for (const auto& [propertyID, propertyContainer] : dataPart->edgeProperties()) {
-            _edgePropertiesCount[propertyID] += propertyContainer->size();
+            const auto [it, inserted] = edgeRangesPerPart[idx].emplace(propertyID, _graphView.tombstones());
```
rjb32: _graphView.tombstones() can be extracted at the top of the function

storage/mergers/DataPartMerger.cpp
```diff
@@ -44,59 +49,100 @@ DataPartMerger::DataPartMerger(const CommitData* commitData,
...
+    // Loop Over The Dataparts and Start Merging Containers
+    for (const auto& [idx, dataPart] : dataParts | rv::enumerate) {
+        // Create labelSets pointing to the current commits metatdata.
+        // As the nodeIDs of the original dataPart span are sorted in order, when we insert the
+        // labelSets into the new dataPart they follow an ascending insertion nodeID order -
+        // this follows the pattern expected by dataPart::load() and will allow us to directly copy
+        // other datapart containers without needing to map the nodeIDs.
         for (const auto& nodeRecord : dataPart->nodes().records()) {
-            labelSets.emplace_back(*graphMetadata.labelsets().getValue(nodeRecord._labelset.getID()));
+            if (!_graphView.tombstones().containsNode(nodeIDVal)) {
```
rjb32: _graphView.tombstones() to be extracted

PR #98, opened 2025-11-04
Refactor getNode/EdgeCount to a more accurate name/usage

cyrusknopf:
Context:
- `GraphReader::getNode/EdgeCount` did not consider tombstones
- However it was used in cases where instead of the count, what was really desired was the "maximum ID which has not yet been assigned" (e.g. declaring the `firstNodeID` for a new `CommitBuilder`.

Changes:
- Rename `getNode/EdgeCount` to `getTotalNodes/EdgesAllocated`
- Add `getNode/EdgeCount` which gets the total number allocated, minus the number of tombstones
- Use new `getNode/EdgeCount` methods in  `DBServerProcessor::get_graph_status`

### DIFF
--- a/storage/DataPart.h
+++ b/storage/DataPart.h
@@ -50,14 +50,14 @@ public:
      * @detail Does not decrement the count for any nodes which are implicitly deleted by
      * an owning Commit's Tombstones
      */
-    size_t getNodeCount() const;
+    size_t getNodeContainerSize() const;
 
     /**
      * @brief Returns the total number of edges stored in @ref _edges.
      * @detail Does not decrement the count for any edges which are implicitly deleted by
      * an owning Commit's Tombstones
      */
-    size_t getEdgeCount() const;
+    size_t getEdgeContainerSize() const;
 
     bool hasNode(NodeID nodeID) const;
     bool hasEdge(EdgeID edgeID) const;
--- a/storage/reader/GraphReader.cpp
+++ b/storage/reader/GraphReader.cpp
@@ -15,7 +15,7 @@ size_t GraphReader::getTotalNodesAllocated() const {
     }
 
     const auto& lastPart = _view.dataparts().back();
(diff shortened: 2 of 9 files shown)


storage/reader/GraphReader.cpp
```diff
@@ -245,11 +257,11 @@ bool GraphReader::nodeHasProperty(PropertyTypeID ptID, NodeID nodeID) const {
...
 
 bool GraphReader::graphHasNode(NodeID nodeID) const {
-    return (nodeID < getNodeCount());
+    return (nodeID < getTotalNodesAllocated());
 }
 
 bool GraphReader::graphHasEdge(EdgeID edgeID) const {
-    return (edgeID < getEdgeCount());
+    return (edgeID < getTotalEdgesAllocated());
 }
```
LucLabarriere, line 269: What happens if the node is marked by a tombstone?

storage/DataPart.h
```diff
@@ -44,7 +44,19 @@ class DataPart {
...
+     * an owning Commit's Tombstones
+     */
     size_t getNodeCount() const;
+
+    /**
+     * @brief Returns the total number of edges stored in @ref _edges.
+     * @detail Does not decrement the count for any edges which are implicitly deleted by
+     * an owning Commit's Tombstones
+     */
     size_t getEdgeCount() const;
```
LucLabarriere: Should they be renamed getTotalNodesAllocated?

storage/reader/GraphReader.cpp
cyrusknopf: Fixed in #97

storage/DataPart.h
cyrusknopf: Yes you are right these should be renamed

cyrusknopf: Renamed to `getNode/EdgeContainerSize`

PR #101, opened 2025-11-04
v2: Implement PipelineBuilder

rjb32:
* Implement typed interfaces for pipeline processors: wrap the port for the topology of the pipeline DAG and give the columns of interest on each interface to know what is the input or output.
* Implement a PipelineBuilder to encapsulate the boilerplate code necessary to manage and connect pipeline interfaces

TO IMPROVE:
* Interfaces use untyped NamedColumn* for now: pipeline processors need to do a dynamic cast in the prepare phase to get a pointer to the appropriate derived class of Column*. This is a bit of waste, but I did that to have NamedColumn as a wrapper on top of Column to hold the ColumnTag without changing anything in Column which is still used by QueryPlanner v1. -> Should we put the tags directly in Column? (as an optional tag for v1 comptability)
* Some things in PipelineBuilder fill very verbose
* Dataframes (and each NamedColumn) are currently owned by each PipelineBuffer: this complicates the propagation of columns in dataframes along the way because we have double-free if the same NamedColumn pointer is in multiple dataframes

### DIFF
--- a/v2/pipeline/PipelineBuilder.cpp
+++ b/v2/pipeline/PipelineBuilder.cpp
@@ -45,25 +45,7 @@ PipelineBuilder& PipelineBuilder::addScanNodes() {
 
 PipelineBuilder& PipelineBuilder::addGetOutEdges() {
     GetOutEdgesProcessor* getOutEdges = GetOutEdgesProcessor::create(_pipeline);
-
-    switch (_pendingOutput->getKind()) {
-        case PipelineInterfaceKind::NODE:
-        {
-            auto* nodeInput = static_cast<PipelineNodeOutputInterface*>(_pendingOutput);
-            nodeInput->connectTo(getOutEdges->inNodeIDs());
-        }
-        break;
-        case PipelineInterfaceKind::EDGE:
-        {
-            auto* edgeInput = static_cast<PipelineEdgeOutputInterface*>(_pendingOutput);
-            edgeInput->connectTo(getOutEdges->inNodeIDs());
-        }
-        break;
-
-        default:
-            throw PipelineException("PipelineBuilder: cannot add get out edges to non-node or edge output");
-        break;
-    }
+    _pendingOutput->connectTo(getOutEdges->inNodeIDs());
 
     PipelineEdgeOutputInterface& outEdges = getOutEdges->outEdges();
     Dataframe* outDf = outEdges.getDataframe();
@@ -97,30 +79,8 @@ PipelineBuilder& PipelineBuilder::closeMaterialize() {
     return *this;
 }
 
-PipelineBuilder& PipelineBuilder::connectToBlockInput(PipelineBlockInputInterface& input) {
-    switch (_pendingOutput->getKind()) {
-        case PipelineInterfaceKind::NODE:
-            static_cast<PipelineNodeOutputInterface*>(_pendingOutput)->connectTo(input);
-        break;
-        case PipelineInterfaceKind::EDGE:
-            static_cast<PipelineEdgeOutputInterface*>(_pendingOutput)->connectTo(input);
-        break;
-        case PipelineInterfaceKind::BLOCK:
-            static_cast<PipelineBlockOutputInterface*>(_pendingOutput)->connectTo(input);
-        break;
-        case PipelineInterfaceKind::VALUES:
-            static_cast<PipelineValuesOutputInterface*>(_pendingOutput)->connectTo(input);
-        break;
-        case PipelineInterfaceKind::VALUE:
-            static_cast<PipelineValueOutputInterface*>(_pendingOutput)->connectTo(input);
-        break;
-    }
-
-    return *this;
-}
-
 PipelineBuilder& PipelineBuilder::addMaterialize() {
-    connectToBlockInput(_matProc->input());
+    _pendingOutput->connectTo(_matProc->input());
     _pendingOutput = &_matProc->output();
     closeMaterialize();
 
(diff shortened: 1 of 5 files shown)


v2/pipeline/PipelineBuilder.h
```diff
@@ -0,0 +1,78 @@
...
+    PipelineBuilder& addLambdaSource(const LambdaSourceProcessor::Callback& callback);
+
+    PipelineBuilder& addMaterialize();
+
+private:
+    LocalMemory* _mem {nullptr};
+    PipelineV2* _pipeline {nullptr};
+    ColumnTagManager& _tagManager;
+    PipelineOutputInterface* _pendingOutput {nullptr};
+    MaterializeProcessor* _matProc {nullptr};
```
LucLabarriere, line 59: Why do you need to store a pointer to a materialize proc here?

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -0,0 +1,227 @@
...
+        {
+            auto* edgeInput = static_cast<PipelineEdgeOutputInterface*>(_pendingOutput);
+            edgeInput->connectTo(getOutEdges->inNodeIDs());
+        }
+        break;
+
+        default:
+            throw PipelineException("PipelineBuilder: cannot add get out edges to non-node or edge output");
+        break;
+    }
```
LucLabarriere: What about an implementation with virtual function calls:
```c++
virtual void PipelineOutputInterface::connectOutNodeIDs() {
    // throw PipelineException
}
virtual void PipelineOutputInterface::connectOutEdgeIDs() {
    // throw PipelineException
}

void PipelineNodeOutputInterface::connectOutNodeIDs(PipelineNodeInputInterface& input) override {
    input.setNodeIDs(_nodeIDs);
    _port->connectTo(input.getPort();
}

// Usage: Instead of the switch, let the virtual function call do the type dispatch
_pendingOutput->connectOutputNodeIDs(getOutEdges->inNodeIDs());
// This call throws if the output is not a Node interface
```

v2/pipeline/PipelineBuilder.h
rjb32: To have access to MaterializeData which is owned by MaterializeProcessor, to be able to add columns into materialize steps on the go

v2/pipeline/PipelineBuilder.cpp
rjb32: Yes good point thanks

v2/pipeline/processors/GetOutEdgesProcessor.cpp
```diff
@@ -25,37 +27,23 @@ GetOutEdgesProcessor* GetOutEdgesProcessor::create(PipelineV2* pipeline) {
...
-    ColumnEdgeTypes* edgeTypes = dynamic_cast<ColumnEdgeTypes*>(_outEdgeTypes.getRawColumn());
-
-    if (!nodeIDs || !indices || !edgeIDs || !targetNodes || !edgeTypes) {
-        throw PipelineException("GetOutEdgesProcessor: Invalid columns");
-    }
+    ColumnNodeIDs* nodeIDs = dynamic_cast<ColumnNodeIDs*>(_inNodeIDs.getNodeIDs()->getColumn());
+    ColumnVector<size_t>* indices = dynamic_cast<ColumnVector<size_t>*>(_outEdges.getIndices()->getColumn());
+    ColumnEdgeIDs* edgeIDs = dynamic_cast<ColumnEdgeIDs*>(_outEdges.getEdgeIDs()->getColumn());
+    ColumnNodeIDs* targetNodes = dynamic_cast<ColumnNodeIDs*>(_outEdges.getTargetNodes()->getColumn());
+    ColumnEdgeTypes* edgeTypes = dynamic_cast<ColumnEdgeTypes*>(_outEdges.getEdgeTypes()->getColumn());
```
LucLabarriere, line 46: I think this is really nice actually, we don't even have to check for nullptr. This is really concise + safe

v2/pipeline/PipelineBuilder.h
```diff
@@ -0,0 +1,78 @@
...
+    {
+    }
+
+    PipelineOutputInterface* getOutput() const { return _pendingOutput; }
+
+    template <typename ColumnType>
+    PipelineBuilder& addColumnToOutput(ColumnTag tag) {
+        allocColumn<ColumnType>(_pendingOutput->getDataframe(), tag);
+        return *this;
+    }
```
LucLabarriere, line 37: I think this is what's trickiest to me, the allocation of columns, named columns, dataframes and blocks. I think other allocations are fine actually (ports etc are not easy but it's clearer that Pipeline is the owner)

rjb32: Columns are owned by LocalMemory, pipeline objects are owned by the Pipeline, Dataframe is owned by each PipelineBuffer, NamedColumn is owned by Dataframe

v2/pipeline/PipelineBuilder.cpp
rjb32: Fixed.

v2/pipeline/processors/GetOutEdgesProcessor.cpp
rjb32: The only thing is that we have to go through an untyped NamedColumn and do a dynamic_cast in the interface, when we kind of already now if it is supposed to be a ColumnNodeIDs..etc

v2/pipeline/PipelineBuilder.h
```diff
@@ -0,0 +1,76 @@
...
+
+    PipelineBuilder& addScanNodes();
```
LucLabarriere, line 39: What do you think of, instead of returning PipelineBuilder&, returning the output interface instead to be able to interact with the columns:

```c++
PipelineNodeOutputInterface& outNodeInterface = builder.addScanNodes();
outNodeIDs.getNodeIDs()->getHeader().setName("n");
```

I don't think the chain builder pattern would be that usefull here anyway (in the unit test you don't really do builder.addScanNodes().addGetOutEdges()...)

In the pipeline generator I think it would be pretty convenient

rjb32: Yes indeed, done here https://github.com/turing-db/turingdb/pull/104

PR #102, opened 2025-11-05
Remove common as a submodule

rjb32:
Remove common as a submodule because common is not used anywhere else anymore.
Put common/lib into common/ and common/external into external/


rjb32:
It's actually used in the C++ SDK which is itself used in the turing-bench tool. So I will close it for now.

PR #104, opened 2025-11-05
PipelineBuilder: remove builder pattern

rjb32:
Remove builder pattern in PipelineBuilder and return directly the type of the last pipeline output interface

### DIFF
--- a/v2/pipeline/PipelineBuilder.cpp
+++ b/v2/pipeline/PipelineBuilder.cpp
@@ -27,7 +27,7 @@ void duplicateDataframeShape(LocalMemory* mem, Dataframe* src, Dataframe* dest) 
 
 }
 
-PipelineOutputInterface& PipelineBuilder::addScanNodes() {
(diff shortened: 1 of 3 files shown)


v2/pipeline/PipelineBuilder.cpp
```diff
@@ -40,10 +40,10 @@ PipelineBuilder& PipelineBuilder::addScanNodes() {
 
     _pendingOutput = &outNodeIDs;
 
-    return *this;
+    return outNodeIDs;
 }
 
-PipelineBuilder& PipelineBuilder::addGetOutEdges() {
+PipelineOutputInterface& PipelineBuilder::addGetOutEdges() {
```
LucLabarriere: Why not return the correct type directly PipelineEdgeOutputInterface? Can allow to avoid virtual function calls after in the pipeline generator

rjb32: Yes indeed good one, done

PR #105, opened 2025-11-05
Add CartesianProductProcessor to v2

cyrusknopf:
Adds the CartesianProductProcessor to compute the Cartesian Product of two input dataframes.

High level intuition for the algorithm [on Notion](https://www.notion.so/turingbio/CartesianProductProcessor-2a43aad664c88045b559d5755f72a553?source=copy_link#2a43aad664c8806ab5c1f142d464ecbb).

### DIFF
--- a/test/v2/pipeline/processors/CartesianProductProcessorTest.cpp
+++ b/test/v2/pipeline/processors/CartesianProductProcessorTest.cpp
@@ -1,12 +1,12 @@
 #include <gtest/gtest.h>
 
 #include <range/v3/view/zip.hpp>
-#include <set>
 
 #include "EdgeRecord.h"
 #include "ID.h"
 #include "LineContainer.h"
 #include "dataframe/ColumnTag.h"
+#include "iterators/ChunkConfig.h"
 #include "processors/ProcessorTester.h"
 
 #include "processors/CartesianProductProcessor.h"
@@ -1137,6 +1137,89 @@ TEST_F(CartesianProductProcessorTest, bothOneRow) {
     for (const auto& [exp, act] : rv::zip(expectedRow, actualRow)) {
         EXPECT_EQ(exp.getValue(),act.getValue());
     }
+}
+
+TEST_F(CartesianProductProcessorTest, tenThousandRowDataframes) {
+    auto [transaction, view, reader] = readGraph();
+
+    constexpr size_t CHUNK_SIZE = ChunkConfig::CHUNK_SIZE;
+
+    // Anything greater than this becomes too expensive
+    constexpr size_t LHS_ROWS = 10'000;
+    constexpr size_t LHS_COLS = 1;
+    constexpr size_t RHS_ROWS = LHS_ROWS;
+    constexpr size_t RHS_COLS = LHS_COLS;
+
+    size_t EXPECTED_NUM_CHUNKS = std::ceil((LHS_ROWS * RHS_ROWS) / (float)CHUNK_SIZE);
+
+    const auto genDF = [&](Dataframe* df, bool& isFinished, auto operation) -> void {
+        if (operation != LambdaSourceProcessor::Operation::EXECUTE) {
+            return;
+        }
+        ASSERT_EQ(df->size(), LHS_COLS);
+        auto* col = df->cols().front()->as<ColumnNodeIDs>();
+        col->resize(LHS_ROWS);
+        std::iota(col->begin(), col->end(), 0);
+
+        isFinished = true;
+    };
+
+    { // Wire up the cartesian product to the two inputs
+        auto& rhsIF = _builder->addLambdaSource(genDF);
+        for (size_t i = 0; i < RHS_COLS; i++) {
+            _builder->addColumnToOutput<ColumnNodeIDs>(
+                _pipeline.getDataframeManager()->allocTag());
+        }
+
+        [[maybe_unused]] auto& lhsIF = _builder->addLambdaSource(genDF);
+        for (size_t i = 0; i < LHS_COLS; i++) {
+            _builder->addColumnToOutput<ColumnNodeIDs>(
+                _pipeline.getDataframeManager()->allocTag());
+        }
+
+        const auto& cartProd = _builder->addCartesianProduct(&rhsIF);
+        ASSERT_EQ(cartProd.getDataframe()->cols().size(), LHS_COLS + RHS_COLS);
+    }
+
+    // It is too expensive to compare for equality
+    /*
+    Rows expectedRows;
+    {
+        for (size_t i = 0; i < LHS_ROWS; i++) {
+            for (size_t j = 0; j < RHS_ROWS; j++) {
+                expectedRows.emplace(i, j);
+            }
+        }
+    }
+    */
+
+    bool executed {false};
+    size_t numChunks {0};
+    size_t numRows {0};
+    const auto VERIFY_CALLBACK = [&](const Dataframe* df,
+                                     LambdaProcessor::Operation operation) -> void {
+        if (operation == LambdaProcessor::Operation::RESET) {
+            return;
+        }
+        ASSERT_EQ(df->size(), LHS_COLS + RHS_COLS);
+
+        executed = true;
+        numChunks++;
+        numRows += df->getRowCount();
+
+        /*
+        for (size_t rowPtr = 0; rowPtr < df->getRowCount(); rowPtr++) {
+            auto* lhs = df->cols().front()->as<ColumnNodeIDs>();
+            auto* rhs = df->cols().back()->as<ColumnNodeIDs>();
+            EXPECT_TRUE(expectedRows.contains({lhs->at(rowPtr), rhs->at(rowPtr)}));
+        }
+        */
+    };
+    _builder->addLambda(VERIFY_CALLBACK);
+    EXECUTE(view, CHUNK_SIZE);
+    ASSERT_TRUE(executed);
+    EXPECT_EQ(EXPECTED_NUM_CHUNKS, numChunks);
+    EXPECT_EQ(LHS_ROWS * RHS_ROWS, numRows);
 }
 
 int main(int argc, char** argv) {


storage/dataframe/Dataframe.cpp
```diff
@@ -26,10 +32,47 @@ size_t Dataframe::getRowCount() const {
 }
 
 void Dataframe::dump(std::ostream& out) const {
+    if (_cols.empty()) {
+        return;
+    }
+
+    std::vector<std::ostringstream> colStreams(_cols.size());
```
cyrusknopf, line 35: this now dumps columns next to each other so its easier to read the dataframe

storage/dataframe/Dataframe.cpp
```diff
@@ -43,6 +86,36 @@ void Dataframe::copyFromLine(const Dataframe* other, size_t startRow, size_t row
...
+
+            dstColumnVector->resize(newRows);
+            auto& srcRaw = srcColumnVector->getRaw();
+            auto& dstRaw = dstColumnVector->getRaw();
+
+            std::copy(begin(srcRaw),end(srcRaw),begin(dstRaw) + oldRows);
+        });
+    }
+}
+
```
LucLabarriere, line 123: Is this appending the lines of one dataframe to the other? i.e. concatenate columns by pair? If so I'm not a fan of the bioasserts, this is highly likely to produce undefined behaviour in production (e.g. if someone expects that to just append the columns at the end of the dataframe). I think exceptions make more sense here.

Also:
```
auto* srcColumnVector = static_cast<decltype(dstColumnVector)>(src);
bioassert(srcColumnVector);
```

here I would dynamic_cast and if (...) throw

Performance cost is completely negligible (< 1μs whereas the actual copy is quite expensive)

storage/dataframe/Dataframe.h
```diff
@@ -51,7 +52,9 @@ class Dataframe {
     void copyFromLine(const Dataframe* other, size_t startRow, size_t rowCount);
     void copyFrom(const Dataframe* other);
 
-    void dump(std::ostream& out) const;
+    void append(const Dataframe* other);
+
+    void dump(std::ostream& out = std::cout) const;
```
LucLabarriere: Do you need the default argument? It comes at the cost of the iostream header

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -21,8 +23,22 @@ namespace {
...
+        Column* newCol = mem->allocSame(col->getColumn());
+        auto* newNamedCol = NamedColumn::create(dfMan, newCol, col->getTag());
+        dest->addColumn(newNamedCol);
+    }
+}
+
+/*
+* @brief Column-wise concatenation of @param src onto @param dest
+*/
+void concatDataframeShape(LocalMemory* mem,
```
LucLabarriere, line 38: Wait I don't see the difference between the two functions

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -81,6 +97,32 @@ PipelineEdgeOutputInterface& PipelineBuilder::addGetOutEdges() {
...
+
+    // The output DF should be the same as the inputs, concatenated
+    duplicateDataframeShape(_mem, _dfMan, leftDf, outDf);
+    concatDataframeShape(_mem, _dfMan, rightDf, outDf);
+
+    _pendingOutput.setInterface(&output);
+
+    // Initialise the processor's "memory" stores for left and right ports
+    duplicateDataframeShape(_mem, _dfMan, leftDf, &cartProd->leftMemory());
+    duplicateDataframeShape(_mem, _dfMan, rightDf, &cartProd->rightMemory());
```
LucLabarriere, line 121: I assume this parts:

```c++
    // The output DF should be the same as the inputs, concatenated
    duplicateDataframeShape(_mem, _dfMan, leftDf, outDf);
    concatDataframeShape(_mem, _dfMan, rightDf, outDf);
```
Allocates the cols for the output, so if lhs and rhs had 3 and 5 cols, respectively, then the output will have 8 cols.

But what does this do?

```c++
    // Initialise the processor's "memory" stores for left and right ports
    duplicateDataframeShape(_mem, _dfMan, leftDf, &cartProd->leftMemory());
    duplicateDataframeShape(_mem, _dfMan, rightDf, &cartProd->rightMemory());
```

Is it because you need to "duplicate" whatever you receive as inputs because it will get overwritten at each executeCycle and you need to keep the full information? In which case I would have thought you only needed to duplicate the output of one branch (you hold the first chunk of lhs, and process all chunks of rhs, then hold the next chunk of lhs and process all chunks of rhs, etc. The rhs information is stored in the processor so that we don't need to replay, but I'm not sure I get why you need to hold onto the lhs too

LucLabarriere:
I don't think I will be able to review the CartesianProductProcessor.cpp file, it's just so big it will take me a full day I think, if we have good unit test coverage maybe it's good enough and I can have a look at it after Connected Data?

storage/dataframe/Dataframe.cpp
cyrusknopf: For dataframes
```
a =
+--+--+--+
| a| b| c|
| e| f| g|
+--+--+--+
```
and 
```
b = 
+--+--+--+
| 1| 2| 3|
+--+--+--+
| 5| 6| 7|
+--+--+--+
```
then `a.append(b)` is
```
+--+--+--+
| a|b |c |
+--+--+--+
|e |f |g |
+--+--+--+
| 1| 2| 3|
+--+--+--+
| 5| 6| 7|
+--+--+--+
```

Are you not describing an issue with misuse by the developer instead of a problem with the function? Thus it is not a user error. 

You are right I messed up with the static cast vs dynamic cast, I'll fix that.

v2/pipeline/PipelineBuilder.cpp
cyrusknopf: They are exactly the same, but the "duplicate dataframe shape" function just coincidentally performs concatenation (because it does not "erase" the existing shape of the dataframe, just adds to it). So using
```
duplicateShape(...)
duplicateShape(...)
```
is misleading because the second call is actually performing concatenation, so I made an explicit function.

v2/pipeline/PipelineBuilder.cpp
cyrusknopf: We decided against holding onto a single port (say, RHS) because it requires complete exhaustion of RHS before the LHS can be consumed. If you do not memorise the LHS then you cannot ever consume it until you have produced all tuples from it, meaning you need to wait for the entire RHS stream to finish. If we memorise both side we unblock both left and right input ports.

LucLabarriere: Ok thanks

cyrusknopf:
> I don't think I will be able to review the CartesianProductProcessor.cpp file, it's just so big it will take me a full day I think, if we have good unit test coverage maybe it's good enough and I can have a look at it after Connected Data?

Sure, I've tried to cover as many edge cases as I can in the unit tests, namely things like:
- Both inputs being a single chunk -> producing a single chunk
- Both inputs being a single chunk -> producing multiple chunks
- Both inputs being multiple chunks -> producing multiple chunks
- One input being multiple chunks, the other being single -> producing multiple chunks (both LHS and RHS symmetric cases)
and variants with varying number of columns in all of the above.

Let me know if there are any other cases you can think of and I can add tests as well before Connected Data.

storage/dataframe/Dataframe.h
cyrusknopf: Not necessary, but out of interest is there a tangible compilation time penalty for including such a header?

LucLabarriere: I'm not sure how big iostream is, but Dataframe is included everywhere so the cost adds up quickly

storage/dataframe/Dataframe.cpp
```diff
@@ -43,6 +86,36 @@ void Dataframe::copyFromLine(const Dataframe* other, size_t startRow, size_t row
...
+    for (size_t i = 0; i < _cols.size(); i++) {
+        NamedColumn* dstNamed = _cols[i];
+        NamedColumn* srcNamed = otherCols[i];
+
+        Column* dst = dstNamed->getColumn();
+        Column* src = srcNamed->getColumn();
+
+        // Resize destination to fit new data
+        dispatchColumnVector(dst, [&](auto* dstColumnVector) {
+            auto* srcColumnVector = static_cast<decltype(dstColumnVector)>(src);
```
ghost: This can probably be const

storage/dataframe/Dataframe.cpp
```diff
@@ -43,6 +86,36 @@ void Dataframe::copyFromLine(const Dataframe* other, size_t startRow, size_t row
...
+            bioassert(srcColumnVector);
+
+            dstColumnVector->resize(newRows);
+            auto& srcRaw = srcColumnVector->getRaw();
```
ghost: same here

storage/dataframe/Dataframe.h
rjb32: Don't put std::cout as default argument

storage/dataframe/Dataframe.cpp
rjb32: Let's prefer indeed to throw exceptions instead of assertions. I do it in my code and we will progressively retire the bioassert system or replace by an implementation that throws exceptions

rjb32:
Ok it looks good to me at a surface level and we talked about it extensively. Let's add query-level tests in another PR.

rjb32:
@cyrusknopf Do you have a unit test for 2 chunks of different size? Or 2 chunks of 1 line?

v2/pipeline/processors/CartesianProductProcessor.h
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    PipelineBlockInputInterface& leftHandSide() { return _lhs; }
+    PipelineBlockInputInterface& rightHandSide() { return _rhs; }
+    PipelineBlockOutputInterface& output() { return _out; }
+
+    Dataframe& leftMemory() { return _leftMemory; }
+    Dataframe& rightMemory() { return _rightMemory; }
+
+    CartesianProductProcessor(CartesianProductProcessor&) = delete;
```
rjb32: The constructor is private, this is implicitly deleted

cyrusknopf:
> @cyrusknopf Do you have a unit test for 2 chunks of different size? Or 2 chunks of 1 line?

Yep I have tests for inputs of different sizes, but not both being 1 line, I'll add a test for that.

storage/dataframe/Dataframe.h
cyrusknopf: Removed std::cout and <iostream> import

storage/dataframe/Dataframe.cpp
cyrusknopf: I've removed the asserts in favour of exceptions

storage/dataframe/Dataframe.cpp
cyrusknopf: Nice catch fixed thanks

storage/dataframe/Dataframe.cpp
cyrusknopf: also fixed

PR #106, opened 2025-11-06
Turing-846: CALL ... YIELD ... support in planner v2

LucLabarriere:
# CALL database procedures
Implementation of CALL ... YIELD ... AS ... in parser, analyzer, planner

> [!WARNING]
Does not implement subqueries yet `CALL { MATCH ... RETURN ... }`, only procedures

@rjb32 @cyrusknopf @suhassn-turing-bio If you have ideas on how to simplify the implementation, feel free!

## Implementation
- Added a `DatabaseProcedureProcessor`
- It has a `Procedure` member which is the object passed as argument to the procedure callback, it allows to control the state of the processor
- The `Procedure` class has a `ProcedureData` member which is the memory of the processor. Similarly to how `ScanNodesProcessor` needs a `ScanNodesChunkWriter _it` member, an arbitrary procedure can inherit from `ProcessorData` to have its own memory to be used in the Procedure callback. `ProcedureData` also stores pointers for the columns it needs to fill. For example, the `db.history` procedure retrieves those columns:

```c++
Column* rawCommitCol = data.getReturnColumn(0);
Column* rawNodeCountCol = data.getReturnColumn(1);
Column* rawEdgeCountCol = data.getReturnColumn(2);
Column* rawPartCountCol = data.getReturnColumn(3);
```

- The `Procedure` class also has a pointer to a static object `ProcedureBlueprint` which represents the procedure
- `ProcedureBlueprint` stores the name of the procedure, the execution callback, the allocation callback (to allocate the `ProcedureData`), the return value information (return type + column name).
- `ProcedureBlueprintMap` is a static object that stores all the procedures we supports:

```c++
const std::array<ProcedureBlueprint, 4> ProcedureBlueprintMap::_blueprints {
    LabelsProcedure::createBlueprint(),
    PropertyTypesProcedure::createBlueprint(),
    EdgeTypesProcedure::createBlueprint(),
    HistoryProcedure::createBlueprint()
};
```

The procedures are automatically registered in the `FunctionDecls` maps that is used during the query analysing/planning phases

## Examples (reactome, modified with 1 commit)
- `CALL db.history() YIELD commit AS hash, nodeCount, edgeCount, partCount`
```
turing:reactome> #v2 CALL db.history() YIELD commit AS hash, nodeCount, edgeCount, partCount
+------------------+-----------+-----------+-----------+
| hash             | nodeCount | edgeCount | partCount |
+------------------+-----------+-----------+-----------+
| 39f384ee3cd36ddf | 0         | 0         | 0         |
+------------------+-----------+-----------+-----------+
| 3baf89c373b254e  | 2588827   | 10042846  | 18        |
+------------------+-----------+-----------+-----------+
| 789a8908a04fb4f2 | 2588827   | 10042846  | 1         |
+------------------+-----------+-----------+-----------+
```
- `CALL db.propertyTypes() YIELD propertyType AS pt, valueType AS vt RETURN pt, vt LIMIT 8`
```
turing:reactome> #v2 CALL db.propertyTypes() YIELD propertyType AS pt, valueType AS vt RETURN pt, vt LIMIT 8
+----------------------+--------+
| pt                   | vt     |
+----------------------+--------+
| dbId (Int64)         | Int64  |
+----------------------+--------+
| dbId (UInt64)        | UInt64 |
+----------------------+--------+
| schemaClass (String) | String |
+----------------------+--------+
| displayName (String) | String |
+----------------------+--------+
| label (String)       | String |
+----------------------+--------+
| coordinate (Int64)   | Int64  |
+----------------------+--------+
| coordinate (UInt64)  | UInt64 |
+----------------------+--------+
| stId (String)        | String |
+----------------------+--------+
```
- `CALL db.edgeTypes() LIMIT 5`
```
turing:reactome> #v2 CALL db.edgeTypes() YIELD edgeType RETURN edgeType LIMIT 10
+-------------------+
| edgeType          |
+-------------------+
| author            |
+-------------------+
| modified          |
+-------------------+
| affiliation       |
+-------------------+
| referenceDatabase |
+-------------------+
| created           |
+-------------------+
| crossReference    |
+-------------------+
| superTaxon        |
+-------------------+
| species           |
+-------------------+
| referenceGene     |
+-------------------+
| referenceEntity   |
+-------------------+
```
- `CALL db.labels() YIELD label MATCH (n) RETURN label, n LIMIT 5`
```
turing:reactome> #v2 CALL db.labels() YIELD label MATCH (n) RETURN label, n LIMIT 5
+----------------+---+
| label          | n |
+----------------+---+
| DatabaseObject | 0 |
+----------------+---+
| DatabaseObject | 1 |
+----------------+---+
| DatabaseObject | 2 |
+----------------+---+
| DatabaseObject | 3 |
+----------------+---+
| DatabaseObject | 4 |
+----------------+---+
```

### DIFF
--- a/v2/AST/FunctionDecls.cpp
+++ b/v2/AST/FunctionDecls.cpp
@@ -26,6 +26,7 @@ std::unique_ptr<FunctionDecls> FunctionDecls::createDefault(const ProcedureBluep
             switch (returnItem._type) {
                 case ProcedureReturnType::INVALID:
                     throw FatalException("Invalid procedure return type");
+                    break;
 
                 case ProcedureReturnType::NODE:
                     declBuilder.addReturnType(EvaluatedType::NodePattern, returnItem._name);
--- a/v2/pipeline/processors/DatabaseProcedureProcessor.cpp
+++ b/v2/pipeline/processors/DatabaseProcedureProcessor.cpp
@@ -90,7 +90,8 @@ void DatabaseProcedureProcessor::allocColumns(LocalMemory& mem,
 
         switch (colType) {
             case ProcedureReturnType::INVALID: {
-                continue;
+                throw PipelineException("Invalid procedure return type");
+                break;
             }
 
             case ProcedureReturnType::NODE: {
--- a/v2/plan/PipelineGenerator.cpp
+++ b/v2/plan/PipelineGenerator.cpp
@@ -167,39 +167,51 @@ PipelineOutputInterface* PipelineGenerator::translateNode(PlanGraphNode* node) {
     switch (node->getOpcode()) {
         case PlanGraphOpcode::VAR:
             return translateVarNode(static_cast<VarNode*>(node));
+        break;
 
         case PlanGraphOpcode::SCAN_NODES:
             return translateScanNodesNode(static_cast<ScanNodesNode*>(node));
+        break;
 
         case PlanGraphOpcode::GET_OUT_EDGES:
             return translateGetOutEdgesNode(static_cast<GetOutEdgesNode*>(node));
+        break;
 
         case PlanGraphOpcode::PRODUCE_RESULTS:
             return translateProduceResultsNode(static_cast<ProduceResultsNode*>(node));
+        break;
 
         case PlanGraphOpcode::FILTER_NODE:
             return translateNodeFilterNode(static_cast<NodeFilterNode*>(node));
+        break;
 
         case PlanGraphOpcode::FILTER_EDGE:
             return translateEdgeFilterNode(static_cast<EdgeFilterNode*>(node));
+        break;
 
         case PlanGraphOpcode::SKIP:
             return translateSkipNode(static_cast<SkipNode*>(node));
+        break;
 
         case PlanGraphOpcode::LIMIT:
             return translateLimitNode(static_cast<LimitNode*>(node));
+        break;
 
         case PlanGraphOpcode::GET_EDGE_TARGET:
             return translateGetEdgeTargetNode(static_cast<GetEdgeTargetNode*>(node));
+        break;
 
         case PlanGraphOpcode::GET_IN_EDGES:
             return translateGetInEdgesNode(static_cast<GetInEdgesNode*>(node));
+        break;
 
         case PlanGraphOpcode::GET_EDGES:
             return translateGetEdgesNode(static_cast<GetEdgesNode*>(node));
+        break;
 
         case PlanGraphOpcode::CARTESIAN_PRODUCT:
             return translateCartesianProductNode(static_cast<CartesianProductNode*>(node));
+        break;
 
         case PlanGraphOpcode::GET_PROPERTY:
             return translateGetPropertyNode(static_cast<GetPropertyNode*>(node));
@@ -207,6 +219,7 @@ PipelineOutputInterface* PipelineGenerator::translateNode(PlanGraphNode* node) {
 
         case PlanGraphOpcode::GET_PROPERTY_WITH_NULL:
             return translateGetPropertyWithNullNode(static_cast<GetPropertyWithNullNode*>(node));
+        break;
 
         case PlanGraphOpcode::AGGREGATE_EVAL:
             return translateAggregateEvalNode(static_cast<AggregateEvalNode*>(node));
@@ -214,6 +227,7 @@ PipelineOutputInterface* PipelineGenerator::translateNode(PlanGraphNode* node) {
 
         case PlanGraphOpcode::PROCEDURE_EVAL:
             return translateProcedureEvalNode(static_cast<ProcedureEvalNode*>(node));
+        break;
 
         case PlanGraphOpcode::GET_ENTITY_TYPE:
         case PlanGraphOpcode::JOIN:
@@ -226,6 +240,7 @@ PipelineOutputInterface* PipelineGenerator::translateNode(PlanGraphNode* node) {
         case PlanGraphOpcode::_SIZE:
             throw PlannerException(fmt::format("PipelineGenerator does not support PlanGraphNode: {}",
                                                PlanGraphOpcodeDescription::value(node->getOpcode())));
+        break;
     }
 
     throw FatalException("Failed to match against PlanGraphOpcode");
@@ -654,7 +669,6 @@ PipelineOutputInterface* PipelineGenerator::translateProcedureEvalNode(Procedure
         for (const auto* item : *yield->getItems()) {
             const Symbol* symbol = item->getSymbol();
 
-            // TODO: handle the AS keyword there
             yieldItems.emplace_back(symbol->getOriginalName(), symbol->getName());
             yieldDecls.push_back(item->getExprVarDecl());
         }


v2/pipeline/procedures/EdgeTypesProcedure.h
```diff
@@ -0,0 +1,36 @@
...
+        try {
+            return {
+                ._name = "db.edgeTypes",
+                ._execCallback = &execute,
+                ._allocCallback = &allocData,
+                ._returnValues = {{"id", ProcedureReturnType::EDGE_TYPE_ID},
+                                  {"edgeType", ProcedureReturnType::STRING_VIEW}},
+                ._valid = true,
+            };
+        } catch (const std::exception& e) {
```
rjb32: In which case this constructor can throw?

v2/pipeline/procedures/ProcedureBlueprint.h
```diff
@@ -0,0 +1,61 @@
...
+    struct YieldItem {
+        std::string_view _baseName;
+        std::string_view _asName;
+        NamedColumn* _col {nullptr};
+    };
+
+    using ExecuteCallback = void (*)(Procedure&);
+    using AllocCallback = std::unique_ptr<ProcedureData> (*)();
+
+    size_t getReturnValueIndex(std::string_view name) const {
```
rjb32: Put in cpp file

v2/pipeline/procedures/ProcedureBlueprint.h
```diff
@@ -0,0 +1,61 @@
...
+        return _returnValues[index]._type;
+    }
+
+    std::string_view _name;
+    ExecuteCallback _execCallback = nullptr;
+    AllocCallback _allocCallback = nullptr;
+    ProcedureReturnValues _returnValues;
+    bool _valid = false;
+
+    void returnAll(std::vector<YieldItem>& yieldItems) const {
```
rjb32: Put in cpp file

v2/pipeline/procedures/ProcedureBlueprint.h
```diff
@@ -0,0 +1,61 @@
...
+
+        throw PipelineException("Column is not returned by the procedure");
+    }
+
+    ProcedureReturnType getReturnValueType(size_t index) const {
+        return _returnValues[index]._type;
+    }
+
+    std::string_view _name;
+    ExecuteCallback _execCallback = nullptr;
```
rjb32: Use {} initialization style

rjb32:
I think it's a very good work! 

Just a thing for the future: we will want to dynamically add procedure blueprints at runtime in the DB one day, when we support extensions that can loaded from a dynamic library. And it's a good justification as well to not have the ProcedureBlueprintMap as a static variable, and to think of where it should be owned. To me it should be a pointer given to you through some of the Context structures.

v2/pipeline/procedures/EdgeTypesProcedure.h
LucLabarriere: Only in the case we couldn't allocate, I'm just unsure what happens if it throws since the Blueprint map is static. Maybe this is unnecessary and I can ignore this

LucLabarriere:
> I think it's a very good work!
> 
> Just a thing for the future: we will want to dynamically add procedure blueprints at runtime in the DB one day, when we support extensions that can loaded from a dynamic library. And it's a good justification as well to not have the ProcedureBlueprintMap as a static variable, and to think of where it should be owned. To me it should be a pointer given to you through some of the Context structures.

Makes sense! I'll change that immediately, will be cleaner anyway

LucLabarriere:
> > I think it's a very good work!
> > Just a thing for the future: we will want to dynamically add procedure blueprints at runtime in the DB one day, when we support extensions that can loaded from a dynamic library. And it's a good justification as well to not have the ProcedureBlueprintMap as a static variable, and to think of where it should be owned. To me it should be a pointer given to you through some of the Context structures.
> 
> Makes sense! I'll change that immediately, will be cleaner anyway

@rjb32 actually I just did that and it becomes quite complicated. I think it's fine but we will soon need to clean up a bit our context objects, and clarify ownership

rjb32:
I think it's looking good that it is in InterpreterContext. The context passing actually seems to look good.

rjb32:
To be seen later: how do we manage "libraries" of procedures in a more systematic way.

LucLabarriere:
> To be seen later: how do we manage "libraries" of procedures in a more systematic way.

Yes ok! functions could also be handled the same way, right now they are "owned" by FunctionDecls which is owned by CypherAST. This will have to be stored elsewhere, probably alongside procedures

storage/columns/ColumnDispatcher.h
```diff
@@ -165,8 +180,11 @@ struct ColumnTypeFromKind<ColumnVector<const Change*>::staticKind()> {
         COL_CASE(ColumnVector<NodeID>)                                                   \
         COL_CASE(ColumnVector<EdgeID>)                                                   \
         COL_CASE(ColumnVector<EntityID>)                                                 \
+        COL_CASE(ColumnVector<LabelID>)                                                  \
+        COL_CASE(ColumnVector<EdgeTypeID>)                                               \
```
cyrusknopf, line 184: Out of interest, did you get a compile error before adding these? Or was it a runtime error? I'm wondering how maintainable this approach is (I suspect not very).

query/AST/QueryCommand.h
```diff
@@ -231,11 +231,11 @@ class CallCommand : public QueryCommand {
...
 
     constexpr std::span<const std::string_view> getColNames() const {
-        return colNameTable[static_cast<size_t>(Type::SIZE)];
+        return colNameTable[static_cast<size_t>(_type)];
     }
 
     constexpr std::span<const ValueType> getColTypes() const {
-        return colTypeTable[static_cast<size_t>(Type::SIZE)];
+        return colTypeTable[static_cast<size_t>(_type)];
     }
```
LucLabarriere, line 239: @suhassn-turing-bio I just had a bug with this in the regression tests, do you know why that was working at all??

v2/AST/expr/Expr.h
```diff
@@ -37,10 +37,18 @@ class Expr {
 
     const VarDecl* getExprVarDecl() const { return _exprVarDecl; }
 
+    [[nodiscard]] const std::string_view& getName() const {
```
cyrusknopf: Why string_view reference?

v2/AST/FunctionDecls.cpp
```diff
@@ -9,90 +14,122 @@ FunctionDecls::FunctionDecls()
...
-    decls->create("db.edgeTypes")
-        .setReturnType(EvaluatedType::String);
+    for (const auto& blueprint : procedures.getAll()) {
+        auto declBuilder = decls->create(blueprint._name);
+        declBuilder.setIsDatabaseProcedure(true);
+
+        for (const auto& returnItem : blueprint._returnValues) {
+            switch (returnItem._type) {
+                case ProcedureReturnType::INVALID:
+                    throw FatalException("Invalid procedure return type");
```
cyrusknopf, line 28: Stylistic but should  we still have break here?

v2/pipeline/procedures/EdgeTypesProcedure.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+            if (!data._it->isValid()) {
+                proc.finish();
+            }
+
+            return;
+        }
+    }
+
+    throw PipelineException("Unknown procedure step");
```
cyrusknopf, line 52: Is there a reason for using early returns instead of breaks? And having a fallthrough exception instead of a default case?

v2/pipeline/processors/DatabaseProcedureProcessor.cpp
```diff
@@ -0,0 +1,175 @@
...
+        const std::string_view& asName = item._asName;
+
+        const size_t colIndex = _procedure._blueprint->getReturnValueIndex(colName);
+        const ProcedureReturnType colType = _procedure._blueprint->getReturnValueType(colIndex);
+
+        Column* col = nullptr;
+
+        switch (colType) {
+            case ProcedureReturnType::INVALID: {
+                continue;
```
cyrusknopf: Should this be an error?

v2/plan/PipelineGenerator.cpp
```diff
@@ -160,64 +167,54 @@ PipelineOutputInterface* PipelineGenerator::translateNode(PlanGraphNode* node) {
     switch (node->getOpcode()) {
         case PlanGraphOpcode::VAR:
             return translateVarNode(static_cast<VarNode*>(node));
-        break;
```
cyrusknopf, line 163: Is this just stylistic? Or is there a reason to remove them

v2/plan/PipelineGenerator.cpp
```diff
@@ -606,3 +615,63 @@ PipelineOutputInterface* PipelineGenerator::translateAggregateEvalNode(Aggregate
...
+        throw PlannerException(fmt::format("Procedure '{}' does not exist", signature->_fullName));
+    }
+
+    if (!yield) {
+        blueprint->returnAll(yieldItems);
+    } else {
+        for (const auto* item : *yield->getItems()) {
+            const Symbol* symbol = item->getSymbol();
+
+            // TODO: handle the AS keyword there
```
cyrusknopf: Is this a leftover TODO? I think you showed us AS in the chat (maybe its a different PR)

storage/columns/ColumnDispatcher.h
LucLabarriere: Yeah I got a compile error once I changed the strong type definitions for LabelID, EdgeTypeID, etc. Yeah it's hard to maintain... At least this ColumnDispatcher is quite centralised, so it's not that bad, but there are also many other places where we do something similar

v2/AST/expr/Expr.h
LucLabarriere: I'll change that yeah, some people prefer to use references because it's still cheaper than building a string_view, not sure that's our take on this

v2/AST/FunctionDecls.cpp
LucLabarriere: That's also something I'm wondering, I'll add the break for now, maybe we always want to break or return?

v2/pipeline/procedures/EdgeTypesProcedure.cpp
LucLabarriere: Usually I try to avoid as much as possible putting a default case in switch cases because when a switch checks an enum, then you get a compile error if you forgot to handle a case which is pretty useful (if you add a default case, then it compiles regardless).

Here it's true that the fallthrough is ugly. it's only necessary because the compiler assumes someone could do something very silly, like attributing the step with a static cast:

```c++
_step = static_cast<Procedure::Step>(1000);
```

in which case it needs to handle the fallthrough possibility

v2/pipeline/processors/DatabaseProcedureProcessor.cpp
LucLabarriere: Good point thanks!

v2/plan/PipelineGenerator.cpp
LucLabarriere: I removed them because there was a bug where on the switch cases didn't return and it took me some time to see the issue (because I was focusing on the breaks rather than the returns). In the end it doesn't matter, I can add them back

v2/plan/PipelineGenerator.cpp
LucLabarriere: Thanks!

v2/AST/expr/Expr.h
cyrusknopf: Ah interesting, I didn't know it would be cheaper, keep it if you prefer

v2/pipeline/procedures/EdgeTypesProcedure.cpp
cyrusknopf: Yeah I agree on the point of default case

v2/plan/PipelineGenerator.cpp
cyrusknopf: Ooh interesting, was it a bug with the switch?

v2/AST/expr/Expr.h
LucLabarriere: In this case it shouldn't matter because everything is inlined, but sometimes, you can avoid unnecessary copies

v2/plan/PipelineGenerator.cpp
LucLabarriere: yeah we were just missing a return and doing break instead, which was causing to trigger the exception after the switch. It's not that big of a deal, I added back the breaks

query/AST/QueryCommand.h
ghost: That is weird, I think we hit some UB that somehow just outputs empty strings. Maybe somehow it happened because of constexpr??

LucLabarriere: I have absolutely no idea ahah, it's even surprising that it was working in the cli

PR #107, opened 2025-11-06
Account for tombstones in Conflict Checking

cyrusknopf:
Context:
 - [Relevant Notion Page](https://www.notion.so/turingbio/Iterators-Filtering-2a33aad664c8806dbd46eafdb7762b6e)

Changes:
- Conflict checking for unseen edges which were created on main and would-be inadvertently deleted (resulting in a conflict) are now only accounted for if they are not deleted (i.e. do not have a tombstone)
- `addHangingEdges` only adds edges which are not already deleted
- Unit tests for all above cases

- Add `VersionControlException` for execution-time errors which are not directly related to the pipeline, and so `PipelineException` does not really make sense

### DIFF
--- a/storage/versioning/ChangeConflictChecker.cpp
+++ b/storage/versioning/ChangeConflictChecker.cpp
@@ -185,11 +185,11 @@ void ChangeConflictChecker::checkDeletedNodeConflicts(const ConflictCheckSets& w
 }
 
 void ChangeConflictChecker::checkDeletedEdgeConflicts(const ConflictCheckSets& writes,
-                                              const CommitWriteBuffer& writeBuffer) {
+                                                      const CommitWriteBuffer& writeBuffer) {
     const CommitWriteBuffer::DeletedEdges& deletedEdges = writeBuffer.deletedEdges();
-    for (const EdgeID deletedEdge : deletedEdges) [[unlikely]] {
+    for (const EdgeID deletedEdge : deletedEdges) {
         const EdgeID newID = _entityIDRebaser.rebaseEdgeID(deletedEdge);
-        if (writes.writtenEdges.contains(newID)) {
+        if (writes.writtenEdges.contains(newID)) [[unlikely]] {
             std::string errorMsg = fmt::format(
                 "This change attempted to delete Edge {} "
                 "(which is now Edge {} on main) which has been modified on main.",


storage/versioning/ChangeConflictChecker.cpp
```diff
@@ -84,32 +85,51 @@ void ChangeConflictChecker::checkNewEdgesIncidentToDeleted(const CommitData& lat
...
+            const EdgeRecord& record = outEdgesIt.get();
+            const EdgeID& edgeID = record._edgeID;
+
+            if (!mostRecentView.tombstones().contains(edgeID)) [[unlikely]] {
+                std::string errorMsg = fmt::format(
+                    "Submit rejected: Commits on main have created an edge (ID: {}) "
+                    "incident to Node {}, which this Change attempts to delete.",
+                    edgeID, record._nodeID);
+                throw VersionControlException(std::move(errorMsg));
+            }
```
LucLabarriere, line 109: That makes me think of something. Doing conflict checking like this makes sense for short-lived changes, but maybe at some point people will use changes as long-lived branches, and will want to view the list of all conflicts so that they can resolve them manually.

Do you think that it can be done ? even if it requires to duplicate this whole class, as long as we can do it in the future it's fine

storage/versioning/ChangeConflictChecker.cpp
```diff
@@ -84,32 +85,51 @@ void ChangeConflictChecker::checkNewEdgesIncidentToDeleted(const CommitData& lat
...
-                  "node this Change attempts to delete.");
+
+        for (; outEdgesIt.isValid(); outEdgesIt.next()) {
+            const EdgeRecord& record = outEdgesIt.get();
+            const EdgeID& edgeID = record._edgeID;
+
+            if (!mostRecentView.tombstones().contains(edgeID)) [[unlikely]] {
+                std::string errorMsg = fmt::format(
+                    "Submit rejected: Commits on main have created an edge (ID: {}) "
+                    "incident to Node {}, which this Change attempts to delete.",
```
LucLabarriere, line 106: I think the error message "incident to node {}" is missleading since it is an outgoing edge (or maybe my english's not good enough)

storage/versioning/ChangeConflictChecker.cpp
```diff
@@ -150,23 +174,27 @@ void ChangeConflictChecker::checkDeletedNodeConflicts(const ConflictCheckSets& w
...
+            throw VersionControlException(std::move(errorMsg));
         }
     }
 }
 
 void ChangeConflictChecker::checkDeletedEdgeConflicts(const ConflictCheckSets& writes,
                                               const CommitWriteBuffer& writeBuffer) {
     const CommitWriteBuffer::DeletedEdges& deletedEdges = writeBuffer.deletedEdges();
-    for (const EdgeID deletedEdge : deletedEdges) {
+    for (const EdgeID deletedEdge : deletedEdges) [[unlikely]] {
```
LucLabarriere: Why the [[unlikely]] on the for loop?

LucLabarriere: If possible in general, I think we should use [[unlikely]] only in case of truly unlikely cases (as in, if it happens, it's a bug). Or in really hot code paths

storage/versioning/CommitWriteBuffer.cpp
```diff
@@ -47,22 +47,28 @@ void CommitWriteBuffer::addHangingEdges(const GraphView& view) {
...
     }
 
     { // Add the out edges of all deleted nodes
         GetOutEdgesRange outEdgesRg {view, &deletedNodesCol};
         for (const EdgeRecord& record : outEdgesRg) {
-            _deletedEdges.insert(record._edgeID);
+            // Only add edges which are not already deleted
+            if (!view.tombstones().containsEdge(record._edgeID)) {
+                _deletedEdges.insert(record._edgeID);
+            }
```
LucLabarriere, line 61: Actually, since _deletedEdges is a unordered_set, isn't it cheaper to just do the insertion anyway?

storage/versioning/ChangeConflictChecker.cpp
rjb32: Yes unlikely in a loop is a bit much. Let's no try to add them everywhere too much

cyrusknopf: Oops this one is a mistake, I was only meant to add it to the error throw underneath

storage/versioning/CommitWriteBuffer.cpp
cyrusknopf: The concern here is not repeated insertion into `_deletedEdges`, because each edge record is only ever encountered once in the loop. The problem is that `_deletedEdge` would could contain edges which already appear in the edge tombstones in the graph, which means when this commit comes to submit, these duplicates will be reinserted into the `TombstoneSet<EdgeID>`. This is fine currently since the tombstone set is a wrapper for unordered set, but in the future if we change it to a datastructure where repeated insertions are not idempotent, then this could cause problems. I think it makes sense from a correctness point of view to do this check.

LucLabarriere: Ok agree yeah

storage/versioning/ChangeConflictChecker.cpp
cyrusknopf: Both source and target nodes of an edge are said to be 'incident' to that edge (at least from the books etc. I've read), but actually since we are checking Out edges and In edges separately, I can explicitly say "with source {}" or "with target {}" which might be nicer

storage/versioning/ChangeConflictChecker.cpp
cyrusknopf: Yeah I think that should be possible, it will be a very similar class except instead of throwing an exception upon encountering a conflict, we can add all conflicts to some data structure and then present it to the user somehow

storage/versioning/ChangeConflictChecker.cpp
LucLabarriere: Ok good to know thanks, you can decide which sentence is better

storage/versioning/ChangeConflictChecker.cpp
LucLabarriere: Great thanks!

rjb32: Let's do it in the future guys in another PR

PR #110, opened 2025-11-07
GetInEdges + GetEdges processors, bug fixes and unit tests

LucLabarriere:
- Fixes a bug when we have multiple bugs
- LineContainer test util to check if all required lines are present (order independent), can  be generated by nested loops/pipelines and compared together
- Dynamic chunk size -> great for unit testing
- Added GetEdgesIterator, GetEdgesChunkWriter, GetEdgesProcessor (returns all edges, regardless of the direction (it basically returns the same edge twice, -> cypher compliant)

### DIFF
--- a/v2/pipeline/processors/GetEdgesProcessor.cpp
+++ b/v2/pipeline/processors/GetEdgesProcessor.cpp
@@ -38,6 +38,8 @@ GetEdgesProcessor* GetEdgesProcessor::create(PipelineV2* pipeline) {
 }
 
 void GetEdgesProcessor::prepare(ExecutionContext* ctxt) {
+    _ctxt = ctxt;
+
     ColumnNodeIDs* nodeIDs = dynamic_cast<ColumnNodeIDs*>(_inNodeIDs.getNodeIDs()->getColumn());
     ColumnVector<size_t>* indices = dynamic_cast<ColumnVector<size_t>*>(_outEdges.getIndices()->getColumn());
     ColumnEdgeIDs* edgeIDs = dynamic_cast<ColumnEdgeIDs*>(_outEdges.getEdgeIDs()->getColumn());
@@ -49,7 +51,6 @@ void GetEdgesProcessor::prepare(ExecutionContext* ctxt) {
     _it->setEdgeIDs(edgeIDs);
     _it->setOtherIDs(otherNodes);
     _it->setEdgeTypes(edgeTypes);
-    _chunkSize = ctxt->getChunkSize();
 
     markAsPrepared();
 }
@@ -60,7 +61,7 @@ void GetEdgesProcessor::reset() {
 }
 
 void GetEdgesProcessor::execute() {
-    _it->fill(_chunkSize);
+    _it->fill(_ctxt->getChunkSize());
 
     if (!_it->isValid()) {
         _inNodeIDs.getPort()->consume();
--- a/v2/pipeline/processors/GetEdgesProcessor.h
+++ b/v2/pipeline/processors/GetEdgesProcessor.h
@@ -27,7 +27,6 @@ private:
     PipelineNodeInputInterface _inNodeIDs;
     PipelineEdgeOutputInterface _outEdges;
     std::unique_ptr<GetEdgesChunkWriter> _it;
-    size_t _chunkSize = 0;
 
     GetEdgesProcessor();
     ~GetEdgesProcessor();


storage/dataframe/Dataframe.h
```diff
@@ -56,6 +57,17 @@ class Dataframe {
...
+        requires std::is_base_of_v<Column, T>
+    const T* getColumn(ColumnTag tag) const {
+        const auto* col = getColumn(tag);
+        if (!col) {
+            return nullptr;
+        }
+
+        return dynamic_cast<const T*>(col->getColumn());
+    }
+
```
LucLabarriere, line 70: Can be convenient to avoid having to do dynamic_cast<T>(getColumn(tag)->getColumn())

test/v2/pipeline/LineContainer.h
```diff
@@ -0,0 +1,96 @@
...
+                return hashImpl(line, std::index_sequence_for<ValueTypes...> {});
+            }
+        };
+
+        struct Predicate {
+            bool operator()(const Line& lhs, const Line& rhs) const {
+                return lhs._values == rhs._values;
+            }
+        };
+    };
```
LucLabarriere, line 33: Do we want to keep this?

We create a "LineContainer" by specifying the types of all column elements: `LineContainer<NodeID, EdgeID, NodeID, EdgeTypeID>`, then we can add lines with `container.add({col1[i], col2[i], col3[i], col4[i]})`, it hashes the line, stores an entry to the map and increment the count. Then we can call container1.equals(container2). This allows to compare the results of nested for loops vs pipeline results (order independent).

test/v2/pipeline/PipelineTest.cpp
```diff
@@ -154,7 +154,7 @@ TEST_F(PipelineTest, scanNodesExpand1) {
     const ColumnTag scanOutNodeIDsTag = scanNodesOut.getNodeIDs()->getHeader().getTag();
 
     PipelineEdgeOutputInterface& getOutEdgesOut = builder.addGetOutEdges();
-    const ColumnTag getOutEdgesOutTargetNodesTag = getOutEdgesOut.getTargetNodes()->getHeader().getTag();
+    const ColumnTag getOutEdgesOutTargetNodesTag = getOutEdgesOut.getNeighbourNodes()->getHeader().getTag();
```
LucLabarriere: Do you keep this name? the issue is that `getTargetNodes` is incompatible with `GetInEdges` (since it should be `getSourceNodes` instead). `getNeighourNodes` is compatible with both outgoing and incoming edges. `getOtherNodes` is also a possibility

test/v2/pipeline/processors/GetInEdgesProcessorTest.cpp
```diff
@@ -0,0 +1,103 @@
...
+        _graph = _env->getSystemManager().createGraph("simpledb");
+        SimpleGraph::createSimpleGraph(_graph);
+    }
+};
+
+TEST_F(GetInEdgesProcessorTest, test) {
+    auto [transaction, view, reader] = readGraph();
+
+    LineContainer<NodeID, EdgeID, NodeID, EdgeTypeID> expLines;
+    LineContainer<NodeID, EdgeID, NodeID, EdgeTypeID> resLines;
```
LucLabarriere, line 24:
```
expLines = expected
resLines = results
```

test/v2/pipeline/processors/GetInEdgesProcessorTest.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    for (const NodeID tgtID : reader.scanNodes()) {
+        ColumnVector<NodeID> tmpNodeIDs = {tgtID};
+
+        for (const EdgeRecord& edge : reader.getInEdges(&tmpNodeIDs)) {
+            expLines.add({edge._nodeID, edge._edgeID, edge._otherID, edge._edgeTypeID});
+        }
+    }
```
LucLabarriere, line 37: Nested loop approach, doesn't take into account tombstones yet, I'll have to add this manually

test/v2/pipeline/processors/GetInEdgesProcessorTest.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    const auto edgeInterface = _builder->addGetInEdges();
+    const ColumnTag edgeIDsTag = edgeInterface.getEdgeIDs()->getTag();
+    const ColumnTag edgeTypesTag = edgeInterface.getEdgeTypes()->getTag();
+    const ColumnTag srcIDsTag = edgeInterface.getNeighbourNodes()->getTag();
+
+    const auto callback = [&](const Dataframe* df, LambdaProcessor::Operation operation) -> void {
+        if (operation == LambdaProcessor::Operation::RESET) {
+            return;
+        }
```
LucLabarriere, line 53: > [!Warning]
> This must not be forgotten

test/v2/pipeline/processors/GetInEdgesProcessorTest.cpp
```diff
@@ -0,0 +1,103 @@
...
+    resLines.clear();
+    EXECUTE(view, 2);
+    resLines.print(std::cout);
+    EXPECT_TRUE(resLines.equals(expLines));
+
+    fmt::println("\n- Executing pipeline with chunk size 1...");
+    resLines.clear();
+    EXECUTE(view, 1);
+    resLines.print(std::cout);
+    EXPECT_TRUE(resLines.equals(expLines));
```
LucLabarriere, line 104: `EXECUTE` macro is not necessarily interesting, it's defined in `ProcessorTester`, allows to run the pipeline multiple times with different chunk sizes each time

```c++
#define EXECUTE(view, chunkSize)                          \
    {                                                     \
        ExecutionContext execCtxt(view);                  \
        execCtxt.setChunkSize(chunkSize);                 \
        _pipeline.clear();                                \
        PipelineExecutor executor(&_pipeline, &execCtxt); \
        executor.execute();                               \
    }
```

test/v2/pipeline/processors/ProcessorTester.h
```diff
@@ -0,0 +1,54 @@
...
+        return std::make_tuple(transaction, view, reader);
+    }
+
+protected:
+    std::unique_ptr<TuringTestEnv> _env;
+    std::unique_ptr<PipelineBuilder> _builder;
+    Graph* _graph {nullptr};
+
+    PipelineV2 _pipeline;
+};
```
LucLabarriere, line 54: Not sure this is necessary tbh

v2/pipeline/ExecutionContext.h
```diff
@@ -14,8 +16,12 @@ class ExecutionContext {
     }
 
     const GraphView& getGraphView() const { return _graphView; }
+    size_t getChunkSize() const { return _chunkSize; }
+
+    void setChunkSize(size_t chunkSize) { _chunkSize = chunkSize; }
 
 private:
     const GraphView& _graphView;
+    size_t _chunkSize = ChunkConfig::CHUNK_SIZE;
```
LucLabarriere, line 25: Right now, I set the chunk size like this, not sure this is the right way. However, it's really useful for unit tests

v2/pipeline/PipelineV2.cpp
```diff
@@ -36,3 +36,11 @@ void PipelineV2::addBuffer(PipelineBuffer* buffer) {
...
     _ports.push_back(port);
 }
+
+void PipelineV2::clear() {
+    for (Processor* processor : _processors) {
+        processor->_prepared = false;
+        processor->_finished = false;
+        processor->_scheduled = false;
+    }
+}
```
LucLabarriere, line 46: To allow to re-run the pipeline

v2/pipeline/processors/GetInEdgesProcessor.h
```diff
@@ -0,0 +1,36 @@
...
+    void execute() override;
+
+    PipelineNodeInputInterface& inNodeIDs() { return _inNodeIDs; }
+    PipelineEdgeOutputInterface& outEdges() { return _outEdges; }
+
+private:
+    PipelineNodeInputInterface _inNodeIDs;
+    PipelineEdgeOutputInterface _outEdges;
+    std::unique_ptr<GetInEdgesChunkWriter> _it;
+    size_t _chunkSize = 0;
```
LucLabarriere: Not sure this is the right way to store the _chunkSize. Maybe the processors should store a pointer to the ExecutionContext instead? Might be convenient once we have other configs/context objects

v2/pipeline/processors/GetOutEdgesProcessor.cpp
```diff
@@ -60,10 +62,10 @@ void GetOutEdgesProcessor::reset() {
 }
 
 void GetOutEdgesProcessor::execute() {
-    _inNodeIDs.getPort()->consume();
-    _it->fill(ChunkConfig::CHUNK_SIZE);
+    _it->fill(_chunkSize);
 
     if (!_it->isValid()) {
+        _inNodeIDs.getPort()->consume();
```
LucLabarriere, line 68: This fixes the multi chunk bug

v2/pipeline/PipelineV2.cpp
LucLabarriere: Maybe we should put do the `proc->prepare()` thing in the init() function instead of in the for loop

storage/dataframe/Dataframe.h
rjb32: Yes I agree

test/v2/pipeline/LineContainer.h
rjb32: Yes it seems to be interesting to solve the problem of testing in the short term at least.

test/v2/pipeline/PipelineTest.cpp
rjb32: Let's use getOtherNodes to be consistent with the rest

v2/pipeline/ExecutionContext.h
rjb32: I think it's a good idea, I am not sure it will have an impact on performance since anyway most of the fill methods can not be inlined

v2/pipeline/processors/GetInEdgesProcessor.h
rjb32: Yes I agree it should be more proper to store a pointer to ExecutionContext

rjb32:
Let's put the dynamic chunkSize in ExecutionContext and just get it from there in processors

LucLabarriere:
> Let's put the dynamic chunkSize in ExecutionContext and just get it from there in processors

Done, only GetEdgesProcessor was missing it (forgot)

PR #120, opened 2025-11-10
v2 PipelineGenerator: GetEdgeTarget and GetOutEdges

rjb32:
* Generate GetOutEdges in PipelineGenerator
* Support GetEdgeTarget node by doing nothing because pipeline interface connect does it already for us
* Fix Materialize bug: PlanGraph does not necessarily provide a MaterializeNode when there are no branches, it is assumed that ProduceResultsNode does a materialize anyway. In PipelineGenerator we need to add a materialize processor if we have a non-trivial materialize in flight

### DIFF
--- a/v2/pipeline/PipelineBuilder.h
+++ b/v2/pipeline/PipelineBuilder.h
@@ -34,12 +34,13 @@ public:
 
     // Sources
     PipelineNodeOutputInterface& addScanNodes();
+    PipelineBlockOutputInterface& addLambdaSource(const LambdaSourceProcessor::Callback& callback);
 
     // Get edges
     PipelineEdgeOutputInterface& addGetOutEdges();
     PipelineEdgeOutputInterface& addGetInEdges();
     PipelineEdgeOutputInterface& addGetEdges();
-    void addLambda(const LambdaProcessor::Callback& callback);
+
 
     // Properties
     template <SupportedType T>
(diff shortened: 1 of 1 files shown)


v2/pipeline/processors/MaterializeData.h
```diff
@@ -31,6 +31,8 @@ class MaterializeData {
 
     void setOutput(Dataframe* output) { _output = output; }
 
+    bool isSingleStep() const { return _step == 0; }
```
cyrusknopf, line 34: Is `_step` like an array index into the number of steps? Or is it the number of steps?

v2/pipeline/PipelineBuilder.h
```diff
@@ -50,9 +50,14 @@ class PipelineBuilder {
     PipelineBlockOutputInterface& addLimit(size_t count);
     PipelineValueOutputInterface& addCount();
 
+    // Lambda sink
```
cyrusknopf, line 54: Should this say sink or source

v2/pipeline/processors/MaterializeData.h
rjb32: It's the index of the current step

v2/pipeline/PipelineBuilder.h
rjb32: Ah you're right good catch, wrong line

PR #121, opened 2025-11-10
Implement The Hash Join Processor

ghost:


storage/RowStore.cpp
```diff
@@ -0,0 +1,83 @@
+#include "RowStore.h"
+#include "ID.h"
+#include "columns/ColumnVector.h"
+#include "dataframe/Dataframe.h"
+
```
rjb32: Too many blank space

storage/RowStore.cpp
```diff
@@ -0,0 +1,83 @@
+#include "RowStore.h"
```
rjb32, line 1: Insert a blank line between the include of your header and the rest of the includes

storage/RowStore.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "columns/ColumnVector.h"
+#include "dataframe/Dataframe.h"
+
+
+using namespace db;
+
+RowOffset RowStore::insertRow(const Dataframe* frame, size_t rowNumber) {
+    // for now assume that we are only dealing with NodeID/NodeIDs
+    const size_t numCols = frame->size();
+    const size_t rowSize {sizeof(NodeID) * numCols};
```
rjb32: Use rather the assignment style of rowSize = expression

storage/RowStore.h
```diff
@@ -0,0 +1,81 @@
+#pragma once
+
+#include <array>
+#include <cstring>
+#include <list>
+#include "BioAssert.h"
```
rjb32, line 8: Insert a blank line before BioAssert include

storage/RowStore.h
```diff
@@ -0,0 +1,81 @@
...
+
+namespace db {
+class Dataframe;
+class RowStore;
+class ColumnTag;
+
+class Slab {
+public:
+    static constexpr size_t BATCH_SIZE = 8 * 1024 * 1024;
```
rjb32, line 19: Insert a blank line after the constant declaration

storage/RowStore.h
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    size_t getCurrentOffset() const {
+        return _freeData - _data.data();
+    }
+
+private:
+    friend RowStore;
+
+    std::array<char, BATCH_SIZE> _data;
```
rjb32: Use uint8_t instead of char for bytes, it will make your intent clearer

storage/RowStore.h
```diff
@@ -0,0 +1,81 @@
...
+namespace db {
+class Dataframe;
+class RowStore;
+class ColumnTag;
+
+class Slab {
+public:
+    static constexpr size_t BATCH_SIZE = 8 * 1024 * 1024;
+    template <typename T>
+    const T* getValue(size_t offset) {
```
rjb32, line 22: Check that offset+sizeof(T) <= BATCH_SIZE to check that the object fits at that offset

storage/RowStore.h
```diff
@@ -0,0 +1,81 @@
...
+        bioassert(offset < BATCH_SIZE);
+
+        return static_cast<const T*>(_data.data() + offset);
+    }
+
+    template <typename T>
+    size_t setValue(T val) {
+        size_t offset = _freeData - _data.data();
```
rjb32: offset can be const

storage/RowStore.h
```diff
@@ -0,0 +1,81 @@
...
+        bioassert(_freeData + sizeof(T) <= _data.data() + BATCH_SIZE);
+
+        std::memcpy(_freeData, &val, sizeof(T));
```
rjb32, line 33: That can just be: `*(T*)_freeData = val;`

storage/RowStore.h
```diff
@@ -0,0 +1,81 @@
...
+    char* _freeData {_data.data()};
+
+    // Row batch allocates X Mb of data -> then we write our values into row batch. Adding a new new batch as needed.
+    // The hashmaps store information such as batch pointer + offset. We have dimension information either stored in the
+    // the processor itself or maybe derived from looking at the input port?
+};
+
+struct RowOffset {
+    Slab* slab;
+    size_t slabOffset;
```
rjb32: Default initialize members of the struct

storage/RowStore.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+RowOffset RowStore::insertRow(const Dataframe* frame, size_t rowNumber, ColumnTag* joinKey) {
+    // for now assume that we are only dealing with NodeID/NodeIDs
+
+    // In the case when we do not want to insert the joinKey
+    const size_t numCols = frame->size() - 1;
+    const size_t rowSize {sizeof(NodeID) * numCols};
+    if (_dataList.back().getRemainingSize() < rowSize) {
+        _dataList.emplace_back();
```
rjb32, line 15: Add a comment to indicate that it creates actually a new slab

storage/RowStore.cpp
```diff
@@ -0,0 +1,83 @@
...
+        const auto& val = cols[i]->as<ColumnVector<NodeID>>()->data()[rowNumber];
+        _dataList.back().setValue<NodeID>(val);
+    }
+
+    return {&currentSlab, offset};
+}
+
+char* RowStore::getRow(RowOffset row) {
+    const auto& slab = row.slab;
+    const auto& slabOffset = row.slabOffset;
```
rjb32: slabOffset does not a reference it's a size_t?

test/v2/pipeline/processors/HashJoinProcessorTest.cpp
```diff
@@ -0,0 +1,321 @@
...
+        ASSERT_TRUE(col != nullptr);
+        ASSERT_TRUE(col->empty());
+        col->resize(RHS_NUM_ROWS);
+        std::iota(col->begin(), col->end(), +1);
+
+        ASSERT_EQ(df->size(), RHS_NUM_COLS);
+        isFinished = true;
+    };
+    ColumnTag joinKey {0};
+    { // Wire up the cartesian product to the two inputs
```
rjb32: Why it says cartesian product here?

test/v2/pipeline/processors/HashJoinProcessorTest.cpp
```diff
@@ -0,0 +1,321 @@
...
+    constexpr size_t NUM_MATCHES = 5;
+    /*
+     * Output Dataframe should be looking like:
+     * 101 201 301 401 1
+     * 102 202 302 402 2
+     * 103 203 303 403 3
+     * 104 204 304 404 4
+     * 105 205 305 405 5
+     */
+    const auto callback = [&](const Dataframe* df, LambdaProcessor::Operation operation) -> void {
```
rjb32, line 135: Add a boolean and an assert outside to check that the lambda is executed at least once to prevent silent failure

test/v2/pipeline/processors/HashJoinProcessorTest.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    constexpr size_t NUM_MATCHES = 4;
+    /*
+     * Output Dataframe should be looking like:
+     * 102 202 302 402 2
+     * 104 204 304 404 4
+     * 106 206 306 406 6
+     * 108 208 308 408 8
+     */
+    const auto callback = [&](const Dataframe* df, LambdaProcessor::Operation operation) -> void {
```
rjb32, line 1318: Add a boolean and an assert outside to check that the lambda is executed at least once to prevent silent failure

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,148 @@
...
+}
+
+void HashJoinProcessor::execute() {
+    // things to do still:
+    // 2. resize dataframe
+    // 3. think about value hash joins and how they might work.
+    // 4.write tests
+
+    if (!(_leftInput.getPort()->hasData() || _rightInput.getPort()->hasData())) {
+        finish();
```
rjb32: Are you sure?

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    Dataframe* outPutDf = _output.getDataframe();
+
+    const auto* leftCol = leftDf->getColumn<ColumnVector<NodeID>>(*_joinKey);
+    const auto* rightCol = rightDf->getColumn<ColumnVector<NodeID>>(*_joinKey);
+
+    // check left inputs
+    for (size_t i = 0; i < leftCol->size(); ++i) {
+        if (_rightMap.contains((*leftCol)[i])) {
+            const auto& rows = _rightMap[(*leftCol)[i]];
```
rjb32: Try to reuse the result of the iterator because you are doing the lookup twice actually with the "contains"

test/v2/pipeline/processors/HashJoinProcessorTest.cpp
```diff
@@ -0,0 +1,321 @@
...
+#include "processors/LambdaSourceProcessor.h"
+#include "columns/ColumnIDs.h"
+#include "SystemManager.h"
+#include "SimpleGraph.h"
+
+using namespace db;
+using namespace db::v2;
+using namespace turing::test;
+
+class HashJoinProcessorTest : public ProcessorTester {
```
rjb32, line 18: Add a case where the match result is empty and where the match result is a subset of the left and right

storage/RowStore.h
ghost: I checked in godbolt and the assembly is the same, isn't it a bit clearer to leave memcpy to indicate intention?

rjb32: Perhaps, it's just a bit weird sometimes when you memcpy something which is not multiple objects or just a blob of bytes. But keep it this time

v2/pipeline/processors/HashJoinProcessor.h
```diff
@@ -0,0 +1,99 @@
+#pragma once
+
+#include <unordered_map>
+
+#include "ID.h"
```
rjb32: Reorder includes

v2/pipeline/processors/HashJoinProcessor.h
```diff
@@ -0,0 +1,99 @@
...
+
+class HashJoinProcessor : public Processor {
+public:
+    static HashJoinProcessor* create(PipelineV2* pipeline, ColumnTag* joinKey);
+
+    PipelineBlockInputInterface& rightInput() { return _rightInput; }
+    PipelineBlockInputInterface& leftInput() { return _leftInput; }
+    PipelineBlockOutputInterface& output() { return _output; }
+
+    size_t processRightStream(const Dataframe* rightDf, Dataframe* outDf, size_t rowsRemaining, size_t& totalRowsInserted);
```
rjb32: These should be private member functions

v2/pipeline/processors/HashJoinProcessor.h
```diff
@@ -0,0 +1,99 @@
...
+    PipelineBlockInputInterface _rightInput;
+    PipelineBlockInputInterface _leftInput;
+    PipelineBlockOutputInterface _output;
+
+    // Information about dimensions of out inputs
+    size_t _leftRowLen {0};
+    size_t _rightRowLen {0};
+
+    // map to a vector of offsets
+    // the offsets point to the corresponding row in our
```
LucLabarriere, line 71: Do you mean in out?

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -47,6 +48,38 @@ void concatDataframeShape(LocalMemory* mem,
...
 
+void createHashJoinDataFrameShape(LocalMemory* mem,
+                                  DataframeManager* dfMan,
+                                  Dataframe* leftSrc,
+                                  Dataframe* rightSrc,
+                                  ColumnTag* joinKeyTag,
+                                  Dataframe* dest) {
+    // Add left input columns
+    for (const NamedColumn* col : leftSrc->cols()) {
+        if (col->getTag() == *joinKeyTag) {
```
LucLabarriere: As said in pm, generally speaking, can it be simple `if tag already in output df`, if not i think you still need to check if the column was not already added (in case the column is shared by both branches: `if tag == joinKey OR tag in output df`

Can you make ColumnTag a value instead of a pointer ? It default constructs to an invalid state that you can check with tag.isValid() so the pointer shouldn't be needed I think

LucLabarriere: I General I suspect that if we join two blocks with columns:
- First block: A B C D E F G
- Second block: A F G H I J
Then, the resulting block should have A B C D E F G H I J, regardless of which one is the join key

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,410 @@
+#include "HashJoinProcessor.h"
+#include "ExecutionContext.h"
+#include "columns/ColumnVector.h"
+#include "dataframe/Dataframe.h"
+
+using namespace db;
+using namespace db::v2;
+
+HashJoinProcessor::HashJoinProcessor(ColumnTag* joinKey)
+
```
LucLabarriere, line 32: Empty line

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+    const Dataframe* rightDf = _rightInput.getDataframe();
+
+    Dataframe* outDf = _output.getDataframe();
+
+    if (_copyState.isValid()) {
+        // resize the output columns to fit the new rows we are inserting
+        const size_t newOutputSize = std::min(_ctxt->getChunkSize(),
+                                              _copyState._offsetVec->size() - _copyState._rowOffsetIdx);
+        for (const auto& col : outDf->cols()) {
+            const auto& colVector = static_cast<ColumnVector<NodeID>*>(col->getColumn());
```
LucLabarriere: replace by `const auto*` instead of `const auto&`

v2/pipeline/PipelineBuilder.cpp
ghost: I think you are right in the case where we have repeating columns in the input blocks we would ideally want to skip them. But I can't think of an example where we would come to that situation - because when we have similar Var Expr across queries don't they just get joined anyways?

For example I generated the mermaid chart [here](https://mermaid.live/edit#pako:eNqVVl1v2kAQ_CvRPZPovu_sh0pRoFGqKFSE9KGlctzgQNoAkQv9ivLf6wGvGnFU8vK047u5W8br3XkWd6tpJXJx_7j6eTcv6_XRuD9ZHjU_-WkibnchfkVxfXZ6VVwN-4Protg9v52Iz7tI7W_-cDqiXbsHy3JRFUV-VO5T9T717cXleDDa3pReZPZ3nw_GxfBmXAz654cSs50T-yH3ue4_meGq9CZ_KLPt1vHpCHHCCJ1z-7JPjSzRMqZoSnZXTSVkxZJNabZuynRO7y7hWpZyynGl892l0wk58KSLfOmyzulNkw9VsqTTiimd1t2lMwnZsKTTli2ddp3TqxKu50kXuNLF7tLZhJyxpDOSLZ3pPh7uEy5zPnAHhGFMCJeQeSPC8GeE4ZaCYZSCT8i8UrD8UrDdS2GWcHmlYNlegVEKISHzSsHyS8F29wvzhMszDJbrGBzDMcSEzHMMju8YXHfH8JBweY7BcR2DYziGLCHzHIPjOwbX3TF8TSwxzzH4pA7eDS-uDuxLCuD9aNi_ORsUo8H1zeX4gMjy-PhNa7dUE5pXYRvrJm69rGnCttXbJmy7pGvCdnb6JmyrODRh-1ZiE7ZtIcNhisw0AF2OK5Uk6wtgydQCOHKpAJQlblaB_CVAJL8I4MkN4j_S0chAKzJu2JaRLcMK_WlkoC1ZKABNBgnAk_UBCORlABw5FQA6GhkYSa4CK5E8A3QkebeyBxrvALSCDIyjwQzgadACWBqjr09DBiaj4YeXJGmYYSXSqMKKpiEEYGiqACiaGQD08pGB9dTfASx1bwA6GhnYjBotQKA2ikpS1CIBNLU8AEkNDcBSqwJw1HsADHUWgEBdAyBSFwCgOkUGno5GBl7T54htGX1sUEeLnpjVD1ORr-tN1ROLql6UgOIZ2yZiPa8W1UTkTTgt628TMVm-NJyncvlxtVoQrV5tZnOR35eP3xu0eZqW66r_UM7q8t-Wajmt6rPVZrkWuXd-e4bIn8UvkTvrTpyJMYZgpdamJ36L_NgHd6Js8FFHL4PO9EtP_Nleqk8yn8lMSSVNs-LNy1-z3y74).

v2/pipeline/processors/HashJoinProcessor.h
ghost: I seemed to have deleted a line from the comments 🙏

v2/pipeline/PipelineBuilder.cpp
LucLabarriere: Yeah in your example only A is common between two branches but it's the joining key so it works, but consider this one for the second join: [mermaid](https://mermaid.live/edit#pako:eNqVVttum0AQ_RW0kt9ItBfM7aGSFbtRqjSuHCeRWiqyCcRGNWARHCe1_O_dASZOwVVZP82s5-wcDmeH3ZHHPIqJTwaDnZFkSekbu4CUyziNA-IbAYniJ7lZlQExg8w4_JqaW1kk8mEVP0PxTq1uC7mugU9y9az2MPZt4NMq3z4uZVEi5nFTvDTdVkkWy6LdrPULSCqLX1G-zUabMr-rO743_A8UCK6TbHGXROWybsoopUD0L6DK9oNBkL2zNebjuoD-CMj9oTYMr89GV-HVdDy5DsN6_T4gP-uItYtvRzOsqhcymcZh6BuvrI3lbezni8v5ZFa16nYS7erzyTyc3szDyfj8GDOrN7MX2sYO_8EMWnU72ceYVaXz0QziDsLprxpvY10t1TxN1RjtL1vnhTKmpRvj2sIx0Zue7GAtLeXYUFc6u790nbfKHD3pXH3pvN703jonlWpJx5mmdJz3l050wEJLOm5pS8d1rcA1rGB1wHpW4PpW4P2t8NAZw3pWELpWEBpWGHbAelYQ-lYQHSt8mV5cHamze9Z1Xva32XR8czYJZ5Prm8v5EYXoycmnZn4wFYpDyJuJzFXchEKFjUyWCu3m-6bCxnc2bNFUOCr2DiFvll0VO83nBHZm-KmABJlAf0ZxsENi4ciGZIgzGBKkDDSYg9MTEhenISQ2zjqgiAkwEA6OJSjzcOgAZYEDBRILBwQkHI__hwfmwIDjbsCAu3gIIbHxiAGG4uGBBOWtXoaHXod_BPoYEgt9CQlH1314HgEMXPQMbKbaEJMsiiQifllsYpOkcZFKSMkOCo9fX9V1TsHWMvue5ykii3yzWBK_ujmaZLOOZBmPE7koZPq-WsRZFBdn-SYriS88q9qE-DvyqtIhP2WMWsLjztChwnVM8kZ8j50yypnlUY_a3OZ7k_yuurJTSoWgnrA96jqcedwkcZSUefG1voVXl_H9H-AZ7wM)

This example comes from this query:
```
MATCH (x1)-->(x2)-->(a)-->(y), (x1)-->(x2)-->(b)-->(y) RETURN a, b
```

And in think in this case, the second join receives blocks from lhs and rhs and have columns that are shared: x1, v0, v3 + the joining key x2

ghost: That is an interesting example! I would have assumed that pipeline generator would have merge (x1)-->(x2) from both branches. So that is not the case? I actually have quite a few questions about this case that maybe we can discuss in the standup (will post updates of the conversation here for visibility)

LucLabarriere: Yeah no I don't think it should merge it because we can't "re-use" edge entities, so you can't do: `MATCH (a)-[e]-(b), (x)-[e]-(y)` (e cannot be used twice). So actually, you have to generate two separate branches and follow them. An optimized could be smart enough to know that we're doing the same thing on both branches here, but in general this is not the case:
```c++
MATCH (a)-[e1:EdgeType1]-(b), (a)-[e2:EdgeType2]-(b)
```

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+    for (size_t j = 0; j < rightDf->size(); ++j) {
+        if (cols[j]->getTag() == *_joinKey) {
+            continue;
+        }
+        auto* inPutCoulmn = static_cast<ColumnVector<NodeID>*>(cols[j]->getColumn());
+        auto* outPutColumn = static_cast<ColumnVector<NodeID>*>(outCols[_leftRowLen - 1 + j]->getColumn());
+
+        bioassert(outPutColumn->size() > totalRowsInserted + rowsToCopy);
+        std::fill_n(outPutColumn->begin() + totalRowsInserted, rowsToCopy, inPutCoulmn->data()[_rightInputIdx]);
+    }
```
cyrusknopf, line 472: If I understood correctly, we are assuming the join key is always a `NodeID`, but does this not assume that ALL columns inthe dataframe are `ColumnVector<NodeID>`? Is it not possible that they could be `EdgeID` as well? If so, I think it would be clearer in the intent to use `EntityID`, otherwise it could be misleading. Let me know if any of my assumptions/understanding is wrong

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+    const auto* inJoinKeyCol = static_cast<ColumnVector<NodeID>*>(rightDf->getColumn(*_joinKey)->getColumn());
```
cyrusknopf: Nitpick but I think using the `NamedColumn::as<T>` method is clearer, it reduces the line size and the number of calls to `getColumn`. You can probably use it a lot throughout this file if you feel it is beneficial.

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+    auto* outJoinKeyCol = static_cast<ColumnVector<NodeID>*>(outCols.back()->getColumn());
+
```
cyrusknopf: Why do you use the joinkey explicitly in the input DF but not to retrieve the joinkey column in the output DF? (Instead you use .back())

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+        if (totalSizeIncrease) {
+            const size_t newOutputSize = std::min(_ctxt->getChunkSize(), totalRowsInserted + totalSizeIncrease);
+            for (const auto& col : outDf->cols()) {
+                const auto& colVector = static_cast<ColumnVector<NodeID>*>(col->getColumn());
+                colVector->resize(newOutputSize);
+            }
+        }
+
+        if (rowsRemaining = processLeftStream(leftDf, outDf, rowsRemaining, totalRowsInserted); rowsRemaining == 0) {
```
cyrusknopf: Do we have a line column limit? My clang-format says its 90

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+            // If there are no rows remaining in our output chunk we can write the output
+            _output.getPort()->writeData();
+            hasWritten = true;
+
+            // if we aren't paused mid-row vector (no valid _copyState)
+            // and we don't have any more columns to read we can consume
+            // the leftInput (the case where we finish reading the left
+            // input stream and the output fits perfectly into a chunk)
+            if (!_copyState.isValid() && _leftInputIdx >= leftDf->getRowCount()) {
```
cyrusknopf: Is the `_leftInputIdx` inclusive? i.e. can we be at `_leftInputIDx == lefDf->getRowCount()` but still processing the final input index? If so should this be strictly greater than instead of greater than or equal?

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,410 @@
...
+    if (totalRowsInserted > 0) {
+        _output.getPort()->writeData();
+        hasWritten = true;
+    }
+
+    if(_leftInput.getPort()->isClosed() && _rightInput.getPort()->isClosed()) {
+        if(!hasWritten) {
+            _output.getPort()->writeData();
+        }
+        _output.getPort()->close();
```
cyrusknopf: I don't think you should call close on your output yourself, this is abstracted in the `finish` call

v2/pipeline/processors/HashJoinProcessor.h
```diff
@@ -0,0 +1,99 @@
...
+    RowStore _leftStore;
+
+    ColumnTag* _joinKey {nullptr};
+
+    // These indexes help us keep track of where to start iterating on the columns
+    // of the input
+    size_t _leftInputIdx {0};
+    size_t _rightInputIdx {0};
+
+    bool hasWritten {false};
```
LucLabarriere: rename _hasWritten

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: I think we do, I have the check off because it does weird formatting sometimes. I'll break down the line.

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: Since we are adding the join key to the last column in the output we know where it is, and I guess is minimally faster than doing a getColumn.

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: I think that is a good shout. I will change that!

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: I think that is a good idea, we could have EdgeIDs that would be included in the join. I will make the change.

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: We can actually have `_leftInputIdx > leftDf->getRowCount()` in the case where we finish processing the input but we haven't filled the chunk. But in that case we wouldn't be hitting this code actually as it's in the if condition that only runs if we have filled the output chunk. So this can actually just be `==` instead of `>=`. Thanks for bringing this up!

LucLabarriere:
If I understand correctly, I think the majority of the complexity comes from the fact that you assume a join might not finish in one chunk and you need to keep track of what was the previous state. But is it really necessary? Since you process one chunk at a time (whether it is a left chunk or right chunk), then the output should never exceed one chunk at a time, meaning you should be able to just process the full thing in one go:
- if Received 1 right chunk:
  - For each line in the chunk, check if the key is in the lhs map. If it is store the index in a temp vector.
  - Once the chunk has been fully checked, build the output chunk
- Else if received 1 left chunk:
  - For each line in the chunk, check if the key is in the lhs map. If it is store the index in a temp vector.
  - Once the chunk has been fully checked, build the output chunk
  
It is guaranteed that the output does not exceed 1 chunk size in this case.

Maybe it doesn't work like that and you wait for two chunks to be available instead of triggering immediately once you receive a single chunk (in a design where it triggers at soon as at least one chunk is available, the first execution doesn't produce an output, or produces an empty output)

v2/pipeline/processors/HashJoinProcessor.cpp
cyrusknopf: IMO explicitly getting the join key column is much clearer on intent

ghost:
> If I understand correctly, I think the majority of the complexity comes from the fact that you assume a join might not finish in one chunk and you need to keep track of what was the previous state. But is it really necessary? Since you process one chunk at a time (whether it is a left chunk or right chunk), then the output should never exceed one chunk at a time, meaning you should be able to just process the full thing in one go:
> 
> * if Received 1 right chunk:
>   
>   * For each line in the chunk, check if the key is in the lhs map. If it is store the index in a temp vector.
>   * Once the chunk has been fully checked, build the output chunk
> * Else if received 1 left chunk:
>   
>   * For each line in the chunk, check if the key is in the lhs map. If it is store the index in a temp vector.
>   * Once the chunk has been fully checked, build the output chunk
> 
> It is guaranteed that the output does not exceed 1 chunk size in this case.
> 
> Maybe it doesn't work like that and you wait for two chunks to be available instead of triggering immediately once you receive a single chunk (in a design where it triggers at soon as at least one chunk is available, the first execution doesn't produce an output, or produces an empty output)
@LucLabarriere 
I like the your solution because then we don't have to iterate over everything again, but then don't we potentially have to incur the cost extra cost of the temporary vector everytime?

So the complexity of handling if an input produces more than 1 chunk of output is not too bad to handle, as we just need to restart the cycle from the next row we stopped on. I think alot of the complexity comes from when we have to handle returning the output mid processing of the vector of rows returned by the hash table.

v2/pipeline/processors/HashJoinProcessor.cpp
cyrusknopf: Yeah so I don't understand why it is valid to have `==` in that case? Because the body of the if statement resets `_leftInputIdx`, and if `_leftInputIdx == leftDf->getRowCount()` then is it not the case that we still need to finish processing the final row in the left DF? Or does `_leftInputIdx` point to the index which we have _completed execution_? in that case I don't see why `_leftInputIdx` would ever be greater than `leftDf->getRowCount()`?

ghost: `_leftInputIdx` points to the index of the left input we are looking at currently. For more context 

At the end of the loop over the indices of the input - 
```c++
processLeftStream(...) {
      for(;_leftIndexInputIdx < col.size(); ++_leftInputIdx) {
      ...
      ...
      ...
              // break if we don't have any rows remaining to write
              if (rowsRemaining == 0) {
                  break;
              }
     }

    // if we don't need to finish of copying a row offset vector
    // we can start from the next index
    if (!_copyState.isValid()) {
        _leftInputIdx += 1;
    }
    return
}
```
so if we fully complete the loop since the copyState is invalid we will increment the index again making it more than the row count. I guess I can add and extra condition on the last if condition to check if we are on the last index.

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: I was thinking about avoiding the cost of looking up the join key but it is honestly minimal compared to the cost of copying here, so I will go with that.

v2/pipeline/processors/HashJoinProcessor.cpp
cyrusknopf: Yeah so if copy state is invalid, then you will have `_leftInputIdx = leftDf->getRowCount() + 1`? So then `_leftInputIdx == leftDf->getRowCount()` is not the correct condition?

ghost: If rowsRemaining = 0 we break out of the loop before the last increment so `_leftInputIdx` is infact equal to `leftDf->getRowCount() - 1`. And the `_leftInputIdx == leftDf->getRowCount()` check is done if `rowsRemaining = 0` after we finish `processLeftStream()`

cyrusknopf: Got it, thanks !

storage/RowStore.h
```diff
@@ -0,0 +1,115 @@
...
+};
+
+class RowStore {
+public:
+    RowStore() = default;
+    RowOffset insertRow(const Dataframe* frame, size_t rowNumber);
+
+    // Skip the join column 
+    RowOffset insertRow(const Dataframe* frame,
+                        const ColumnTag& joinColTag,
```
rjb32: Why pass the ColumnTag as a reference, it is designed to be quite light

test/v2/pipeline/processors/HashJoinProcessorTest.cpp
```diff
@@ -0,0 +1,1548 @@
...
+        }
+        _builder->addColumnToOutput<ColumnNodeIDs>(joinKey);
+
+        const auto& hashJoin = _builder->addHashJoin(&rhsIF, joinKey, joinKey);
+        ASSERT_EQ(hashJoin.getDataframe()->cols().size(), LHS_NUM_COLS + RHS_NUM_COLS - 1);
+    }
+
+    constexpr size_t NUM_MATCHES = 0;
+    /*
+     * Output Dataframe should be looking like:
```
rjb32: How can it be if there are no matches?

test/v2/pipeline/processors/HashJoinProcessorTest.cpp
```diff
@@ -0,0 +1,1548 @@
...
+    constexpr size_t LHS_NUM_COLS = 3;
+
+    constexpr size_t RHS_NUM_ROWS = 5;
+    constexpr size_t RHS_NUM_COLS = 3;
+
+    /*
+     * Generate Dataframe looking like:
+     * 11    21    1
+     * 12    22    1
+     * 13    23    2 14    24    2
```
rjb32: I think you want to go to the next line here

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -47,6 +48,45 @@ void concatDataframeShape(LocalMemory* mem,
     }
 }
 
+void createHashJoinDataFrameShape(LocalMemory* mem,
+                                  DataframeManager* dfMan,
+                                  Dataframe* leftSrc,
+                                  Dataframe* rightSrc,
+                                  Dataframe* dest,
+                                  ColumnTag& leftJoinKeyTag,
```
rjb32: No need for reference to tag

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -47,6 +48,45 @@ void concatDataframeShape(LocalMemory* mem,
...
+        }
+
+        Column* newCol = mem->allocSame(col->getColumn());
+        auto* newNamedCol = NamedColumn::create(dfMan, newCol, col->getTag());
+        dest->addColumn(newNamedCol);
+    }
+
+    //allocate a join-key column tag in the output with a new column tag if the left
+    //and right join column tags aren't the same
+    ColumnTag joinColTag = leftJoinKeyTag == rightJoinKeyTag ? leftJoinKeyTag : dfMan->allocTag();
```
rjb32: joinColTag can be const

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,557 @@
...
+                      const size_t offset,
+                      const size_t rowsToCopy) {
+    bioassert(outputColumn->size() >= offset+ rowsToCopy) ;
+    std::fill_n(outputColumn->begin() + offset,
+                rowsToCopy,
+                (*inputColumn)[rowIndex]);
+}
+
+template<typename Key, typename Value>
+auto findInMap(std::unordered_map<Key,Value>& map, Column* col, size_t index) {
```
rjb32: This is a bit ugly to just have a wrapper for find

rjb32:
Looks good to me for now, besides a few stylistic remarks.
Please do a final review @LucLabarriere @cyrusknopf in the light of what we said this Monday afternoon.

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,557 @@
...
+void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
+    _ctxt = ctxt;
+
+    const auto* leftDf = _leftInput.getDataframe();
+    const auto* outDf = _output.getDataframe();
+
+    // The stored rows do not contain the join column
+    _leftRowLen = leftDf->size() - 1;
+    _rightRowLen = outDf->size() - _leftRowLen - 1;
+
```
LucLabarriere, line 92: What's happening with _rightRowLen? why is it outSize - leftSize - 1?

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,557 @@
...
+void HashJoinProcessor::execute() {
+    size_t rowsRemaining = _ctxt->getChunkSize();
+    size_t totalRowsInserted = 0;
+
+    const Dataframe* leftDf = _leftInput.getDataframe();
+    const Dataframe* rightDf = _rightInput.getDataframe();
+
+    Dataframe* outDf = _output.getDataframe();
+
+    if (_copyState.isValid()) {
```
LucLabarriere: Find a better function name for isValid, what are you checking if there are data left to be copied other to the output? Add a comment to explain that stage of the algorithm

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,557 @@
...
+        // resize the output columns to fit the new rows we are inserting
+        const size_t rowsLeft = _copyState._offsetVec->size() - _copyState._rowOffsetIdx;
```
LucLabarriere: rowsLeft means rowsLeftToBeCopied/remainingRows or something like that? It may mean lhs rows which is a bit misleading. Why is it computed as `_copyState._offsetVec->size() - _copyState._rowOffsetIdx`, can you clarify the copyState API to have meaningful functions ? Things like (not sure of the names):
```c++
// Instead of _copyState.isValid()
if (_copyState.hasLeftover())

// And instead of rowsLeft = ...
const size_t remainingRows = _copyState.remainingRowCount();
```

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,557 @@
...
+            dispatchColumnVector(namedCol->getColumn(),
+                                 [&](auto* col) {
+                                 auto* colVector = static_cast<decltype(col)>(col);
+                                 colVector->resize(newOutputSize);
+                                 });
+        }
+
+        if (_copyState._df == rightDf) {
+            finishRightStream(rowsRemaining,
+                              totalRowsInserted);
```
LucLabarriere: `finishRightStream` and `finishLeftStream` are a bit misleading, they suggest the function just switches a finished boolean or something

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,557 @@
...
+            // here we calculate the starting index of the output column that we are
+            // inserting the row into  as _leftRowLen as we we want to copy the left input of
+            // the join into the first columns of the output. 
+            for (size_t k = 0; k < rowsToCopy; ++k) {
+                _store.copyRow(outDf,
+                                    _leftRowLen,
+                                    totalRowsInserted + k,
+                                    _rightRowLen,
+                                    rows[k]);
+            }
```
LucLabarriere, line 314: I'm a bit worried here, basically for each row that we want to copy, we go through each column, do a type dispatch and perform the copy on the element. I would be more efficient to do it column -> type dispatch -> copy rows instead.

v2/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -0,0 +1,557 @@
...
+        if (rowsRemaining == 0) {
+            break;
+        }
+    }
+
+    // if we don't need to finish of copying a row offset vector
+    // we can start from the next index
+    if (!_copyState.isValid() && _leftInputIdx != leftCol->size()) {
+        _leftInputIdx += 1;
+    }
```
LucLabarriere: What do you mean

storage/RowStore.cpp
```diff
@@ -0,0 +1,152 @@
...
+    const auto& cols = frame->cols();
+
+    for (size_t i = 0; i < numCols; ++i) {
+        // do we need a check here
+        const auto& val = cols[i]->as<ColumnVector<NodeID>>()->data()[rowNumber];
+        _dataList.back().setValue<NodeID>(val);
+    }
+
+    return {&currentSlab, offset};
+}
```
LucLabarriere: (not for this PR)
This is heavily row oriented, why is column oriented approach not possible? We could just store the rows in a regular Dataframe, then use a vector of indices to copy over the rows that we want in batches.

```c++
void RowStore::insertRows(const Dataframe* inDf, const ColumnVector<size_t>& indices) {
    // For each column
    // Dispatch the type once
    // For each row index in the vector, copy over the elements to the corresponding output Col
}

void RowStore::copyRows(Dataframe* outDf, const ColumnVector<size_t>* indices) {
    // For each output column to be built
    //   - Dispatch the column type once
    //   - For each index in the list
    //     copy the element from the corresponding input column to the output column
}
```

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: Cool, I wasn't happy with the naming here - I will try renaming.

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: I don't think there is a very clean way - unless we assume that the Map Key value will be equal to the type of the join column. We would need to cast the key value with a type derived from the map in anycase.

storage/RowStore.cpp
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    auto& currentSlab = _dataList.back();
+    const size_t offset = currentSlab.getCurrentOffset();
+
+    const auto& cols = frame->cols();
+
+    for (size_t i = 0; i < numCols; ++i) {
+        // do we need a check here
+        const auto& val = cols[i]->as<ColumnVector<NodeID>>()->data()[rowNumber];
```
cyrusknopf: Is it possible that we want to store EdgeIDs in the row store? Or other types?

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: I will write a comment, but by convention all rows in the leftInput will be processed and rows in the right input will be skipped if they are found to be present in the left row.

test/v2/pipeline/processors/HashJoinProcessorTest.cpp
ghost: needs to be fixed 🙏

storage/RowStore.cpp
cyrusknopf: Might we need to store EdgeIDs? Or other types?

v2/pipeline/processors/HashJoinProcessor.cpp
ghost: Hey Luc I looked into this and I think maybe it is a better idea to do a column oriented store of rows as well (there are some drawbacks as well depending on the pattern). I do agree that we are dispatching more times than necessary - but changing it to go over the columns once -> do a dispatch -> and the copy all the row values for that column seems to throw out all of the benefits of using a row oriented store for now. So maybe this change can come in a PR related to supporting a column oriented row store.

storage/RowStore.cpp
ghost: So as I said in my above column, on more thinking  a column oriented approach could be more beneficial (with caveats!). 
With the store being row oriented :
- we have non contiguous memory access for reading from the input columns into the store
- locality in writing data to the rowStore
- locality in reading data from the rowStore to the output columns
- non contiguous memory access for writing the the rows to the output columns

With the column oriented-store we have:
- contiguous memory access when reading from the input into the store
- non contiguous memory access when writing data into the store
- reading data from the store into the output can be non-contiguous but in the case where we are outputting multiple rows per hash match we would get the benefit of memcpying whole sections of the stores columns
- we get similar benefits for writing to the output columns.
- however the benefits for these disappear if the cardinality of the hash join matches are low.
- only have to the dispatch one time for each column is very nice as well (but this benefit is minimised in cases where the cardinality of the matches is low again)

So looking at it ideally we could be able to switch the backing store used for the hash join processor depending on the cardinality of the match (which ties into the idea of having a join library where we could even have different storage data structures)

v2/pipeline/processors/HashJoinProcessor.h
```diff
@@ -0,0 +1,103 @@
...
+
+    PipelineBlockInputInterface& rightInput() { return _rightInput; }
+    PipelineBlockInputInterface& leftInput() { return _leftInput; }
+    PipelineBlockOutputInterface& output() { return _output; }
+
+    void prepare(ExecutionContext* ctxt) override;
+    void reset() override;
+    void execute() override;
+
+    std::string describe() const final {
```
rjb32: Put describe in the cpp file

storage/RowStore.cpp
ghost: Should be fixed now, that was an old function that should have been deleted!

rjb32:
I think the PR is good as it is, and let's improve in following PRs if needed.

PR #122, opened 2025-11-10
Fix StringIndex returning an output which exceeds a chunk

cyrusknopf:
Context:

```
turing:reactome> match (n{`displayName (String)`~="Trans-autophosphorylation"}) return n, n.`displayName (String)`
[2025-11-10 15:37:54] [info] Column of size: 97751  // <- StringIndexLookup exceeding a chunk
[2025-11-10 15:37:54] [info] Column of size: 65536 // <- Property step within a chunk
[2025-11-10 15:37:54] [info] Column of size: 97751  // <- StringIndexLookupStep repeated
[2025-11-10 15:37:54] [info] Column of size: 32215 // <- Property step remaining after first chunk
Thread 1 "turingdb" received signal SIGSEGV, Segmentation fault.
```
- Columns being differing sizes caused SEGFAULT on turingdb -cli and exception in Python SDK

Changes:
I have added a very quick hotfix which prevents a more meaningful error message to the user. I have an idea for an actual fix (storing all matches which exceed a chunk size in a temporary buffer), but I thought v2 should be the priority for now.

Verified with:
```
turing:reactome> match (n{`displayName (String)`~="Trans-autophosphorylation"}) return n, n.`displayName (String)`
[2025-11-10 16:07:02] [error] EXEC_ERROR: String Index output exceeded CHUNK_SIZE. Please try a more restrictive query.
```

### DIFF
--- a/plan/QueryPlanner.cpp
+++ b/plan/QueryPlanner.cpp
@@ -465,9 +465,9 @@ void QueryPlanner::caseScanNodesStringConstraint(const std::vector<const BinExpr
             // TOMBSTONES: base column @ref firstStageOutput always allocd by
             // @ref planScanNodes
             _pipeline->add<ScanNodesStringApproxStep>(firstStageOutput, _view,
-                                                     propType._id, queryString);
-        break;
+                                                      propType._id, queryString);
         }
+        break;
 
         default:
             throw PlannerException("Unsupported operator for type 'String'");
--- a/pipeline/operations/ScanNodesStringApproxStep.cpp
+++ b/pipeline/operations/ScanNodesStringApproxStep.cpp
@@ -16,7 +16,6 @@
 using namespace db;
 using Step = ScanNodesStringApproxStep;
 
-
 Step::ScanNodesStringApproxStep(ColumnVector<NodeID>* nodes, const GraphView& view,
                                 PropertyTypeID propID, std::string_view strQuery)
     : _nodes(nodes),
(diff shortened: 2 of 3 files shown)


plan/QueryPlanner.cpp
```diff
@@ -465,9 +465,9 @@ void QueryPlanner::caseScanNodesStringConstraint(const std::vector<const BinExpr
             // TOMBSTONES: base column @ref firstStageOutput always allocd by
             // @ref planScanNodes
             _pipeline->add<ScanNodesStringApproxStep>(firstStageOutput, _view,
-                                                     propType._id, queryString);
-        break;
+                                                      propType._id, queryString);
         }
+        break;
```
rjb32, line 470: Good find

rjb32:
I am not sure to understand why generating more than one chunk in StringIndex causes a segfault?

cyrusknopf:
> I am not sure to understand why generating more than one chunk in StringIndex causes a segfault?

The segfault is due to Tabulate because we request creating a table which is not rectangular in shape. In the DB there is no segfault itself, only via the shell

rjb32:
> > I am not sure to understand why generating more than one chunk in StringIndex causes a segfault?
> 
> The segfault is due to Tabulate because we request creating a table which is not rectangular in shape. In the DB there is no segfault itself, only via the shell

But why is this happening?

cyrusknopf:
> > > I am not sure to understand why generating more than one chunk in StringIndex causes a segfault?
> > 
> > 
> > The segfault is due to Tabulate because we request creating a table which is not rectangular in shape. In the DB there is no segfault itself, only via the shell
> 
> But why is this happening?

Because the output of the StringIndex query is not bounded by CHUNK_SIZE, so it can return more than CHUNK_SIZE output. Which causes a mismatch with any other column because no other column can ever be greater than CHUNK_SIZE

PR #124, opened 2025-11-11
Fix rename of column names for nodes and edges

rjb32:
* Centralize the map ColumnTag->std::string_view name into ColumnTagManager because all tags have the same name
* Remove ColumnHeader
* Use a vector indexed by ColumnTag for retrieving the optional name
* Implement a PendingOutputView to allow to have an edge or node view out of a PipelineEdgeOutputInterface and manage renames in PipelineBuilder

Allows correct renaming on node IDs:
`MATCH (n)-->(m) RETURN n`

and on edge IDs:
`MATCH (n)-[e]->(m) RETURN n`

### DIFF
--- a/test/v2/queries/QueriesTest.cpp
+++ b/test/v2/queries/QueriesTest.cpp
@@ -47,10 +47,13 @@ TEST_F(QueriesTest, scanAll) {
         ASSERT_EQ(df->cols().size(), 1);
         ASSERT_EQ(df->size(), 1);
 
-        const ColumnNodeIDs* nodeIDs = df->cols()[0]->as<ColumnNodeIDs>();
-        ASSERT_TRUE(nodeIDs != nullptr);
-        ASSERT_FALSE(nodeIDs->empty());
-        returnedNodeIDs.insert(returnedNodeIDs.end(), nodeIDs->begin(), nodeIDs->end());
+        for (auto col : df->cols()) {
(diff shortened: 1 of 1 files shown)


storage/dataframe/Dataframe.h
```diff
@@ -70,7 +55,7 @@ class Dataframe {
 
 private:
     NamedColumns _cols;
```
LucLabarriere, line 57: I'm not sure why we can't have simply:

```c++
std::vector<NamedColumn*> _cols;
```

and do a linear search

```c++
for (const auto* col : _cols) {
    if (col->getTag() == requestedTag) {
        return col;
    }
}

return nullptr;
```

I think overall the gains of having O(1) accesses for columns are really not worth the cost of everything else (DataFrames are really really heavy with this design (48 bytes if I'm not mistaken + the content of the vectors, especially DynamicLookupTable can get really big)

rjb32: Because this is a linear search, this is O(n) in general, this is generally bad. This is a good reflex to have not to do. This may appear ok now, but this is hardly not future proof and not to be done, because what if we forget that it must remain small? This can also be misused by future contributors of different levels of attention, awareness of the reasons, skills..etc.
The 48 bytes for me are ok. Anyway it is to be compared with the space taken by the unordered_maps which are much more costly in terms of space just to keep a low occupancy factor.

PR #126, opened 2025-11-12
GetProperty nodes support in the PlanGraph

LucLabarriere:
- GetPropertyNode to retrieve properties + filter
- GetPropertyNodeWithNull to retrieve properties for the return stmt
- GetEntityType to retrieve node labelset / edge type
- Caching of get properties and get entity types


LucLabarriere:
@rjb32 Added a GetPropertyWithNull node for the return stmt (that was not handled properly)

v2/plan/ReadStmtGenerator.cpp
```diff
@@ -559,7 +580,7 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, VarNode* dependency) {
         case PlanGraphTopology::PathToDependency::BackwardPath: {
             // If only walked backward
             // Materialize
-            _tree->insertBefore<MaterializeNode>(filter);
+            // _tree->insertBefore<MaterializeNode>(filter);
```
cyrusknopf: Leftover?

LucLabarriere: Good point thanks!

v2/plan/GetEntityTypeCache.h
```diff
@@ -0,0 +1,19 @@
...
+
+#include <unordered_set>
+
+namespace db::v2 {
+
+class VarDecl;
+
+class GetEntityTypeCache {
+public:
+    bool insert(const VarDecl* var) {
```
rjb32, line 22: It's a bit confusing when we don't see the usage context to understand why there is no lookup method and that it's given by the result of insert. Maybe a comment?

v2/plan/ReadStmtGenerator.cpp
LucLabarriere: @rjb32 Can I safely remove the MaterializeNode from the PlanGraph? I think we agreed that the Materialize should be taken care of automatically by the PipelineGenerator (or the PipelineBuilder)

v2/plan/GetEntityTypeCache.h
LucLabarriere: I documented the classes, is it better?

rjb32: Yes thanks

PR #127, opened 2025-11-13
Projection Processor in pipeline v2

LucLabarriere:


v2/pipeline/PipelineBuilder.cpp
```diff
@@ -191,6 +192,26 @@ PipelineValueOutputInterface& PipelineBuilder::addCount() {
...
+    PipelineBlockInputInterface& input = projection->input();
+    PipelineBlockOutputInterface& output = projection->output();
+
+    _pendingOutput.connectTo(input);
+
+    Dataframe* inDf = input.getDataframe();
+    Dataframe* outDf = output.getDataframe();
+
+    for (const ColumnTag& tag : tags) {
+        outDf->addColumn(inDf->getColumn(tag));
```
cyrusknopf, line 208: Doesn't this make the output DF have pointers to the columns of the input DF? Is this intended?

LucLabarriere: Yes that's exactly what it does, the Projection does not do anything in the executor, it only "forwards" the columns that the user requested in the RETURN. So the next node (the lambda in this case), receives only the columns that were projected as input

cyrusknopf: Ah of course, sounds good. Do we expect other processors to do a similar "forwarding"? If so maybe it would beneficial to wrap it in a "forwardInput" function to make it explicit, just as we have a "copyDataframeShape" function

LucLabarriere: You mean like this?:

```c++
for (const ColumnTag& tag : tags) {
    forwardInput(outDf, inDf, tag);
}
```

ghost: maybe like 
```c++
void forwardInput(outDf,inDf,tags) {
    for(const ColumnTag& tag : tags) {
        outDf->addColumn(inDf->getColumn(tag));
    }
}
```

LucLabarriere: Hm I'm not sure, it means we have to find the function to know what it does. Would comments do the trick? Else I'll add the function. (I thought of it for a bit and I don't think other nodes will have to do that, maybe when we have the WITH statements but not sure)

cyrusknopf: If this is the only processor that forwards like this then maybe we dont need the function. But whether we add a function or not comments are needed IMO to make it explicit that we are reusing the pointers of the previous input - otherwise it is easy to miss/misinterpret what is being done.

PR #128, opened 2025-11-13
Implement Filters in V2

rjb32:
* EvalExprProcessor to evaluate expressions already translated into 3 operands code
* Reuse v1 Filter implementation based on ColumnOperators

### DIFF
--- a/v2/plan/ExprProgramGenerator.cpp
+++ b/v2/plan/ExprProgramGenerator.cpp
@@ -48,8 +48,8 @@ ColumnOperator ExprProgramGenerator::unaryOperatorToColumnOperator(UnaryOperator
         break;
 
         default:
-            throw PlannerException(fmt::format("ExprProgramGenerator: unary expression "
-                                               "operator of kind {} not supported",
+            throw PlannerException(fmt::format("Unary operator "
+                                               "{} not yet supported.",
                                                UnaryOperatorDescription::value(op)));
         break;
     }
@@ -70,8 +70,7 @@ ColumnOperator ExprProgramGenerator::binaryOperatorToColumnOperator(BinaryOperat
         break;
 
         default:
-            throw PlannerException(fmt::format("ExprProgramGenerator: binary expression "
-                                               "operator of kind {} not supported",
+            throw PlannerException(fmt::format("Binary operator {} not yet supported.",
                                                BinaryOperatorDescription::value(op)));
         break;
     }


storage/columns/ColumnOperators.h
```diff
@@ -416,6 +443,68 @@ class ColumnOperators {
             }
         }
     }
+
+private:
+    template <BooleanOpt T, BooleanOpt U>
+    static std::optional<bool> optionalOr(const T& a, const U& b) {
+        if (a == CustomBool {true} || b == CustomBool {true}) {
```
rjb32, line 450: It would be better to do one if for checking if a or b is nullopt and then return std::nullopt, and then return the OR on CustomBool. To avoid 2 if statements

storage/columns/ColumnOperators.h
```diff
@@ -416,6 +443,68 @@ class ColumnOperators {
...
+        }
+        if (a == CustomBool {false} && b == CustomBool {false}) {
+            return false;
+        }
+        return std::nullopt;
+    }
+
+    template <BooleanOpt T, BooleanOpt U>
+    static std::optional<bool> optionalAnd(const T& a, const U& b) {
+        if (a == CustomBool {true} && b == CustomBool {true}) {
```
rjb32, line 461: Same here

storage/columns/ColumnOperators.h
```diff
@@ -416,6 +443,68 @@ class ColumnOperators {
...
+        }
+        if (a == CustomBool {false} || b == CustomBool {false}) {
+            return false;
+        }
+        return std::nullopt;
+    }
+
+    template <BooleanOpt T>
+    static std::optional<bool> optionalNot(const T& a) {
+        if (a == CustomBool {true}) {
```
rjb32, line 472: Same here

storage/columns/ColumnVector.h
```diff
@@ -50,6 +50,12 @@ class ColumnVector : public Column {
     {
     }
 
+    explicit ColumnVector(size_t size, const T& value)
+        : Column(_staticKind),
+          _data(size, value)
```
rjb32: Align _data with the ":" character

v2/plan/ExprProgramGenerator.cpp
```diff
@@ -1,5 +1,6 @@
 #include "ExprProgramGenerator.h"
 
+#include <iostream>
```
rjb32: Remove iostream

PR #129, opened 2025-11-13
Support for the projection in v2 + VarDecl on expressions + fixed skip/limit bug

LucLabarriere:
- This adds basic support for the ProduceResults node. Does not support the wildcard for now (*), can only return nodes and edges
- Fixed skip/limit bug
- Fixed head bug in TuringShell
- Added VarDecl to the root for expressions
- Added a VarDecl -> ColumnTag mapping in PipelineGenerator. Using it in translateVarNode.

### DIFF
--- a/v2/plan/nodes/ProduceResultsNode.h
+++ b/v2/plan/nodes/ProduceResultsNode.h
@@ -14,15 +14,15 @@ public:
     }
 
     const Projection* getProjection() const {
-        return m_proj;
+        return _proj;
     }
 
     void setProjection(const Projection* proj) {
-        m_proj = proj;
+        _proj = proj;
     }
 
 private:
-    const Projection* m_proj {nullptr};
+    const Projection* _proj {nullptr};
 };
 
 }


v2/pipeline/PipelineExecutor.cpp
```diff
@@ -9,7 +9,7 @@ using namespace db::v2;
 PipelineExecutor::PipelineExecutor(PipelineV2* pipeline,
                                    ExecutionContext* ctxt)
     : _pipeline(pipeline),
-      _ctxt(ctxt)
+    _ctxt(ctxt)
```
rjb32, line 12: Is ExecutionContext actually used anywhere inside PipelineExecutor?

LucLabarriere: Yes it is passed to the processors in the prepare method

v2/plan/nodes/ProduceResultsNode.h
```diff
@@ -4,12 +4,25 @@
...
+    const Projection* getProjection() const {
+        return m_proj;
+    }
+
+    void setProjection(const Projection* proj) {
+        m_proj = proj;
+    }
+
+private:
+    const Projection* m_proj {nullptr};
```
cyrusknopf: Why `m_proj` instead of `_proj`?

LucLabarriere: Oups ahah

PR #132, opened 2025-11-17
Add The Ability To Allow Processors To Execute With Only Some Ports ReadyWithData

ghost:


v2/pipeline/PipelinePort.h
```diff
@@ -54,8 +62,9 @@ class PipelinePort {
     PipelinePort* _connectedPort {nullptr};
     PipelineBuffer* _buffer {nullptr};
     bool _open {true};
+    bool _needsData {true};
 
-    PipelinePort(Processor* processor)
+    explicit PipelinePort(Processor* processor)
```
cyrusknopf, line 67: Why did you want to mark it as explicit?

rjb32:
Looks good to me!

v2/pipeline/PipelinePort.h
ghost: It's generally considered good practice in c++, to prevent bugs from implicit conversions.

PR #134, opened 2025-11-17
Fix LambdaSourceProcessor, CountProcessor and PipelineExecutor

rjb32:
* Make LambdaSourceProcessor writeData after the callback is executed
* Remove resetting _countRunning to zero in CountProcessor::reset()
* Fix count processor
* Fix PipelineExecutor by using an activeStack instead of an activeQueue

### DIFF
--- a/v2/pipeline/PipelineBuilder.h
+++ b/v2/pipeline/PipelineBuilder.h
@@ -46,9 +46,6 @@ public:
     PipelineEdgeOutputInterface& addGetEdges();
     void projectEdgesOnOtherIDs() { _pendingOutput.projectEdgesOnOtherIDs(); }
 
-    // Lambda transform
-    PipelineBlockOutputInterface& addLambdaTransform(const LambdaTransformProcessor::Callback& cb);
-
     // Properties
     template <SupportedType T>
     PipelineValuesOutputInterface& addGetNodeProperties(PropertyType propertyType);


rjb32:
To show that it is failing in main: https://github.com/turing-db/turingdb/pull/137

It needs as well LambdaSourceProcessor fixed in main, which goes into an infinite loop without the executor fix.

rjb32:
> Seems like it should be working. If you want I can try to retrieve the changes in my projection branch and see if it works 

Yes please test :)

LucLabarriere:
> > Seems like it should be working. If you want I can try to retrieve the changes in my projection branch and see if it works
> 
> Yes please test :)

Tested on my GetProperties branch (which also has the scheduler bug), it fixes the issues! My pipeline is:

```
ScanNodes -> GetProperties -> Materialize -> GetOutEdges -> Materialize -> Lambda
```

will test on the projection now but should be good

LucLabarriere:
> > > Seems like it should be working. If you want I can try to retrieve the changes in my projection branch and see if it works
> > 
> > 
> > Yes please test :)
> 
> Tested on my GetProperties branch (which also has the scheduler bug), it fixes the issues! My pipeline is:
> 
> ```
> ScanNodes -> GetProperties -> Materialize -> GetOutEdges -> Materialize -> Lambda
> ```
> 
> will test on the projection now but should be good

Seems to work for the projection too!

LucLabarriere:
Quick question, we now have addLambda, addLambdaTransform and addLambdaSource. What are the differences between them?

rjb32:
> Quick question, we now have addLambda, addLambdaTransform and addLambdaSource. What are the differences between them?

It appears that we need 3 ways to inject arbitrary code or data:

* Lambda (sink): this is the classic lambda processor to execute a callback at the end of the pipeline. For encoding a dataframe or in tests to extract data to perform checks

* Lambda source: we need a way in tests to be able to generate arbitrary chunks of data

* Lambda transform: to make an arbitrary transform processor from dataframe to dataframe. Useful in tests to mimick a GetOutEdges-like processor emitting a custom number of chunks and to test "finish" behavior and pipeline scheduling

LucLabarriere:
> > Quick question, we now have addLambda, addLambdaTransform and addLambdaSource. What are the differences between them?
> 
> It appears that we need 3 ways to inject arbitrary code or data:
> 
> * Lambda (sink): this is the classic lambda processor to execute a callback at the end of the pipeline. For encoding a dataframe or in tests to extract data to perform checks
> * Lambda source: we need a way in tests to be able to generate arbitrary chunks of data
> * Lambda transform: to make an arbitrary transform processor from dataframe to dataframe. Useful in tests to mimick a GetOutEdges-like processor emitting a custom number of chunks and to test "finish" behavior and pipeline scheduling

ok thanks

PR #142, opened 2025-11-19
GetProperties + GetPropertiesWithNull in PipelineGenerator

LucLabarriere:


samples/v2/main.cpp
```diff
@@ -113,7 +111,7 @@ int main(int argc, char** argv) {
         {
             auto callback = [](const Dataframe* dataframe) {};
 
-            PipelineGenerator pipelineGen(&planGraph, &pipeline, &mem, callback);
+            PipelineGenerator pipelineGen(&planGraph, view, &pipeline, &mem, ast.getSourceManager(), callback);
```
LucLabarriere, line 114: `view` is used to retrieve metadata (property types, etc.). Source manager is used to set the column names for expressions (`RETURN n.name`, and later: `RETURN n.age > 5` etc)

storage/dataframe/ColumnTag.h
```diff
@@ -1,19 +1,22 @@
...
 
+#include <limits>
 #include <stddef.h>
+#include <stdint.h>
 
 namespace db {
 
 class ColumnTag {
 public:
+    using ValueType = uint32_t;
```
LucLabarriere, line 11: @rjb32 is it fine to make it `uint32_t` instead of `uint64_t` ? It reduces the size of the `EntityOutputStream` class

test/v2/pipeline/PipelineTest.cpp
```diff
@@ -210,72 +171,6 @@ TEST_F(PipelineTest, scanNodesExpand1) {
...
-TEST_F(PipelineTest, scanNodesExpandGetProperties) {
-    LocalMemory mem;
-    PipelineV2 pipeline;
-
-    PipelineBuilder builder(&mem, &pipeline);
-    
-    PipelineNodeOutputInterface& scanNodesOut = builder.addScanNodes();
-    const ColumnTag scanOutNodeIDsTag = scanNodesOut.getNodeIDs()->getTag();
-
-    builder.addGetOutEdges();
```
LucLabarriere, line 222: Basically moved to GetPropertiesProcessorTest

v2/AST/SourceManager.cpp
```diff
@@ -11,11 +36,33 @@ SourceManager::~SourceManager() {
...
+    if (firstLineNo != lastLineNo) {
+        ignoreLines(repr, lastLineNo - firstLineNo);
+        nextLine(repr);
+    } else {
+        lastCol = loc->endColumn - loc->beginColumn;
+    }
+
+    repr = repr.substr(0, lastCol);
+    return repr;
 }
```
LucLabarriere, line 66: This is now used to retrieve a string representation of expressions. This might be a bit expensive and we might want to have the option of disabling it. It's nice in a cli mode where the column headers are outputted nicely. For json output and the sdk it's also nice, it can allow to do that:

```python
result = client.query('MATCH (n) RETURN n.name, n.age > 20')
names = result["n.name"]
condition = result["n.age > 2"]
```

v2/pipeline/EntityOutputStream.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include <array>
+
+#include "dataframe/ColumnTag.h"
+
+namespace db::v2 {
+
+class EntityOutputStream {
```
LucLabarriere, line 9: Imo, this is a temporary fix to the Stream issue right now, basically the idea is that we can now do `ScanNodes` -> `GetProperty` -> `GetOutEdges`. The problem is that `GetOutEdges` follows a `GetProperty` which does not have a 'Node' output interface necessary for the `GetOutEdges` to work. The solution right now is to store an object of this class in the BlocK/Values Input/Output pipeline interfaces so that we keep track of what was the stream of entities that we were working on. This way, although `GetProperty` outputs a `ValuesInterface`, the stream is still a node stream.

v2/pipeline/PendingOutputView.h
```diff
@@ -2,7 +2,10 @@
 
 #include <string_view>
 
-#include "PipelineInterface.h"
+#include "interfaces/PipelineBlockInputInterface.h"
+#include "interfaces/PipelineEdgeInputInterface.h"
+#include "interfaces/PipelineEdgeOutputInterface.h"
+#include "interfaces/PipelineNodeInputInterface.h"
```
LucLabarriere, line 8: Split the interfaces into their own files because they were getting quite complex

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -69,21 +70,31 @@ PipelineNodeOutputInterface& PipelineBuilder::addScanNodes() {
 
 PipelineEdgeOutputInterface& PipelineBuilder::addGetOutEdges() {
     // Create get out edges processor
+    if (!_isMaterializeOpen) {
+        bioassert(_matProc != nullptr);
+        _matProc = MaterializeProcessor::createFromPrev(_pipeline, _mem, *_matProc);
+        _isMaterializeOpen = true;
+    }
+
```
LucLabarriere, line 78: It is now not so trivial anymore to know if we should put a materialize or not. For example:
```
ScanNodes -> GetProperty -> GetOutEdges
```
The `GetProperty` "filters" implicitly the nodes coming from the `ScanNodes`, meaning that we have to put a `Materialize` right before the `GetOutEdges`. Moreover, we need to "propagate" the columns from the previous `Materialize` to the next one. This is what the createFromPrev does.

The `closeMaterialize` method is not necessary anymore, if there is a pending materialize at the end that does not actually do anything, it is simply ignored and not added to the pipeline.

Propagating columns here means:
- If we have columns A B C after the materialize
- and columns D E F after the second materialize

We must also apply the transformation of the second materialize to the columns produced by the first one. Propagating columns means take all the columns of the previous materialize and add them as an initial step to the next one (as if they were coming from a ScanNodes or ScanOutEdges)

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -69,21 +70,31 @@ PipelineNodeOutputInterface& PipelineBuilder::addScanNodes() {
...
     GetOutEdgesProcessor* getOutEdges = GetOutEdgesProcessor::create(_pipeline);
-    _pendingOutput.connectTo(getOutEdges->inNodeIDs());
 
-    PipelineEdgeOutputInterface& outEdges = getOutEdges->outEdges();
-    Dataframe* outDf = outEdges.getDataframe();
+    PipelineNodeInputInterface& input = getOutEdges->inNodeIDs();
+    PipelineEdgeOutputInterface& output = getOutEdges->outEdges();
+
+    _pendingOutput.connectTo(input);
+    input.propagateColumns(output);
```
LucLabarriere, line 85: We also need the propagation of columns in certain cases. The reason is that we may have a chain of GetPropertiesWithNull (which don't have indices and thus don't require to materialize after), but we need to be able to forward all the columns to the projection.

```
ScanNodes -> GetPropertyWithNull(name) -> GetPropertyWithNull(age) -> Projection
```

Once we get to the Projection, we need to have access to all columns, meaning we need to them to be propagated

storage/dataframe/ColumnTag.h
```diff
@@ -23,16 +26,26 @@ class ColumnTag {
...
 
-    size_t getValue() const { return _value; }
+    ValueType getValue() const { return _value; }
 
     bool operator==(const ColumnTag& other) const { return _value == other._value; }
     bool operator!=(const ColumnTag& other) const { return _value != other._value; }
 
-    ColumnTag& operator++() { _value++; return *this; }
-    ColumnTag operator++(int) { ColumnTag old = *this; _value++; return old; }
+    ColumnTag& operator++() {
```
rjb32: What does it mean to be able to do ++ on a ColumnTag?

v2/AST/SourceManager.h
```diff
@@ -25,18 +25,13 @@ class SourceManager {
         setLocation((uintptr_t)obj, loc);
     }
 
-    void setDebugLocations(bool debugLocations) {
-        _debugLocations = debugLocations;
-    }
-
     const SourceLocation* getLocation(uintptr_t obj) const;
+    std::string_view getStringRepr(uintptr_t obj) const;
```
rjb32: Please add a comment on the purpose of this function

v2/analyzer/ExprAnalyzer.cpp
```diff
@@ -357,10 +357,14 @@ db::ValueType ExprAnalyzer::analyze(PropertyExpr* expr, bool allowCreate, ValueT
         } break;
     }
 
-    expr->setDecl(varDecl);
+    expr->setEntityVarDecl(varDecl);
     expr->setType(type);
     expr->setDynamic();
 
+    // Create a variable declaration for the property expression
```
rjb32, line 364: Very useful thank you :)

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -125,75 +136,82 @@ PipelineBlockOutputInterface& PipelineBuilder::addCartesianProduct(PipelineOutpu
...
 PipelineBlockOutputInterface& PipelineBuilder::addMaterialize() {
-    _pendingOutput.connectTo(_matProc->input());
-    _pendingOutput.setInterface(&_matProc->output());
-
+    auto& input = _matProc->input();
     auto& output = _matProc->output();
 
-    closeMaterialize();
+    _pendingOutput.connectTo(input);
+    output.setStream(input.getStream());
```
LucLabarriere, line 211: In general, this is how we forward the stream. @suhassn-turing-bio @cyrusknopf I assume the join/cartesian product we also need to forward the stream like this, we need to be able to do:

```
CartesianProduct -> Filter -> GetOutEdges
```

If the cartesian product comes from a Node stream, GetOutEdges needs to know it

v2/pipeline/PipelineBuilder.h
```diff
@@ -70,9 +100,8 @@ class PipelineBuilder {
 
     // Materialize
     PipelineBlockOutputInterface& addMaterialize();
-    bool isMaterializeOpen() const { return _matProc != nullptr; }
+    bool isMaterializeOpen() const { return _isMaterializeOpen; }
```
rjb32, line 103: This is creating an additional bit of state to maintain & keep correct compared to just using the matProc pointer. Is it really semantically necessary?

v2/pipeline/interfaces/PipelineInterface.cpp
```diff
@@ -0,0 +1,187 @@
...
+    throw PipelineException(fmt::format("{}: cannot connect to block input", PipelineInterfaceKindName::value(getKind())));
+}
+
+void PipelineOutputInterface::connectTo(PipelineEdgeInputInterface& input) {
+    throw PipelineException(fmt::format("{}: cannot connect to edge input", PipelineInterfaceKindName::value(getKind())));
+}
+
+void PipelineOutputInterface::rename(const std::string_view& name) {
+    throw PipelineException(fmt::format("{}: cannot rename", PipelineInterfaceKindName::value(getKind())));
+}
```
LucLabarriere, line 38: Added default implementations that throw if not overridden in children

v2/pipeline/interfaces/PipelineEdgeOutputInterface.h
```diff
@@ -0,0 +1,53 @@
...
+
+    void setEdges(NamedColumn* edgeIDs,
+                  NamedColumn* edgeTypes,
+                  NamedColumn* otherNodes) {
+        _edgeIDs = edgeIDs;
+        _edgeTypes = edgeTypes;
+        _otherNodes = otherNodes;
+    }
+
+    void setIndices(NamedColumn* indices) { _indices = indices; }
```
rjb32, line 32: It's interesting that the indices are now part of the interface

v2/pipeline/interfaces/PipelineInterface.cpp
```diff
@@ -0,0 +1,187 @@
...
+    Dataframe* outDf = output.getDataframe();
+
+    for (const auto& col : inDf->cols()) {
+        if (outDf->getColumn(col->getTag())) {
+            continue;
+        }
+
+        outDf->addColumn(col);
+    }
+}
```
LucLabarriere, line 51: The propagation of columns needs to happen from input(ProcA) -> output(ProcA). For example, a GetOutEdges has a NodeInputInterface and produce an EdgeOutputInterface. We need to forward the source ID column to the output so we do nodeInput.propagateColumns(edgeOutput).

I originally thought we had to do this between two processor interfaces output(ProcA) -> input(ProcB) but no

v2/pipeline/interfaces/PipelineInterface.cpp
```diff
@@ -0,0 +1,187 @@
...
+    _port->connectTo(input.getPort());
+}
+
+// Block output
+void PipelineBlockOutputInterface::connectTo(PipelineBlockInputInterface& input) {
+    _port->connectTo(input.getPort());
+    input.setStream(_stream);
+}
+
+void PipelineBlockOutputInterface::connectTo(PipelineNodeInputInterface& input) {
```
rjb32, line 108: Looks practical thanks

v2/pipeline/processors/GetPropertiesProcessor.cpp
```diff
@@ -2,47 +2,106 @@
...
+        }
+        ids = dynamic_cast<const ColumnNodeIDs*>(inDf->getColumn(idsTag)->getColumn());
     } else {
-        idTypeStr = "Edge";
+        const ColumnTag idsTag = stream.getEdgeIDsTag();
+        if (!idsTag.isValid()) {
+            throw PipelineException(fmt::format("GetPropertiesProcessor<Edge, {}> must act on an Edge stream",
+                                                ValueTypeName::value(T::_valueType)));
+        }
+        ids = dynamic_cast<const ColumnEdgeIDs*>(inDf->getColumn(idsTag)->getColumn());
```
LucLabarriere, line 60: In the current impl, the `EntityOutputStream` stores `ColumnTags`, so we retrieve the actual columns in the `::prepare`. In might be better to store the actual columns themselves but I wasn't sure this was going to work

v2/pipeline/processors/MaterializeData.cpp
```diff
@@ -14,11 +13,25 @@ using namespace db;
 
 MaterializeData::MaterializeData(LocalMemory* mem, DataframeManager* dfMan)
     : _mem(mem),
-    _dfMan(dfMan),
-    _columnsPerStep(1)
+      _dfMan(dfMan),
```
rjb32: Fix indent

v2/pipeline/processors/MaterializeData.cpp
```diff
@@ -14,11 +13,25 @@ using namespace db;
...
+    _colCount = prevData._colCount;
+
+    for (const auto& col : prevData._output->cols()) {
+        Column* newCol = mem->allocSame(col->getColumn());
+        auto* newNamedCol = NamedColumn::create(dfMan, newCol, col->getTag());
+
+        _columnsPerStep[0].push_back(col->getColumn());
+        _output->addColumn(newNamedCol);
+    }
+}
```
LucLabarriere, line 33: All the columns produced by the previous materialize are added to the initial step of the new materialize (as if they were produced by a ScanNodes)

v2/plan/PipelineGenerator.h
```diff
@@ -32,12 +35,16 @@ class CartesianProductNode;
 class PipelineGenerator {
 public:
     PipelineGenerator(const PlanGraph* graph,
+                      const GraphView& view,
                       PipelineV2* pipeline,
                       LocalMemory* mem,
+                      SourceManager* sourceManager,
```
rjb32, line 41: Why sourceManager is necessary here?

rjb32:
While you are at it on pipeline interfaces, I will need a PipelineValuesInputInterface for the filter processor

v2/plan/PipelineGenerator.cpp
```diff
@@ -197,6 +229,65 @@ void PipelineGenerator::translateGetEdgeTargetNode(GetEdgeTargetNode* node) {
...
+    }
+
+    const std::string_view exprStrRep = _sourceManager->getStringRepr((std::uintptr_t)expr);
+    if (exprStrRep.empty()) {
+        throw PlannerException("Unknown error. Could not get string representation of expression");
+    }
+
+    output->rename(exprStrRep);
+
+    _declToColumn[exprDecl] = output->getValues()->getTag();
```
LucLabarriere, line 288: Set the name of the column to match the query string, and map the expr decl of the property expression to the column. @rjb32 you'll be able to retrieve the columns for the filter step from this map

v2/pipeline/PipelineBuilder.h
LucLabarriere: I think we can get rid of it completely, just checking isSingleStep should be enough, ew can't compare anymore to nullptr though because there is always a materialize ready to be built now (because we never know when we might need one)

storage/dataframe/ColumnTag.h
LucLabarriere: It was your code actually, I just incorrectly formatted it. I reverted it back to how it was. I think you use it to increment the column tag when you allocate a new one

v2/plan/PipelineGenerator.h
LucLabarriere: One of the things I'm not sure of, I use it to get a representation of the expression to set the column name. In `RETURN n.age > 5`, the `Expr*` object is mapped to location in the query string. We can retrieve that location to build the representation of the expression so that the column name is "n.age > 5"

LucLabarriere:
> While you are at it on pipeline interfaces, I will need a PipelineValuesInputInterface for the filter processor

Are you sure you shouldn't be working on a Block in general? Because we could more complex expressions like WHERE n == m which compare entity IDs together and not just values.

storage/dataframe/ColumnTag.h
rjb32: Yes it's enough

v2/AST/SourceManager.cpp
rjb32: Ah yes indeed agree!

v2/pipeline/EntityOutputStream.h
rjb32: I like it as a solution, I don't think it is that temporary. It makes sense. We will see of course in the future if we need to refine the idea

LucLabarriere:
@suhassn-turing-bio @cyrusknopf

In `PipelineGenerator::generator()` we have:

```c++
// ...
for (PlanGraphNode* nextNode : node->outputs()) {
    if (!nextNode->isBinary()) {
        nodeStack.push(nextNode);
        continue;
    }
 
    // Binary node case
    // Cartesian Product and Join require their inputs to be materialised
    if (_builder.isMaterializeOpen()) {
        throw PipelineException("Attempted to add input to binary node "
                                "whilst Materialize is open.");
    }

    const bool visited = _binaryVisitedMap.contains(nextNode);
    if (visited) {
        nodeStack.push(nextNode);
        continue;
    }
    // ...
```

Now, it's much harder than before to know when we should add a `Materialize` processor, meaning that requirement `_builder.isMaterializeOpen() throw` is hard to pass. After this PR, I think a solution instead would be to do:

```c++
// Binary node case
// Cartesian Product and Join require their inputs to be materialised
if (_builder.isMaterializeOpen()) {
    _builder.addMaterialize();
}
```

I checked quickly and it seems to work, but that needs testing

v2/pipeline/interfaces/PipelineInterface.cpp
rjb32: Thanks it's more concise

v2/plan/PipelineGenerator.cpp
rjb32: Great thanks

cyrusknopf:
> @suhassn-turing-bio @cyrusknopf
> 
> In `PipelineGenerator::generator()` we have:
> 
> ```c++
> // ...
> for (PlanGraphNode* nextNode : node->outputs()) {
>     if (!nextNode->isBinary()) {
>         nodeStack.push(nextNode);
>         continue;
>     }
>  
>     // Binary node case
>     // Cartesian Product and Join require their inputs to be materialised
>     if (_builder.isMaterializeOpen()) {
>         throw PipelineException("Attempted to add input to binary node "
>                                 "whilst Materialize is open.");
>     }
> 
>     const bool visited = _binaryVisitedMap.contains(nextNode);
>     if (visited) {
>         nodeStack.push(nextNode);
>         continue;
>     }
>     // ...
> ```
> 
> Now, it's much harder than before to know when we should add a `Materialize` processor, meaning that requirement `_builder.isMaterializeOpen() throw` is hard to pass. After this PR, I think a solution instead would be to do:
> 
> ```c++
> // Binary node case
> // Cartesian Product and Join require their inputs to be materialised
> if (_builder.isMaterializeOpen()) {
>     _builder.addMaterialize();
> }
> ```
> 
> I checked quickly and it seems to work, but that needs testing

Yep in my PR I actually have


```
if (_builder.isMaterializeOpen() && !_builder.isSingleStep()) {
    _builder.addMaterialize();
}
```

storage/dataframe/ColumnTag.h
rjb32: Ah ok 😅

rjb32:
> > While you are at it on pipeline interfaces, I will need a PipelineValuesInputInterface for the filter processor
> 
> Are you sure you shouldn't be working on a Block in general? Because we could more complex expressions like WHERE n == m which compare entity IDs together and not just values.

I mean for the ComputeExprProcessor I do before a filter in general, it creates a column of values as output that the filter processor will take as input

PR #144, opened 2025-11-19
Implement the WriteProcessor for v2

cyrusknopf:
NOTE: This currently does not support creating nodes/edges with properties.
Once we have properties populated in value columns, it should be trivial to add.

### DIFF
--- a/v2/pipeline/processors/WriteProcessor.h
+++ b/v2/pipeline/processors/WriteProcessor.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <cmath>
+#include <math.h>
 #include <optional>
 #include <string_view>
 
--- a/v2/pipeline/processors/WriteProcessorTypes.h
+++ b/v2/pipeline/processors/WriteProcessorTypes.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <cstddef>
+#include <stddef.h>
 #include <string_view>
 
 #include "dataframe/ColumnTag.h"
--- a/v2/plan/PipelineGenerator.cpp
+++ b/v2/plan/PipelineGenerator.cpp
@@ -103,7 +103,9 @@ struct PropertyTypeDispatcher {
 
 }
 
-db::ColumnTag PipelineGenerator::getCol(const VarDecl* var) {
+namespace db {
+
+ColumnTag PipelineGenerator::getCol(const VarDecl* var) {
     if (!var) {
         throw FatalException("Attempted to get column for null variable");
     }
@@ -112,8 +114,10 @@ db::ColumnTag PipelineGenerator::getCol(const VarDecl* var) {
     if (it == end(_declToColumn)) {
         throw PlannerException(fmt::format("Failed to find column for variable {}.", var->getName()));
     }
-    ColumnTag tag = it->second;
+    const ColumnTag tag = it->second;
     return tag;
+}
+
 }
 
 void PipelineGenerator::generate() {
--- a/test/v2/queries/WriteQueriesTest.cpp
+++ b/test/v2/queries/WriteQueriesTest.cpp
@@ -66,17 +66,16 @@ TEST_F(WriteQueriesTest, scanNodesCreateNode) {
     constexpr std::string_view CREATE_QUERY = "MATCH (n) CREATE (m:NEWNODE) RETURN n, m";
     constexpr std::string_view MATCH_QUERY = "MATCH (n) RETURN n";
 
+    const size_t numNodesPrior = read().getTotalNodesAllocated();
+
     { // CREATE query execution and ensure correct DF is returned
         using Rows = LineContainer<NodeID, NodeID>;
         Rows expectedRows;
         {
-            auto transaction = _graph->openTransaction();
-            auto reader = transaction.readGraph();
-            const size_t numNodes = reader.getTotalNodesAllocated();
             // For each existing node we create a new node. New nodes start from current
             // max node ID
-            for (const NodeID n : reader.scanNodes()) {
-                expectedRows.add({n, n + numNodes});
+            for (const NodeID n : read().scanNodes()) {
+                expectedRows.add({n, n + numNodesPrior});
             }
         }
 
@@ -107,8 +106,8 @@ TEST_F(WriteQueriesTest, scanNodesCreateNode) {
 
         Rows expectedRows;
         { // We should now have 26 nodes
-            constexpr size_t EXP_NUM_NODES = 26;
-            for (size_t i = 0; i < EXP_NUM_NODES; i++) {
+            const size_t expectedNumNodes = numNodesPrior * 2;
+            for (size_t i = 0; i < expectedNumNodes; i++) {
                 expectedRows.add({i});
             }
         }
@@ -144,18 +143,16 @@ TEST_F(WriteQueriesTest, scanNodesCreateNode) {
 TEST_F(WriteQueriesTest, scanNodesCreateNodes) {
     constexpr std::string_view CREATE_QUERY = "MATCH (n) CREATE (m:NEWNODE), (p:NEWERNODE) RETURN n, m, p";
     constexpr std::string_view MATCH_QUERY = "MATCH (n) RETURN n";
+    const size_t numNodesPrior = read().getTotalNodesAllocated();
 
     { // CREATE query execution and ensure correct DF is returned
         using Rows = LineContainer<NodeID, NodeID, NodeID>;
         Rows expectedRows;
         {
-            auto transaction = _graph->openTransaction();
-            auto reader = transaction.readGraph();
-            const size_t numNodes = reader.getTotalNodesAllocated();
             // For each existing node we create a new node. New nodes start from current
             // max node ID
-            for (const NodeID n : reader.scanNodes()) {
-                expectedRows.add({n, n + numNodes, n + (2 * numNodes)});
+            for (const NodeID n : read().scanNodes()) {
+                expectedRows.add({n, n + numNodesPrior, n + (2 * numNodesPrior)});
             }
         }
 
@@ -186,9 +183,9 @@ TEST_F(WriteQueriesTest, scanNodesCreateNodes) {
         using Rows = LineContainer<NodeID>;
 
         Rows expectedRows;
-        { // We should now have 26 nodes
-            constexpr size_t EXP_NUM_NODES = 13 + (2 * 13);
-            for (size_t i = 0; i < EXP_NUM_NODES; i++) {
+        {
+            const size_t expectedNumNodes = numNodesPrior + (2 * numNodesPrior);
+            for (size_t i = 0; i < expectedNumNodes; i++) {
                 expectedRows.add({i});
             }
         }


v2/plan/PipelineGenerator.cpp
```diff
@@ -542,3 +575,130 @@ PipelineOutputInterface* PipelineGenerator::translateAggregateEvalNode(Aggregate
...
+            // Source and target nodes could be existing nodes (passed to WriteProcesor as
+            // input), or pending nodes (produced by WriteProcessor as output). If they
+            // are input, they will already have a registered column from a previous
+            // processor => present in @ref _declToColumn => set the tag here. Otherwise,
+            // call to @ref PipelineBuilder::addWrite will alloc column for the pending
+            // node, and we pass an invalid (default initialised) ColumnTag for such nodes
+            // to the @ref WriteProcessorTypes::PendingEdge to be later updated in the
+            // builder call.
+            const VarDecl* srcVar = pendingPlanEdge._src;
+            ColumnTag srcTag {};
```
rjb32, line 819: Why adding additional {} ? ColumnTag already has a default constructor

v2/plan/PipelineGenerator.cpp
```diff
@@ -542,3 +575,130 @@ PipelineOutputInterface* PipelineGenerator::translateAggregateEvalNode(Aggregate
...
+            for (const EntityPropertyConstraint& propConstr : data->exprConstraints()) {
+                throw PipelineException("Unsupported: Creating edges with properties");
+                const auto& [name, type, expr] = propConstr;
+                const ColumnTag propCol = getCol(expr->getExprVarDecl());
+                props.emplace_back(name, type, propCol);
+            }
+
+            // Initialise with invalid column tag, then later update after @ref addWrite
+            // allocs the column
+            const ColumnTag edgeTag {};
```
rjb32, line 850: What happens if all these tags don't get assigned a value?

rjb32:
Looks good to me overall. @LucLabarriere @suhassn-turing-bio Please review

storage/writers/MetadataBuilder.h
```diff
@@ -16,7 +15,7 @@ class MetadataRebaser;
 class MetadataBuilder {
 public:
     // Labels
-    LabelID getOrCreateLabel(const std::string& labelName);
+    LabelID getOrCreateLabel(std::string_view labelName);
```
LucLabarriere, line 18: Can you do that change to other metadata too? `PropertyTypeMap` and `EdgeTypeMap`, if not it's ok

test/v2/queries/DeleteQueriesTest.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    EXPECT_TRUE(expectedRows.equals(actualRows));
+}
+
+TEST_F(DeleteQueriesTest, deleteEdges) {
+    using NodeRows = LineContainer<NodeID>;
+    using EdgeRows = LineContainer<EdgeID>;
+
+    constexpr std::string_view deleteQuery = "MATCH (n)-[e]->(m) DELETE e";
```
LucLabarriere, line 141: Can you check what happens if you do:
```
MATCH (n)-[e]-(m) DELETE n
// or
MATCH (n), (m) DELETE n, m
```

Basically deleting multiple times the same entities

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+    _writeBuffer = &commitBuilder->writeBuffer();
+}
+
+void WriteProcessor::reset() {
+    markAsReset();
+}
+
+void WriteProcessor::execute() {
+    // NOTE: We currently do not have `CREATE (n) DELETE n` supported in PlanGraph,
+    // meaning if we are deleting, we require an input. This may change.
```
LucLabarriere, line 110: If CREATE (n) generates a column of IDs, can't the DELETE n take that column as input?

LucLabarriere: Oh I see the issue, there's a single WriteNode in the PlanGraph so the previous CREATES are not considered inputs. Maybe you could change the condition to something like that:
```
if (!deletedNodes.empty() || !_deletedEdges.empty()) {
    if (!_input) {
        throw NotImplementedYet("Only support MATCH ... DELETE ... for now")
    }
}
```

You could also try in performDeletions to get the deletedTag from inDf as you do, but if you don't find it, also check the outDf if it has it (you would have to reverse the operations in the `::execute()` method: creations first, then deletions.

But I'm starting to think a single WriteNode wasn't such a great idea.

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+    }
+}
+
+LabelSet WriteProcessor::getLabelSet(std::span<const std::string_view> labels) {
+    LabelSet labelset;
+    for (const std::string_view label : labels) {
+        const LabelID id = _metadataBuilder->getOrCreateLabel(label);
+        labelset.set(id);
+    }
+    return labelset;
```
LucLabarriere, line 182: We will have to think about what happens when we cancel a write during execution. We don't want partial results. In a regular DatabaseSystem, we would create one change per query (or at least one commit per query) and apply the whole thing automatically in a single transaction. In such a system, when we encounter an issue, we can just discard the change, but in our design, there's no way to do that, we have to remove everything up to the last commit

LucLabarriere: Yeah I really think we should experiment with 1 query == 1 commit, maybe even have a setting to test 1 query == 1 change and benchmark that

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+        auto* col = _output.getDataframe()->getColumn(node._tag)->as<ColumnNodeIDs>();
+        if (!col) { // @ref as<> performs dynamic_cast
+            throw FatalException(fmt::format("Column {} was marked as a column of"
+                                             " pending nodes, but is not a NodeID"
+                                             " column.", node._tag.getValue()));
+        }
+        bioassert(col->size() == 0);
+
+        const std::span labels = node._labels;
+        // TODO: Is this checked in planner?
```
LucLabarriere: Yes in `WriteStmtAnalyzer::analyze(NodePattern* nodePattern)`

```c++
const auto& labels = nodePattern->labels();
if (labels == nullptr) {
    throwError("Node pattern must have at least one label", nodePattern);
}

for (const Symbol* label : *labels) {
    data->addLabelConstraint(label->getName());
}
```

Can you just add a check for `!labels->empty()` (with an exception)? Because implicitly, if a NodePattern has a non nullptr SymbolChain* for labels, then it means it has at least one label, but if the implement changes someday, we might forget to change the code here

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+        bioassert(labels.size() > 0);
+        const LabelSet lblset = getLabelSet(labels);
```
LucLabarriere: I don't know if it changes anything but can you try to define the lblset variable outside the pendingNode loop and reassign it at each step? Maybe it helps the compiler a bit

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+        }
+        bioassert(col->size() == 0);
+
+        ColumnNodeIDs* srcCol {nullptr};
+        const ColumnTag srcTag = edge._srcTag;
+        // If the tag exists in the input, it must be an already existing NodeID
+        // column - e.g. something returned by a MATCH query - otherwise it is a
+        // pending node which was created in this CREATE query
+        const bool srcIsPending =
+            !_input || _input->getDataframe()->getColumn(srcTag) == nullptr;
```
LucLabarriere: can you define once inDf like you did for outDf?

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+        // However, all input columns are also propagated to the output, so we can
+        // always retrieve the column from the output dataframe, regardless of whether
+        // the source node was the result of a CREATE or a MATCH
+        if (!outDf->hasColumn(srcTag)) {
+            throw FatalException(fmt::format("Attempted to create edge {} with "
+                                             "source node with no such column: {}",
+                                             edge._name, edge._srcTag.getValue()));
+        }
+        srcCol = _output.getDataframe()->getColumn(srcTag)->as<ColumnNodeIDs>();
```
LucLabarriere: reuse outDf instead of retrieving it again from _output

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+        const ColumnTag tgtTag = edge._tgtTag;
+        const bool tgtIsPending =
+            !_input || _input->getDataframe()->getColumn(tgtTag) == nullptr;
+
+        if (!outDf->hasColumn(tgtTag)) {
+            throw FatalException(fmt::format("Attempted to create edge {} with "
+                                             "target node with no such column: {}",
+                                             edge._name, edge._tgtTag.getValue()));
+        }
+        tgtCol = _output.getDataframe()->getColumn(tgtTag)->as<ColumnNodeIDs>();
```
LucLabarriere: same thing, here use outDf. Throughout the function, reuse variables outDf and inDf

v2/pipeline/processors/WriteProcessor.cpp
LucLabarriere: maybe something like DataFrame* inDf = _input ? _input->getDataframe() : nullptr;

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+        // Populate the output column for this edge with the index in the CWB which it
+        // appears. These indexes are later transformed into "fake IDs" (an estimate as to
+        // what the EdgeID will be when it is committed) in @ref postProcessFakeIDs.
+        raw.resize(raw.size() + numIters);
+        std::iota(col->begin(), col->end(), nextEdgeID);
+        nextEdgeID += numIters;
+    }
+}
+
+void WriteProcessor::postProcessTempIDs() {
```
LucLabarriere, line 334: I really don't like this but I don't know how to fix that, it's starting to feel like we needed a unique ID system ahah

LucLabarriere, line 334: Same thing as above, in this function, try to reuse outDf everywhere

v2/pipeline/processors/WriteProcessor.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+void WriteProcessor::performCreations() {
+    // We apply the CREATE command for each row in the input, or just a single row if we
+    // have no inputs
+    const size_t numIters = _input ? _input->getDataframe()->getRowCount() : 1;
+
+    createNodes(numIters);
+    createEdges(numIters);
+
+    postProcessTempIDs();
```
LucLabarriere, line 389: This is a really clear way to handle the inputs vs no inputs case I think

v2/pipeline/processors/WriteProcessor.h
```diff
@@ -0,0 +1,127 @@
+#pragma once
+
+#include <cmath>
+#include <optional>
+#include <string_view>
+
+#include "FatalException.h"
```
LucLabarriere: FatalException at the end of the includes if I'm not mistaken

v2/pipeline/processors/WriteProcessorTypes.h
```diff
@@ -0,0 +1,83 @@
...
+              _edgeType(type),
+              _name(name),
+              _srcName(srcName),
+              _tgtName(tgtName),
+              _tag(tag),
+              _srcTag(srcTag),
+              _tgtTag(tgtTag)
+        {
+        }
+    };
```
LucLabarriere, line 69: The copies might become quite expensive, we might want to use those structs directly in the PlanGraph node so that the WriteProcessor can just take them as is an only fill the columns.

v2/pipeline/processors/WriteProcessorTypes.h
```diff
@@ -0,0 +1,83 @@
...
+            : _properties(std::move(props)),
+              _edgeType(type),
+              _name(name),
+              _srcName(srcName),
+              _tgtName(tgtName),
+              _tag(tag),
+              _srcTag(srcTag),
+              _tgtTag(tgtTag)
+        {
+        }
```
LucLabarriere, line 68: Try to remove the constructor and use aggregate initialization instead:

```c++
// When creating an edge:
PendingEdge edge {
    ._properties = std::move(properties),
    ._edgeType = edgeType,
    /* ... */
};
```

v2/pipeline/processors/WriteProcessorTypes.h
```diff
@@ -0,0 +1,83 @@
...
+        ColumnTag _tag;
+
+        PendingNode(Labels&& labels, PropertyConstraints&& props, std::string_view name, ColumnTag tag)
+            : _labels(std::move(labels)),
+            _properties(std::move(props)),
+            _name(name),
+            _tag(tag)
+        {
+        }
+    };
```
LucLabarriere, line 39: Test an aggregate initialisation style

v2/plan/PipelineGenerator.cpp
```diff
@@ -542,3 +575,130 @@ PipelineOutputInterface* PipelineGenerator::translateAggregateEvalNode(Aggregate
...
+    }
+
+    WriteProcessor::PendingNodes penNodes;
+    penNodes.reserve(node->pendingNodes().size());
+    {
+        for (const WriteNode::PendingNode& pendingPlanNode : node->pendingNodes()) {
+            const NodePatternData* const data = pendingPlanNode._data;
+
+            const std::string_view nodeVarName = pendingPlanNode._name->getName();
+            // XXX: Does the plan graph ensure a valid node? e.g. at least one label, etc?
```
LucLabarriere: The checks are performed in WriteStmtAnalyzer::analyzer(NodePattern*)

v2/plan/PipelineGenerator.cpp
cyrusknopf: I wanted to make it explicit that it was default initialised and would therefore be overwritten later when we know the tag, but I can remove.

v2/plan/PipelineGenerator.cpp
cyrusknopf: There is a check in `WriteProcessor::execute`, but if they ever do not get assigned a value it is a logic error somewhere on our part.
```
if (!outDf->hasColumn(edge._tag)) {
            throw FatalException(
                fmt::format("Could not get column in WriteProcessor "
                            "output dataframe for PendingEdge with tag {}.",
                            edge._tag.getValue()));
        }
```

storage/writers/MetadataBuilder.h
cyrusknopf: Done!

test/v2/queries/DeleteQueriesTest.cpp
cyrusknopf: `MATCH (n)-[e]-(m) DELETE n` works fine - did you mean something else? `n` only gets deleted once

`MATCH (n), (m) DELETE n, m` seems to work fine as well, I will add a unit test to make sure

v2/pipeline/processors/WriteProcessor.cpp
cyrusknopf: Yeah I noticed that Neo4j seems to have two separate processors/nodes for Writes and Deletes. Maybe we should look into that. I'll add the exception as you suggest for now

v2/pipeline/processors/WriteProcessor.cpp
cyrusknopf: I agree that this is not ideal at all to have the side effects of creating labels even if a query cancels during execution.

I think 1 change = 1 query is a good idea, but I think we need to separate the concepts of "change" and "branch" first

v2/pipeline/processors/WriteProcessor.cpp
cyrusknopf: Yes good point I've added an exception there

v2/pipeline/processors/WriteProcessor.cpp
cyrusknopf: Done

test/v2/queries/DeleteQueriesTest.cpp
LucLabarriere: "`n only gets deleted once`" Ok that's what I wanted to know, thanks!

v2/pipeline/processors/WriteProcessor.cpp
cyrusknopf: Nice idea, done

v2/pipeline/processors/WriteProcessor.cpp
cyrusknopf: Good catch, done

v2/pipeline/processors/WriteProcessor.cpp
cyrusknopf: I agree, I don't like it at all. Not even sure unique IDs would help us here

cyrusknopf: Done

v2/pipeline/processors/WriteProcessor.h
cyrusknopf: Yep absolutely good catch

v2/pipeline/processors/WriteProcessorTypes.h
cyrusknopf: Yeah it felt odd redefining these structs with so many members, but the problem is that if we do not I think both need to access VarDecl and other Plan types. Which I thought kind of muddled the boundaries of concerns

v2/pipeline/processors/WriteProcessorTypes.h
cyrusknopf: I explicitly define these constructors so that using `emplace_back` requires everything to be initialised. I use emplace back because otherwise there are a lot of temporaries that need to be created in the loop due to the nature of how PendingNodes/Edges are created. So I never instantiate an edge as you show in the aggregate initialisation above.

I'm not sure if theres a way to enforce a struct to have all members explicitly initialised (especially during an "indirect" instantiation such as `emplace_back`) without defining the constructors like this

v2/pipeline/processors/WriteProcessor.h
```diff
@@ -0,0 +1,128 @@
+#pragma once
+
+#include <cmath>
```
rjb32: Rather include math.h

v2/pipeline/processors/WriteProcessorTypes.h
```diff
@@ -0,0 +1,83 @@
+#pragma once
+
+#include <cstddef>
```
rjb32: stddef.h

v2/plan/PipelineGenerator.cpp
```diff
@@ -91,6 +103,19 @@ struct PropertyTypeDispatcher {
 
 }
 
+db::ColumnTag PipelineGenerator::getCol(const VarDecl* var) {
```
rjb32: Do using namespace db to get ColumnTag type directly

v2/plan/PipelineGenerator.cpp
```diff
@@ -91,6 +103,19 @@ struct PropertyTypeDispatcher {
...
+    if (!var) {
+        throw FatalException("Attempted to get column for null variable");
+    }
+
+    const auto it = _declToColumn.find(var);
+    if (it == end(_declToColumn)) {
+        throw PlannerException(fmt::format("Failed to find column for variable {}.", var->getName()));
+    }
+    ColumnTag tag = it->second;
```
rjb32: tag can be const

v2/plan/PipelineGenerator.cpp
rjb32: it's probably best to add a check

cyrusknopf: Where should I add the extra check? The check I showed above is in `WriteProcessor::execute`

v2/pipeline/processors/WriteProcessor.h
cyrusknopf: Sure, why do we prefer the C headers?

rjb32:
@cyrusknopf I think it can be merged modulo a few last minor stylistic fixes, and then go on implementing it with properties later

v2/plan/PipelineGenerator.cpp
cyrusknopf: Done

v2/plan/PipelineGenerator.cpp
cyrusknopf: Done

v2/pipeline/processors/WriteProcessorTypes.h
cyrusknopf: Done

v2/pipeline/processors/WriteProcessor.h
cyrusknopf: Done

PR #148, opened 2025-11-20
Fix incorrect pending output state when blocked on binary nodes in PipelineGenerator

cyrusknopf:
Context:
- Binary nodes are "blocking" in the DFS traversal because they require both their inputs to be translated before they themselves can be translated
- This "blocking" causes inconsistencies in the `_pendingOutput` member of builder since branches are no longer traversed from root to leaf
- See example below for `MATCH (n)-[e]->(m), (n)<-[f]-(m) RETURN e`:
IMG_4BE07F52F4D4-1
- DFS begins on the red path, but is blocked at node (a)
- DFS continues on the blue path, but at node (b) the `_pendingOutput` member is set to node (a), which is obviously incorrect

Changes:
- DFS stack now stores "previous interface" which contains the output interface of the node which pushed it onto the stack

Verified with:
- Additional unit test in `QueriesTest.cpp` which shows the query `MATCH (n)-[e]->(m), (n)<-[f]-(m) RETURN e` (whose plan graph is pictured in the above example) executes correctly

### DIFF
--- a/v2/plan/PipelineGenerator.cpp
+++ b/v2/plan/PipelineGenerator.cpp
@@ -103,9 +103,7 @@ void PipelineGenerator::generate() {
         const bool binaryAndVisited =
             node->isBinary() && _binaryVisitedMap.contains(node);
         if (!binaryAndVisited) {
-            if (_builder.getPendingOutput().getInterface() != prevIf) {
-                _builder.getPendingOutput().setInterface(prevIf);
-            }
+            _builder.getPendingOutput().setInterface(prevIf);
         }
 
         // If we have no previous interface to take inputs from when we should: error
@@ -456,7 +454,7 @@ PipelineOutputInterface* PipelineGenerator::translateCartesianProductNode(Cartes
     PipelineOutputInterface* rhs = isBLhs ? inputA : inputB;
 
     // LHS is implicit in @ref _pendingOutput
-    _builder.getPendingOutput().setInterface(lhs);
+    _builder.getPendingOutput().updateInterface(lhs);
 
     _builder.addCartesianProduct(rhs);
     return _builder.getPendingOutputInterface();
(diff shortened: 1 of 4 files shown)


v2/plan/PipelineGenerator.cpp
```diff
@@ -79,20 +87,40 @@ void PipelineGenerator::generate() {
...
-        translateNode(node);
+        // Binary nodes that are encountered for the second time should not reset @ref
+        // _builder::_pendingOutput with the first-encounter-parent (that stored in @ref
+        // prevIf), because that is already stored in @ref _binaryVisitedMap. Instead, on
+        // the second visit to binary nodes, keep the current pending output: do not
+        // overwrite it.
+        const bool binaryAndVisited =
+            node->isBinary() && _binaryVisitedMap.contains(node);
+        if (!binaryAndVisited) {
+            if (_builder.getPendingOutput().getInterface() != prevIf) {
```
cyrusknopf: This "if" should probably be removed actually, and just always update the pending output

rjb32: Yes agreed

rjb32:
The idea seems good to me

cyrusknopf:
@rjb32 @LucLabarriere Since your approval I have renamed `PendingOutputView::setInterface` to `PendingOutputView::updateInterface` because it was not a true setter (it had other side effects, namely changing the "edges projected" member flag. I made `PendingOutputView::setInterface` to be a true setter.

rjb32:
> @rjb32 @LucLabarriere Since your approval I have renamed `PendingOutputView::setInterface` to `PendingOutputView::updateInterface` because it was not a true setter (it had other side effects, namely changing the "edges projected" member flag. I made `PendingOutputView::setInterface` to be a true setter.

Indeed makes sense

PR #150, opened 2025-11-25
Add comments to the loop in PipelineGenerator

rjb32:
Add comments to the for each next node in PipelineGenerator to clarify a bit what's going on.
Expand to if-else style instead of relying on continue statements to make the different cases clearer.

### DIFF
--- a/v2/plan/PipelineGenerator.cpp
+++ b/v2/plan/PipelineGenerator.cpp
@@ -43,8 +43,8 @@ using namespace db::v2;
 namespace {
 
 struct TranslateNodeToken {
-    PlanGraphNode* node;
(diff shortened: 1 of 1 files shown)


v2/plan/PipelineGenerator.cpp
```diff
@@ -43,8 +43,8 @@ using namespace db::v2;
 namespace {
 
 struct TranslateNodeToken {
-    PlanGraphNode* node;
-    PipelineOutputInterface* previousInterface;
+    PlanGraphNode* _node {nullptr};
+    PipelineOutputInterface* _previousInterface {nullptr};
```
cyrusknopf, line 47: Oops good catch, my mistake on this

PR #151, opened 2025-11-26
GetProperties implementation in PipelineGenerator. Generalized EntityOutputStream to all interfaces

LucLabarriere:
- `EntityOutputStream` is now common to all interfaces which simplifies greatly parts of the code, such as `translateVarNode` that now only operates on the stream
- It's a bit ugly right now that in the PipelineBuilder we have to remember to update the stream every time we create a new processor.
- GetProperties adds a Materialize right after. Later on it could be better to write a new NodeIDs/EdgeIDs column in the ChunkWriter that would take over the output of ScanNodeIDs (and other processors)

### DIFF
--- a/v2/pipeline/interfaces/PipelineInputInterface.h
+++ b/v2/pipeline/interfaces/PipelineInputInterface.h
@@ -1,8 +1,9 @@
 #pragma once
 
-#include "EntityOutputStream.h"
 #include "EnumToString.h"
 #include "PipelinePort.h"
+
+#include "EntityOutputStream.h"
 
 namespace db {
 class NamedColumn;
--- a/v2/pipeline/interfaces/PipelineOutputInterface.h
+++ b/v2/pipeline/interfaces/PipelineOutputInterface.h
@@ -2,9 +2,10 @@
 
 #include <string_view>
 
-#include "EntityOutputStream.h"
 #include "PipelineInputInterface.h"
 #include "PipelinePort.h"
+
+#include "EntityOutputStream.h"
 
 namespace db {
 class NamedColumn;
--- a/v2/plan/PipelineGenerator.cpp
+++ b/v2/plan/PipelineGenerator.cpp
@@ -246,15 +246,15 @@ PipelineOutputInterface* PipelineGenerator::translateVarNode(VarNode* node) {
 
     const auto visitor = Overloaded {
         [&](const EntityOutputStream::NodeStream& stream) {
-            bioassert(stream._nodeIDsTag.isValid() && "NodeStream does not have a nodeIDsTag");
-            bioassert(outDf->getColumn(stream._nodeIDsTag) && "NodeStream does not have a nodeIDs column");
+            msgbioassert(stream._nodeIDsTag.isValid(), "NodeStream does not have a nodeIDsTag");
+            msgbioassert(outDf->getColumn(stream._nodeIDsTag), "NodeStream does not have a nodeIDs column");
 
             _declToColumn[node->getVarDecl()] = stream._nodeIDsTag;
             outDf->getColumn(stream._nodeIDsTag)->rename(varName);
         },
         [&](const EntityOutputStream::EdgeStream& stream) {
-            bioassert(stream._edgeIDsTag.isValid() && "EdgeStream does not have a edgeIDsTag");
-            bioassert(outDf->getColumn(stream._edgeIDsTag) && "EdgeStream does not have a edgeIDs column");
+            msgbioassert(stream._edgeIDsTag.isValid(), "EdgeStream does not have a edgeIDsTag");
+            msgbioassert(outDf->getColumn(stream._edgeIDsTag), "EdgeStream does not have a edgeIDs column");
 
             _declToColumn[node->getVarDecl()] = stream._edgeIDsTag;
             outDf->getColumn(stream._edgeIDsTag)->rename(varName);
@@ -298,34 +298,39 @@ PipelineOutputInterface* PipelineGenerator::translateGetPropertyNode(GetProperty
         throw PlannerException("GetPropertyNode does not have an entity variable declaration");
     }
 
-
     const std::string propName {node->getPropName()};
 
     PipelineValuesOutputInterface* output = nullptr;
 
+    // Retrieving the property type from the graph metadata
     const std::optional<PropertyType> foundProp = _view.read().getMetadata().propTypes().get(propName);
     if (!foundProp) {
         throw PlannerException(fmt::format("Property type {} does not exist", propName));
     }
 
-    if (entityDecl->getType() == EvaluatedType::NodePattern) {
-        const auto process = [&]<SupportedType Type> {
-            output = &_builder.addGetProperties<EntityType::Node, Type>(*foundProp);
-        };
+    // Adding the GetProperty processor to the pipeline
+    switch (entityDecl->getType()) {
+        case EvaluatedType::NodePattern: {
+            const auto process = [&]<SupportedType Type> {
+                output = &_builder.addGetProperties<EntityType::Node, Type>(*foundProp);
+            };
+            PropertyTypeDispatcher {foundProp->_valueType}.execute(process);
+        }
+        case EvaluatedType::EdgePattern: {
+            const auto process = [&]<SupportedType Type> {
+                output = &_builder.addGetProperties<EntityType::Edge, Type>(*foundProp);
+            };
 
-        PropertyTypeDispatcher {foundProp->_valueType}.execute(process);
-    } else if (entityDecl->getType() == EvaluatedType::EdgePattern) {
-        const auto process = [&]<SupportedType Type> {
-            output = &_builder.addGetProperties<EntityType::Edge, Type>(*foundProp);
-        };
-
-        PropertyTypeDispatcher {foundProp->_valueType}.execute(process);
-    } else {
-        throw PlannerException(fmt::format(
-            "GetProperty must act on a Node/EdgePattern. Instead acting on {}",
-            EvaluatedTypeName::value(entityDecl->getType())));
+            PropertyTypeDispatcher {foundProp->_valueType}.execute(process);
+        }
+        default: {
+            throw PlannerException(fmt::format(
+                "GetProperty must act on a Node/EdgePattern. Instead acting on {}",
+                EvaluatedTypeName::value(entityDecl->getType())));
+        }
     }
 
+    // Mapping the expr decl to the column tag
     const Expr* expr = node->getExpr();
     if (!expr) {
         throw PlannerException("GetPropertyNode does not have an expression");
@@ -338,6 +343,7 @@ PipelineOutputInterface* PipelineGenerator::translateGetPropertyNode(GetProperty
 
     _declToColumn[exprDecl] = output->getValues()->getTag();
 
+    // Adding the materialize step
     _builder.addMaterialize();
 
     return _builder.getPendingOutputInterface();


v2/plan/PipelineGenerator.cpp
```diff
@@ -234,23 +239,30 @@ PipelineOutputInterface* PipelineGenerator::translateVarNode(VarNode* node) {
...
+        [&](const EntityOutputStream::EdgeStream& stream) {
+            bioassert(stream._edgeIDsTag.isValid() && "EdgeStream does not have a edgeIDsTag");
+            bioassert(outDf->getColumn(stream._edgeIDsTag) && "EdgeStream does not have a edgeIDs column");
+
+            _declToColumn[node->getVarDecl()] = stream._edgeIDsTag;
+            outDf->getColumn(stream._edgeIDsTag)->rename(varName);
+        },
+    };
+
+    stream.visit(visitor);
```
LucLabarriere, line 264: New code for the translateVarNode

v2/plan/PipelineGenerator.cpp
```diff
@@ -280,6 +292,62 @@ PipelineOutputInterface* PipelineGenerator::translateGetEdgeTargetNode(GetEdgeTa
...
+    }
+
+    const std::string_view exprStrRep = _sourceManager->getStringRepr((std::uintptr_t)expr);
+    if (!exprStrRep.empty()) {
+        output->rename(exprStrRep);
+    }
+
+    _declToColumn[exprDecl] = output->getValues()->getTag();
+
+    _builder.addMaterialize();
```
LucLabarriere, line 347: GetProperties adds a Materialize node after. Maybe we can find a better way later on (produce a new id column directly in the GetPropertiesChunkWriter)

v2/plan/PipelineGenerator.cpp
```diff
@@ -280,6 +292,62 @@ PipelineOutputInterface* PipelineGenerator::translateGetEdgeTargetNode(GetEdgeTa
...
+    }
+
+    const VarDecl* exprDecl = expr->getExprVarDecl();
+    if (!exprDecl) [[unlikely]] {
+        throw PlannerException("GetPropertyNode does not have an expression variable declaration");
+    }
+
+    const std::string_view exprStrRep = _sourceManager->getStringRepr((std::uintptr_t)expr);
+    if (!exprStrRep.empty()) {
+        output->rename(exprStrRep);
```
LucLabarriere: Right now, this does not work for property expressions like this:
```
MATCH (n { name: "Luc" }) RETURN n.name
```

The column header will be something like `$3` instead of `n.name`. That's because the expression n.name is not the one that is returned. Instead we return the dummy expression that is created in the Analyzer to evaluate the filter expression (meaning that it does not have a name)

v2/plan/ReadStmtGenerator.cpp
```diff
@@ -441,25 +441,46 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+                const auto* cached = getEntityTypeCache.cacheOrRetrieve(entityDecl, exprDecl);
+
+                if (cached) {
+                    // GetEntityType is already present in the cache. Map the existing expr to the current one
+                    if (!cached->_exprDecl) [[unlikely]] {
+                        throwError("GetEntityType expression does not have an expression variable declaration", expr);
+                    }
 
-                if (!_tree->cacheGetEntityType(entityDecl)) {
+                    expr->setExprVarDecl(cached->_exprDecl);
```
LucLabarriere, line 483: This is a bit of a hack, we actually need a proper expression map that knows that expressions and sub-expressions might be reused. For example in `MATCH (n { name: "Luc" }) RETURN n.name` there are two expressions that generate the name of node n, the one in the return and the one in the property constraint. This hack covers this case, but more generally, we could want to reuse sub-expressions:

```
MATCH (n), (m) WHERE n.age + 5 > m.age RETURN n.age + 5
```

We could detect that n.age + 5 is used in multiple places

LucLabarriere:
@rjb32 Actually, my hack to reuse expressions broke the expr dependency management

LucLabarriere:
> @rjb32 Actually, my hack to reuse expressions broke the expr dependency management

Actually, fixed it

v2/plan/PipelineGenerator.cpp
```diff
@@ -234,23 +239,30 @@ PipelineOutputInterface* PipelineGenerator::translateVarNode(VarNode* node) {
...
 
-    if (const auto* out = dynamic_cast<const PipelineNodeOutputInterface*>(output)) {
-        _declToColumn[node->getVarDecl()] = out->getNodeIDs()->getTag();
-    } else if (const auto* out = dynamic_cast<const PipelineEdgeOutputInterface*>(output)) {
-        const PendingOutputView& outView = _builder.getPendingOutput();
+    Dataframe* outDf = output->getDataframe();
+
+    const auto visitor = Overloaded {
+        [&](const EntityOutputStream::NodeStream& stream) {
+            bioassert(stream._nodeIDsTag.isValid() && "NodeStream does not have a nodeIDsTag");
```
cyrusknopf: Would we prefer to use `msgbioassert` instead of the `&&` method?

v2/plan/PipelineGenerator.cpp
LucLabarriere: Fixed in latest commit, now we do a final rename in the projection which should always work

LucLabarriere:
There's a bug right now if I run this second query

```
turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x
[QueryInterpreterV2::execute]: 992.726 us (100.00 %)

Query returned 26 rows.
Query executed in 0.905026 ms.
turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x.name
[2025-11-27 08:06:24] [error] EXEC_ERROR: Unexpected exception: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
```

The exception occurs in the CartesianProduct. I think it's because there's a missing materialize in one of the branches. @cyrusknopf @rjb32 Any idea? Could it be related to the issue with how we handle branches in the pipeline generator algorithm?

cyrusknopf:
> There's a bug right now if I run this second query
> 
> ```
> turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x
> [QueryInterpreterV2::execute]: 992.726 us (100.00 %)
> 
> Query returned 26 rows.
> Query executed in 0.905026 ms.
> turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x.name
> [2025-11-27 08:06:24] [error] EXEC_ERROR: Unexpected exception: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
> ```
> 
> The exception occurs in the CartesianProduct. I think it's because there's a missing materialize in one of the branches. @cyrusknopf @rjb32 Any idea? Could it be related to the issue with how we handle branches in the pipeline generator algorithm?

Yeah this is to do with the materialise I believe, we currently have one materialise for all "branches", but in fact we need one materialise proc for each branch. This is part of Remy's fixes, but I'm not sure if you want to just disable this test for now, or something else.

v2/plan/PipelineGenerator.cpp
LucLabarriere: I was wondering too. I guess '&&' is more idiomatic and it avoids having two separate macros that do the same thing, but maybe it's ugly. What do you prefer?

LucLabarriere:
> > There's a bug right now if I run this second query
> > ```
> > turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x
> > [QueryInterpreterV2::execute]: 992.726 us (100.00 %)
> > 
> > Query returned 26 rows.
> > Query executed in 0.905026 ms.
> > turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x.name
> > [2025-11-27 08:06:24] [error] EXEC_ERROR: Unexpected exception: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The exception occurs in the CartesianProduct. I think it's because there's a missing materialize in one of the branches. @cyrusknopf @rjb32 Any idea? Could it be related to the issue with how we handle branches in the pipeline generator algorithm?
> 
> Yeah this is to do with the materialise I believe, we currently have one materialise for all "branches", but in fact we need one materialise proc for each branch. This is part of Remy's fixes, but I'm not sure if you want to just disable this test for now, or something else.

Ok good to know. It's not in a unit test yet. Actually there are no unit tests on this yet because the implementation can only be tested once with have filter expressions (here I disable the check on non-empty filter steps to run my tests). So I assume we'll find out if it works once we have branches in the pipeline generator?

ghost:
> > There's a bug right now if I run this second query
> > ```
> > turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x
> > [QueryInterpreterV2::execute]: 992.726 us (100.00 %)
> > 
> > Query returned 26 rows.
> > Query executed in 0.905026 ms.
> > turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x.name
> > [2025-11-27 08:06:24] [error] EXEC_ERROR: Unexpected exception: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The exception occurs in the CartesianProduct. I think it's because there's a missing materialize in one of the branches. @cyrusknopf @rjb32 Any idea? Could it be related to the issue with how we handle branches in the pipeline generator algorithm?
> 
> Yeah this is to do with the materialise I believe, we currently have one materialise for all "branches", but in fact we need one materialise proc for each branch. This is part of Remy's fixes, but I'm not sure if you want to just disable this test for now, or something else.

@cyrusknopf this test worked in your CartesianProduct tests right?

LucLabarriere:
> > > There's a bug right now if I run this second query
> > > ```
> > > turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x
> > > [QueryInterpreterV2::execute]: 992.726 us (100.00 %)
> > > 
> > > Query returned 26 rows.
> > > Query executed in 0.905026 ms.
> > > turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x.name
> > > [2025-11-27 08:06:24] [error] EXEC_ERROR: Unexpected exception: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The exception occurs in the CartesianProduct. I think it's because there's a missing materialize in one of the branches. @cyrusknopf @rjb32 Any idea? Could it be related to the issue with how we handle branches in the pipeline generator algorithm?
> > 
> > 
> > Yeah this is to do with the materialise I believe, we currently have one materialise for all "branches", but in fact we need one materialise proc for each branch. This is part of Remy's fixes, but I'm not sure if you want to just disable this test for now, or something else.
> 
> @cyrusknopf this test worked in your CartesianProduct tests right?

This works:
```c++
#v2 MATCH (a), (x) RETURN a.age, x.name
```

But it's not the same. In this case there is no need for materialize so it works. This query is more complex:

```
#v2 MATCH (a { age: 32 }), (x) RETURN a.age, x.name
```

Because the first branch needs a Materialize:

```
1st branch: ScanNodes(a) -> GetProperties(a.age) -> Materialize
2nd branch: ScanNodes(x)

Then we join the two branches with a cartesian product and we get:
CartesianProduct -> GetPropertiesWithNull(x.name)
```

v2/plan/PipelineGenerator.cpp
ghost: This is what herb stutter was using as well so I think it's pretty idiomatic haha.

rjb32: It's better to use msgbioassert we are not like we are doing C with good old asserts aha.. Especially if we implement bioassert with exceptions to not abort completely one day

v2/plan/PipelineGenerator.cpp
```diff
@@ -280,6 +292,62 @@ PipelineOutputInterface* PipelineGenerator::translateGetEdgeTargetNode(GetEdgeTa
     return _builder.getPendingOutputInterface();
 }
 
+PipelineOutputInterface* PipelineGenerator::translateGetPropertyNode(GetPropertyNode* node) {
+    const VarDecl* entityDecl = node->getEntityVarDecl();
+    if (!entityDecl) {
+        throw PlannerException("GetPropertyNode does not have an entity variable declaration");
+    }
+
```
rjb32, line 300: Too many blank lines

v2/plan/PipelineGenerator.cpp
```diff
@@ -280,6 +292,62 @@ PipelineOutputInterface* PipelineGenerator::translateGetEdgeTargetNode(GetEdgeTa
...
+    const std::string propName {node->getPropName()};
+
+    PipelineValuesOutputInterface* output = nullptr;
+
+    const std::optional<PropertyType> foundProp = _view.read().getMetadata().propTypes().get(propName);
+    if (!foundProp) {
+        throw PlannerException(fmt::format("Property type {} does not exist", propName));
+    }
+
+    if (entityDecl->getType() == EvaluatedType::NodePattern) {
```
rjb32: It's cleaner to use a switch here on the type instead of relying on a bunch of if statements.
Also please put comments to delimit visually the different steps

v2/plan/PipelineGenerator.cpp
rjb32: Yes that would be great

ghost: the get properties node will only be inserted before a filter or a produce results node right? Aren't they already adding the materialize processor?

v2/plan/ReadStmtGenerator.cpp
rjb32: Indeed, a solution to this could be to do structural hashing on expressions so that you have unique subexpressions in query that are reused

v2/pipeline/interfaces/PipelineOutputInterface.h
```diff
@@ -2,6 +2,7 @@
 
 #include <string_view>
 
+#include "EntityOutputStream.h"
```
rjb32: It's better to put it below the others

v2/pipeline/interfaces/PipelineInputInterface.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "EntityOutputStream.h"
```
rjb32: Order of includes

v2/pipeline/EntityOutputStream.h
```diff
@@ -1,13 +1,27 @@
...
+#include <variant>
 
+#include "PipelineException.h"
 #include "dataframe/ColumnTag.h"
 
 namespace db::v2 {
 
 class EntityOutputStream {
 public:
+    struct NodeStream {
```
rjb32, line 13: It's quite nice

cyrusknopf:
> > > There's a bug right now if I run this second query
> > > ```
> > > turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x
> > > [QueryInterpreterV2::execute]: 992.726 us (100.00 %)
> > > 
> > > Query returned 26 rows.
> > > Query executed in 0.905026 ms.
> > > turing:simpledb>  #v2 MATCH (a { age: 32 }), (x) RETURN a.age, x.name
> > > [2025-11-27 08:06:24] [error] EXEC_ERROR: Unexpected exception: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The exception occurs in the CartesianProduct. I think it's because there's a missing materialize in one of the branches. @cyrusknopf @rjb32 Any idea? Could it be related to the issue with how we handle branches in the pipeline generator algorithm?
> > 
> > 
> > Yeah this is to do with the materialise I believe, we currently have one materialise for all "branches", but in fact we need one materialise proc for each branch. This is part of Remy's fixes, but I'm not sure if you want to just disable this test for now, or something else.
> 
> @cyrusknopf this test worked in your CartesianProduct tests right?

I've never done a test like this

v2/plan/PipelineGenerator.cpp
cyrusknopf: Its idiomatic if you are using standard C asserts which have no builtin messaging but IMO since we have an explicit `msgbioassert` defined it makes sense to use it.

And since we have talked about changing the `bioassert` definition to throw an exception, I think this && syntax would need to be changed in every place it is used, whereas a bare `bioassert` or `msgbioassert` will be able to be changed to throw an exception without needing any changes at the callsite

cyrusknopf: but who knows if we will change them to throw an exception :p

v2/plan/PipelineGenerator.cpp
LucLabarriere: No not necessarily, right now we have to materialize after each GetProperties because it implicitely filters meaning that the GetOutEdges that comes after needs to work on filtered out ids:

```
ScanNodes -> GetProperties -> GetOutEdges
```
This would not work properly because the GetOutEdges is acting on all nodeIDs although that were filtered by the GetProperties

Instead, this is better for now:
```
ScanNodes -> GetProperties -> Materialize -> GetOutEdges
```

Even better would be to override only the NodeIDs column in the GetProperties so that we don't need the materialize, but this is quite complex for this PR

PR #152, opened 2025-11-26
New pipeline generation by branches

rjb32:
New translation of the PlanGraph branch-by-branch.

1. **PipelineBranches**: extract the branches of a PlanGraph and do a topological sort of the branches. So no more DFS afterwards.

2. **PipelineGenerator** is now an object that wraps the translation of the entire PlanGraph, gets the branches and invokes BranchGenerator to translate each branch according to the topological sort.

3.  **BranchGenerator** contains the "sequential" translation part of the current PipelineGenerator. The big for loop has disappeared. Its mission is really to translate a sequential branch of nodes and it can use PipelineBuilder under the hood to hide the complexity of pipeline connections & interface management.

For binary nodes such as cartesian product: the binaryVisitedMap has been removed because I have now the interfaces of the previous branches that are connected to the node. Binary nodes are always in their own branch for simplicity sake. 

Bonus: I implemented a quick PipelineDebug::dumpMermaid which is very practical to see the pipeline

### DIFF
--- a/v2/plan/BranchGenerator.cpp
+++ b/v2/plan/BranchGenerator.cpp
@@ -95,14 +95,12 @@ BranchGenerator::~BranchGenerator() {
 void BranchGenerator::translateBranch(PipelineBranch* branch) {
     setupBranch(branch);
 
-    fmt::print("Branch\n");
     for (PlanGraphNode* node : branch->nodes()) {
-        fmt::print("Translate node {} {}\n", fmt::ptr(node), PlanGraphOpcodeDescription::value(node->getOpcode()));
         translateNode(node);
     }
 
     // Add materialize at the end of the branch if necessary
-    if (_builder.isMaterializeOpen() && !_builder.isSingleMaterializeStep()) {
+    if (!_builder.isSingleMaterializeStep()) {
         _builder.addMaterialize();
     }
 
@@ -194,6 +192,7 @@ PipelineOutputInterface* BranchGenerator::translateNode(PlanGraphNode* node) {
         case PlanGraphOpcode::WRITE:
         case PlanGraphOpcode::FUNC_EVAL:
         case PlanGraphOpcode::ORDER_BY:
+        case PlanGraphOpcode::PROCEDURE_EVAL:
         case PlanGraphOpcode::UNKNOWN:
         case PlanGraphOpcode::_SIZE:
             throw PlannerException(fmt::format("BranchGenerator does not support PlanGraphNode: {}",
@@ -320,7 +319,7 @@ PipelineOutputInterface* BranchGenerator::translateGetPropertyNode(GetPropertyNo
 }
 
 PipelineOutputInterface* BranchGenerator::translateGetPropertyWithNullNode(GetPropertyWithNullNode* node) {
-    if (_builder.isMaterializeOpen() && !_builder.isSingleMaterializeStep()) {
+    if (!_builder.isSingleMaterializeStep()) {
         _builder.addMaterialize();
     }
 
@@ -390,7 +389,7 @@ PipelineOutputInterface* BranchGenerator::translateEdgeFilterNode(EdgeFilterNode
 PipelineOutputInterface* BranchGenerator::translateProduceResultsNode(ProduceResultsNode* node) {
     // If MaterializeNode has not been seen at that point,
     // we materialize if we have data in flight
-    if (_builder.isMaterializeOpen() && !_builder.isSingleMaterializeStep()) {
+    if (!_builder.isSingleMaterializeStep()) {
         _builder.addMaterialize();
     }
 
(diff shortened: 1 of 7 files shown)


v2/plan/PipelineBranch.h
```diff
@@ -0,0 +1,52 @@
...
+        _nodes.push_back(node);
+    }
+
+    void connectTo(PipelineBranch* nextBranch) {
+        _outputs.emplace_back(nextBranch);
+        nextBranch->_inputs.emplace_back(this);
+    }
+
+    bool isSortDiscovered() const { return _sortDiscovered; }
+    void setSortDiscovered(bool discovered) { _sortDiscovered = discovered; }
```
LucLabarriere, line 34: what's sortDiscovered?

v2/plan/PipelineBranches.cpp
```diff
@@ -0,0 +1,171 @@
...
+    while (!stack.empty()) {
+        auto [node, branch] = stack.top();
+        stack.pop();
+
+        PipelineBranch* currentBranch = branch;
+        
+        bool endOfBranch = false;
+
+        // If the current node is a binary node,
+        // put it in its own branch
```
LucLabarriere, line 60: Can you maybe add comment as to why it was better to add binary nodes in their own branch? If we support parallel branch execution some day, we might want to know if we can put the binary node in the next branch:
```
A --|
    | -> C  // For parallel execution, we might want C to contain the binary node
B --|
```

v2/plan/BranchGenerator.cpp
```diff
@@ -0,0 +1,582 @@
...
+{
+}
+
+BranchGenerator::~BranchGenerator() {
+}
+
+void BranchGenerator::translateBranch(PipelineBranch* branch) {
+    setupBranch(branch);
+
+    fmt::print("Branch\n");
```
LucLabarriere: Remove print

v2/plan/BranchGenerator.cpp
```diff
@@ -0,0 +1,582 @@
...
+    for (PlanGraphNode* node : branch->nodes()) {
+        fmt::print("Translate node {} {}\n", fmt::ptr(node), PlanGraphOpcodeDescription::value(node->getOpcode()));
```
LucLabarriere: Remove print

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -81,6 +85,9 @@ PipelineEdgeOutputInterface& PipelineBuilder::addGetOutEdges() {
     PipelineNodeInputInterface& input = getOutEdges->inNodeIDs();
     PipelineEdgeOutputInterface& output = getOutEdges->outEdges();
 
+    PipelineDumper dumper(_pipeline);
+    dumper.dumpMermaid(std::cout);
+    fmt::print("Processor @={} @={}", fmt::ptr(_pendingOutput.getInterface()->getPort()->getProcessor()), fmt::ptr(getOutEdges));
```
LucLabarriere, line 122: Remove print

v2/plan/BranchGenerator.cpp
```diff
@@ -0,0 +1,582 @@
...
+        _builder.setPendingOutputInterface(nullptr);
+        return;
+    }
+
+    // The first input branch gives the input of the builder
+    const PipelineBranch* firstInput = prevBranches.front();
+    _builder.setPendingOutputInterface(firstInput->getOutputInterface());
+
+    _currentBranch = branch;
+}
```
LucLabarriere, line 124: Do you think it would be possible to have `PipelineBuilder::newBranch()` and `PipelineBuilder::newBranchFromInput(firstInput)` for clarity, and to be able to define new branches cleanly from PipelineBuilder directly (for unit tests for example)

rjb32:
It's a DRAFT and debugging is in progress :)

v2/plan/PipelineBranch.h
rjb32: It's because branches are topologically sorted before translation. So a way to eliminate a map is to store the information physically in the structure

v2/plan/PipelineBranches.cpp
rjb32: Because when we define a branch, it's a straight line sequence of plan graph nodes. Which a binary node is not. A binary node to me is a node that connects 2 branches of the PlanGraph.

v2/plan/BranchGenerator.cpp
rjb32: I think there is a bit of confusion induced by me by calling it PipelineBranch. It's a branch of the plan graph. This is not something that is constructed by pipeline builder. This is a branch to be translated into the pipeline.

v2/plan/PipelineBranches.cpp
LucLabarriere: Yes but that's semantic concern only, technically speaking a binary node could be the first node of a branch, and for parallel execution it might be what we want to do. Adding a comment would remind ourselves why we didn't do it in the first place and if it's a good idea or not

v2/plan/BranchGenerator.cpp
LucLabarriere: No I see what you mean yeah... It's not confusing but when we construct a pipeline manually with the pipeline builder, it's a bit weird to have to do `builder.setPendingOutputInterface(nullptr)` when we want to add a new source.

And when we want to add a cartesian product, it also means we have to build both branches manually, then do `builder.setPendingOutputInterface(endOfBranchA.getOutputInterface());` Maybe it's fine though, but the concept of a branch also exists in the PipelineBuilder implicitly, so it would make sense to have methods that talk about branches

LucLabarriere:
> It's a DRAFT and debugging is in progress :)

Oh ok I thought it was ready!

PR #153, opened 2025-11-27
Add check in CartProdProc for rectangular input

cyrusknopf:
Soon we may be integrating Exprs or ColumnConst, so added a defensive check for a rectangular input.

### DIFF
--- a/v2/pipeline/processors/CartesianProductProcessor.cpp
+++ b/v2/pipeline/processors/CartesianProductProcessor.cpp
@@ -1,5 +1,7 @@
 #include "CartesianProductProcessor.h"
 
+#include <algorithm>
+
 #include "PipelineV2.h"
 #include "PipelinePort.h"
 #include "ExecutionContext.h"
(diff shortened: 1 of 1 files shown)


v2/pipeline/processors/CartesianProductProcessor.cpp
```diff
@@ -44,6 +46,18 @@ void verifyAllColumnVectors(Dataframe* df) {
     }
 }
 
+void verifyRectangular(const Dataframe* df) {
+    const size_t rowCount = df->getRowCount();
+    const bool rectangular = std::ranges::all_of(df->cols(), [rowCount](const NamedColumn* ncol) {
```
rjb32, line 51: Is it really simpler that writing a for loop?

cyrusknopf: In my opinion the intent is much clearer

rjb32: Fair enough

PR #154, opened 2025-12-03
Feature/fork processor

ghost:


v2/pipeline/PipelineBuilder.cpp
```diff
@@ -109,13 +108,6 @@ PipelineNodeOutputInterface& PipelineBuilder::addScanNodes() {
 }
 
 PipelineEdgeOutputInterface& PipelineBuilder::addGetOutEdges() {
-    // Create get out edges processor
```
rjb32, line 112: Can you clarify why all of these are removed in PipelineBuilder?

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -247,24 +227,35 @@ PipelineEdgeOutputInterface& PipelineBuilder::addGetEdges() {
...
-void PipelineBuilder::openMaterialize() {
-    _matProc = MaterializeProcessor::create(_pipeline, _mem);
-    _isMaterializeOpen = true;
-}
-
 bool PipelineBuilder::isSingleMaterializeStep() const {
     return _matProc->getMaterializeData().isSingleStep();
 }
 
+std::vector<PipelineBlockOutputInterface>& PipelineBuilder::addFork(uint8_t count) {
```
rjb32: Don't use a machine type for count but just size_t

v2/pipeline/processors/ForkProcessor.cpp
```diff
@@ -0,0 +1,59 @@
...
+    ForkProcessor* fork = new ForkProcessor(count);
+    PipelineInputPort* input = PipelineInputPort::create(pipeline, fork);
+    fork->_input.setPort(input);
+    fork->addInput(input);
+
+    for (size_t i = 0; i < fork->_outputs.size(); ++i) {
+        PipelineOutputPort* output = PipelineOutputPort::create(pipeline, fork);
+        fork->_outputs[i].setPort(output);
+        fork->addOutput(output);
+    }
```
rjb32, line 34: Why the prepare function is not called like for other processors?

v2/pipeline/processors/ForkProcessor.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    if (isFinished) {
+        finish();
+        return;
+    }
+
+    for (const auto& output : _outputs) {
+        // Copy the dataframes over - sharing buffers gets messy
+        // without having a specialised port.
+        output.getDataframe()->copyFrom(input().getDataframe());
```
rjb32, line 63: Hmm so we do a copy in fact

v2/pipeline/processors/ForkProcessor.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+ForkProcessor::ForkProcessor(uint8_t count)
+    :_outputs(count)
+{
+}
+
+ForkProcessor* ForkProcessor::create(PipelineV2* pipeline,
+                                     uint8_t count) {
+
+    ForkProcessor* fork = new ForkProcessor(count);
```
rjb32, line 25: We create the fork processor but nowhere do we register it into the Pipeline for ownership?

v2/pipeline/processors/ForkProcessor.h
```diff
@@ -0,0 +1,32 @@
...
+    static ForkProcessor* create(PipelineV2* pipeline, uint8_t count);
+
+    PipelineBlockInputInterface& input() { return _input; }
+    std::vector<PipelineBlockOutputInterface>& outputs() { return _outputs; }
+
+    void prepare(ExecutionContext* ctxt) override;
+    void reset() override;
+    void execute() override;
+
+    std::string describe() const final {
```
rjb32: You can put the implementation in the cpp file

v2/pipeline/processors/MaterializeProcessor.h
```diff
@@ -26,6 +33,9 @@ class MaterializeProcessor : public Processor {
 
     std::string describe() const override;
 
+    bool isConnected() const { return _connected; };
```
rjb32: Do we really need that bit of additional state? The isConnected predicate could be implemented by checking our inputs no?

v2/plan/PipelineGenerator.cpp
```diff
@@ -52,8 +54,9 @@ using namespace db::v2;
 namespace {
 
 struct TranslateNodeToken {
-    PlanGraphNode* _node {nullptr};
-    PipelineOutputInterface* _previousInterface {nullptr};
+    PlanGraphNode* node;
```
rjb32: Keep the default initialisation of pointers to nullptr

v2/plan/PipelineGenerator.cpp
```diff
@@ -431,9 +457,7 @@ PipelineOutputInterface* PipelineGenerator::translateEdgeFilterNode(EdgeFilterNo
 }
 
 PipelineOutputInterface* PipelineGenerator::translateProduceResultsNode(ProduceResultsNode* node) {
-    // If MaterializeNode has not been seen at that point,
-    // we materialize if we have data in flight
-    if (_builder.isMaterializeOpen() && !_builder.isSingleMaterializeStep()) {
+    if (!_builder.isSingleMaterializeStep()) {
```
rjb32, line 460: Is it really the same thing? We could have a case where it's not open after a node that materialises?

v2/pipeline/processors/ForkProcessor.cpp
ghost: Ah sorry I missed that!

v2/pipeline/processors/ForkProcessor.cpp
ghost: What do you mean exactly? From what I can see we there is nothing specific we need to prepare for each cycle like we do in the getOutEdges processor etc.

v2/pipeline/processors/ForkProcessor.cpp
ghost: Yes - at least with the current ports we have we will need to copy as sharing a buffer will break assumptions we have about execution - for example: If we are sharing the buffer between the fork input and the outputs, in our canExecute() check we can only execute a processor if the input has data and the outputs can be written to (they do not have data) which will never be true and cause our pipeline to break.

v2/pipeline/PipelineBuilder.cpp
ghost: I think we discussed that rather than waiting till next node that needs a materialize processor to recreate the processor once it has consumed, we can just recreate it as soon as it has been consumed in pipeline generator.

v2/pipeline/processors/MaterializeProcessor.h
ghost: Yep, I have removed this!

v2/plan/PipelineGenerator.cpp
ghost: Since we are opening the materialise as soon as it has been closed - we don't really need to check if materializeIsOpen().

v2/pipeline/processors/ForkProcessor.cpp
rjb32: The Processor::postCreate function

PR #155, opened 2025-12-05
Reduce ColumnVector resizes in ScanEdgesChunkWriter

cyrusknopf:
This approach could be applied to most other chunk writers

### DIFF
--- a/storage/iterators/ScanEdgesIterator.h
+++ b/storage/iterators/ScanEdgesIterator.h
@@ -70,7 +70,7 @@ private:
 
     TombstoneFilter _filter;
 
-    size_t _ptr;
+    size_t _ptr {0};
 
     void filterTombstones();
 };


storage/iterators/ScanEdgesIterator.cpp
```diff
@@ -95,42 +95,15 @@ void ScanEdgesChunkWriter::fill(size_t maxCount) {
...
-    if (_tgts) {
-        _tgts->clear();
-    }
-    if (_edgeIDs) {
-        _edgeIDs->clear();
-    }
-    if (_types) {
-        _types->clear();
-    }
+    _ptr = 0;
```
rjb32, line 98: It's probably a good idea to work on a local copy of _ptr in the loop and then update this->_ptr afterwards

PR #158, opened 2025-12-09
Integrate Hash Join Processor Into Pipeline Generator

ghost:


v2/pipeline/PipelineBuilder.cpp
```diff
@@ -52,11 +52,11 @@ void concatDataframeShape(LocalMemory* mem,
...
 void createHashJoinDataFrameShape(LocalMemory* mem,
                                   DataframeManager* dfMan,
-                                  Dataframe* leftSrc,
-                                  Dataframe* rightSrc,
+                                  const Dataframe* leftSrc,
+                                  const Dataframe* rightSrc,
                                   Dataframe* dest,
-                                  ColumnTag leftJoinKeyTag,
-                                  ColumnTag rightJoinKeyTag) {
+                                  const ColumnTag leftJoinKeyTag,
```
rjb32: Why putting const here if they are already passed by value?

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -332,18 +331,19 @@ PipelineBlockOutputInterface& PipelineBuilder::addProjection(std::span<Projectio
     return output;
 }
 
-PipelineBlockOutputInterface& PipelineBuilder::addHashJoin(PipelineBlockOutputInterface* rhs,
-                                                           ColumnTag leftJoinKey,
-                                                           ColumnTag rightJoinKey) {
+PipelineBlockOutputInterface& PipelineBuilder::addHashJoin(PipelineOutputInterface* rhs,
+                                                           const ColumnTag leftJoinKey,
```
rjb32: Same why putting const here

v2/plan/PipelineGenerator.cpp
```diff
@@ -503,7 +505,57 @@ PipelineOutputInterface* PipelineGenerator::translateProduceResultsNode(ProduceR
...
+            rightJoinTag = _declToColumn.find(node->getRightVarDecl())->second;
+            break;
+        }
+        case JoinType::COMMON_SUCCESSOR: {
+            leftJoinTag = lhs->getStream().visit(visitor);
+            rightJoinTag = rhs->getStream().visit(visitor);
+            break;
+        }
+        case JoinType::DIAMOND: {
+            throw PlannerException("Undirected Join Path On Common Successor Not Supported");
```
rjb32, line 579: Why is this not supported?

ghost: As discussed offline - the simplest way to implement this is a join + filter, so we are holding of on this until the filter pr is done.

v2/pipeline/EntityOutputStream.h
```diff
@@ -78,21 +78,23 @@ class EntityOutputStream {
...
-            _visitor(stream);
+        ReturnType operator()(const EdgeStream& stream) const {
+            return _visitor(stream);
         }
     };
 
-    void visit(const auto& visitor) const {
-        std::visit(StreamVisitor<decltype(visitor)> {visitor}, _stream);
+    auto visit(const auto& visitor) const {
+        return std::visit(StreamVisitor<decltype(visitor)> {visitor}, _stream);
```
cyrusknopf, line 97: Nice

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -332,7 +331,7 @@ PipelineBlockOutputInterface& PipelineBuilder::addProjection(std::span<Projectio
     return output;
 }
 
-PipelineBlockOutputInterface& PipelineBuilder::addHashJoin(PipelineBlockOutputInterface* rhs,
+PipelineBlockOutputInterface& PipelineBuilder::addHashJoin(PipelineOutputInterface* rhs,
```
cyrusknopf, line 339: Why is it no longer a block output interface as rhs?

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -357,6 +357,9 @@ PipelineBlockOutputInterface& PipelineBuilder::addHashJoin(PipelineBlockOutputIn
                                  leftJoinKey,
                                  rightJoinKey);
 
+    auto joinTag = outDf->cols().back()->getTag();
```
cyrusknopf, line 365: Can the output DF ever have no columns? If both inputs have no columns? (not sure that even makes sense)

v2/plan/PipelineGenerator.cpp
```diff
@@ -503,7 +505,57 @@ PipelineOutputInterface* PipelineGenerator::translateProduceResultsNode(ProduceR
...
+    PipelineOutputInterface* rhs = isBLhs ? inputA : inputB;
+
+    // LHS is implicit in @ref _pendingOutput
+    _builder.getPendingOutput().updateInterface(lhs);
+
+    const auto visitor = Overloaded {
+        [&](const EntityOutputStream::NodeStream& stream) -> ColumnTag {
+            return stream._nodeIDsTag;
+        },
+        [&](const EntityOutputStream::EdgeStream& stream) -> ColumnTag {
```
cyrusknopf: Is the total lambda capture necessary? Or can you narrow it to only what you need

v2/plan/ReadStmtGenerator.cpp
```diff
@@ -419,10 +419,16 @@ void ReadStmtGenerator::placeJoinsOnVars() {
...
-                return _tree->create<JoinNode>();
+                return _tree->create<JoinNode>(node->getVarDecl(),
+                                               node->getVarDecl(),
+                                               JoinType::COMMON_SUCCESSOR);
             }
+            case PlanGraphTopology::PathToDependency::UndirectedPath:
+                return _tree->create<JoinNode>(node->getVarDecl(),
+                                               node->getVarDecl(),
+                                               JoinType::DIAMOND);
+                throwError("Undirected Join Path On Common Successor Not Implemented");
```
cyrusknopf: Exception after return

v2/plan/ReadStmtGenerator.cpp
```diff
@@ -419,10 +419,16 @@ void ReadStmtGenerator::placeJoinsOnVars() {
...
             case PlanGraphTopology::PathToDependency::NoPath: {
-                return _tree->create<JoinNode>();
+                return _tree->create<JoinNode>(node->getVarDecl(),
+                                               node->getVarDecl(),
+                                               JoinType::COMMON_SUCCESSOR);
             }
+            case PlanGraphTopology::PathToDependency::UndirectedPath:
+                return _tree->create<JoinNode>(node->getVarDecl(),
+                                               node->getVarDecl(),
+                                               JoinType::DIAMOND);
```
cyrusknopf, line 429: Why do we pass `node->getVarDecl()` twice here?

v2/plan/PipelineGenerator.cpp
ghost: Yeah good point, I will review it!

v2/plan/ReadStmtGenerator.cpp
ghost: Good catch!

v2/plan/ReadStmtGenerator.cpp
ghost: This is to pass values as the left and right join columns - just to keep a common abstraction for cases when we will actually have different join columns (when we will have predicate joins).

v2/pipeline/PipelineBuilder.cpp
ghost: I don't think this is possible due to the nature of joins - this would imply that the left or right df is empty as well.

v2/pipeline/PipelineBuilder.cpp
ghost: I did this to match style with cartProd processor but also before this I wasn't using this function in the context of the Pipeline Generator so didn't know what the rhs type would actually be in practice.

cyrusknopf: Hmm I didn't realise I used PendingOutputInterface as argument to addCartesianProduct. What do you think we should use?

ghost: I think PendingOutputInterface makes sense to use here, as it is the pointer of this type that we are passing around pipeline builder and generator. casting to PipelineBlockOutputInterface is unnecessary as we have virtual methods that handle all the specifics.

PR #159, opened 2025-12-11
Use exception in bioassert and require mandatory message

rjb32:
Raise FatalException in bioassert instead of abort.
Execute bioassert both in release and debug mode.

### DIFF
--- a/storage/dump/CommitLoader.h
+++ b/storage/dump/CommitLoader.h
@@ -82,8 +82,8 @@ public:
 
         // Reading journal
         {
-            CommitJournal& journal = *commit->_data->_history._journal;
-            bioassert(&journal, "invalid journal"); // Should be initialised in commit constructor
+            CommitJournal* journal = commit->_data->_history._journal.get();
+            bioassert(journal, "invalid journal"); // Should be initialised in commit constructor
 
             const fs::Path journalPath = path / "journal";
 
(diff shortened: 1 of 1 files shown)


storage/dump/CommitLoader.h
```diff
@@ -81,7 +83,7 @@ class CommitLoader {
         // Reading journal
         {
             CommitJournal& journal = *commit->_data->_history._journal;
-            bioassert(&journal); // Should be initialised in commit constructor
+            bioassert(&journal, "invalid journal"); // Should be initialised in commit constructor
```
LucLabarriere: Wait does this work? Can we dereference a nullptr and check if &ptr is not nullptr?

cyrusknopf: No you're right I think this doesn't work

rjb32: It does not dereference a pointer from a compiler point of view, &journal just gives the address which is already the value of the reference for the compiler. After code generation a reference and a pointer are implemented the same, it's just an address

rjb32: It just looks weird

LucLabarriere: I mean we dereference first: `CommitJournal& journal = *...` then check if the address is valid with `assert(&journal)`

rjb32: Ah yes that's not ideal I will fix

PR #162, opened 2025-12-12
Fix Hash Join Processor Tests

ghost:
The disabled test was actually failing because it wasn't implemented correctly. While initially debugging it I saw the issue with the pipeline execution (parallel pipelines with no fork connection, do not run in parallel due to how we execute the pipeline) and realised that it wasn't testing what I wanted it to test. I mistakenly assumed the test was broken because of this, but that was not the case!


ghost:
I also noticed a naming issue in another test that I fixed in this PR as well.

PR #163, opened 2025-12-12
C++ VectorDatabase using Faiss. Uses LSH partitioning and Shard cache system

LucLabarriere:
# C++ VectorDatabase using Faiss

Only used in a samples/vector-db for now.

## Features
- A database of vector libraries (mapped by a unique name)
- LSH partitioning that produces independent shards
- Shard cache system to limit RAM usage
- Brute force indexes (trivial to switch to HNSW later)

## Thread safety model
- `VectorDatabase` is behind with a `shared_mutex`
- `StorageManager` is behind a `shared_mutex`
- `VecLibShard` is behind a `mutex`
- `VecLib` is behind a `shared_mutex`

We call `VectorDatabase::getLibrary(name)` to retrieve a `VecLibAccessor` which is a wrapper around a `std::shared_lock` on the `VecLib` mutex. This means that multiple transactions can occur on the same vector library. If the library is deleted, the deletion acquires a unique lock on the `VecLib` so that we must way for all operations to finish before deleting.

The `VecLibAccessor` allows to call `addEmbeddings` and `search`. It doesn't require a `unique_lock` because the operations are performed on the shards, not the `VecLib` itself.

## Future considerations:
- Explore LSH partitioning using higher Hammet distances (if we couldn't find enough results)
- Automatically switch to HNSW if the index becomes too big

### DIFF
--- a/vector/CMakeLists.txt
+++ b/vector/CMakeLists.txt
@@ -1,12 +1,3 @@
-# faiss
-find_package(Faiss REQUIRED)
-
-# threads
-find_package(Threads REQUIRED)
-
-# OpenMP (required by Faiss)
-find_package(OpenMP REQUIRED)
-
 set(VECTOR_SOURCES
         RandomGenerator.cpp
         StorageManager.cpp
--- a/vector/LSHShardRouter.cpp
+++ b/vector/LSHShardRouter.cpp
@@ -1,7 +1,8 @@
 #include "LSHShardRouter.h"
 
-#include "BioAssert.h"
 #include "RandomGenerator.h"
+
+#include "BioAssert.h"
 
 using namespace vec;
 
--- a/vector/LSHShardRouter.h
+++ b/vector/LSHShardRouter.h
@@ -29,8 +29,8 @@ private:
     friend class LSHShardRouterLoader;
     friend class LSHShardRouterWriter;
 
-    size_t _dim;
-    uint8_t _nbits;
+    size_t _dim {};
+    uint8_t _nbits {};
     std::vector<std::vector<float>> _hyperplanes;
 };
 
--- a/vector/StorageManager.cpp
+++ b/vector/StorageManager.cpp
@@ -1,9 +1,10 @@
 #include "StorageManager.h"
 
 #include <charconv>
+#include <mutex>
+
 #include <faiss/IndexFlat.h>
 #include <faiss/index_io.h>
-#include <mutex>
 
 #include "LSHShardRouter.h"
 #include "VecLibStorage.h"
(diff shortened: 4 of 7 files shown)


cyrusknopf:
Is it worth trimming faiss? For example do we need `tutorial/`?

LucLabarriere:
> Is it worth trimming faiss? For example do we need `tutorial/`?

I thought about this yes, not sure if it's worth spending the time though, it was supposed to be a submodule but I had to manually change the cmake_minimum_required statement to version 3.10, I think we will later add it as submodule anyway

vector/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+# faiss
+find_package(Faiss REQUIRED)
```
rjb32: Please move the find_package to the root CMakeLists.txt

vector/LSHShardRouter.h
```diff
@@ -0,0 +1,37 @@
...
+    void initialize();
+
+    LSHSignature getSignature(std::span<const float> vector) const;
+    void getSearchSignatures(std::span<const float> vector, std::vector<LSHSignature>& signatures) const;
+
+private:
+    friend class LSHShardRouterLoader;
+    friend class LSHShardRouterWriter;
+
+    size_t _dim;
```
rjb32: Default initialize class members

vector/LSHShardRouter.cpp
```diff
@@ -0,0 +1,65 @@
+#include "LSHShardRouter.h"
+
+#include "BioAssert.h"
+#include "RandomGenerator.h"
```
rjb32, line 3: Fix order of includes

vector/StorageManager.cpp
```diff
@@ -0,0 +1,211 @@
+#include "StorageManager.h"
+
+#include <charconv>
+#include <faiss/IndexFlat.h>
+#include <faiss/index_io.h>
```
rjb32, line 7: Reorder includes

vector/VecLib.cpp
```diff
@@ -0,0 +1,168 @@
+#include "VecLib.h"
+
+#include <faiss/Index.h>
+#include <faiss/IndexFlat.h>
+#include <faiss/index_io.h>
+#include <mutex>
```
rjb32: Put std includes first

vector/VecLib.h
```diff
@@ -0,0 +1,139 @@
+#pragma once
+
+
```
rjb32: Too many blank lines

rjb32:
Looks very good to me. Great work

PR #164, opened 2025-12-12
Migrate to use GetPropertyWithNull by default

cyrusknopf:
Context:
- Filters require consistent columns of values to properly evaluate predicates
- GetPropertiesProcessors cannot be chained together as they indirectly filter their output
- We must now use GetPropertiesWithNull by default so that the entire Entity column is maintained while chaining GetProperties prior to a filter

Changes:
- Read statements now use GetPropertiesWithNull instead of GetProperties
- Update tests to expect the above change

- PipelineBuilder now passes GetPropertiesWithNull the correct tag, which may be from the Var node (in case of projection, etc.) or may be from the incoming stream (in most other cases)

### DIFF
--- a/v2/plan/ReadStmtGenerator.cpp
+++ b/v2/plan/ReadStmtGenerator.cpp
@@ -27,6 +27,7 @@
 #include "nodes/GetOutEdgesNode.h"
 #include "nodes/GetPropertyNode.h"
 #include "nodes/GetEntityTypeNode.h"
+#include "nodes/GetPropertyWithNullNode.h"
 #include "nodes/JoinNode.h"
 #include "nodes/ProcedureEvalNode.h"
 #include "nodes/ProduceResultsNode.h"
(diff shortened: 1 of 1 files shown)


v2/plan/PipelineGenerator.cpp
```diff
@@ -432,7 +433,22 @@ PipelineOutputInterface* PipelineGenerator::translateGetPropertyWithNullNode(Get
...
+        if (stream.isNodeStream()) {
+            entityTag = stream.asNodeStream()._nodeIDsTag;
+        } else if (stream.isEdgeStream()) {
+            entityTag = stream.asEdgeStream()._edgeIDsTag;
+        }
+        else {
+            throw FatalException("Attempted to add GetPropertiesWithNull to pipeline "
+                                 "with unkown entity and empty stream.");
+        }
+    }
```
cyrusknopf, line 451: @rjb32 @LucLabarriere @suhassn-turing-bio Would you guys be able to double check my logic here to make sure it looks correct? The rest of the PR is just altering tests. thanks!

LucLabarriere: It looks good to me. Did you test it? I assume the RETURN statement will trigger the first case, and the filters will trigger the second case. You can always try in the PipelineGenerator to disable the filters (don't do anything when you translate the Filter nodes and run a query that triggers the second case:
```
MATCH (n) RETURN n.name // Triggers if branch
MATCH (n) WHERE n.name = "Cyrus" RETURN n.name // Triggers else branch
```

cyrusknopf: Yeah exactly, if branch is for return projections and otherwise it should use the stream. This branch doesn't have filters, so I've tested insofar as the current query tests, but not with filters.

PR #165, opened 2025-12-12
v2 Implement LOAD_GRAPH

rjb32:


v2/analyzer/CypherAnalyzer.cpp
```diff
@@ -163,6 +172,13 @@ void CypherAnalyzer::analyze(Limit* limitSt) {
     }
 }
 
+void CypherAnalyzer::analyze(const LoadGraphQuery* loadGraph) {
+    std::string_view graphName = loadGraph->getGraphName();
+    if (graphName.empty()) {
```
ghost, line 177: We should have the special character check like we have in v1?

rjb32: Yes thank you good point

PR #167, opened 2025-12-15
Add LIST GRAPH command

ghost:


v2/plan/nodes/ListGraphNode.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "nodes/PlanGraphNode.h"
```
LucLabarriere, line 3: Blank line after include

test/v2/queries/QueriesTest.cpp
```diff
@@ -1473,6 +1473,22 @@ TEST_F(QueriesTest, scanNodesByLabel) {
...
+
+    _db->queryV2("list graph", _graphName, &_env->getMem(), [&](const Dataframe* df) -> void {
+        ASSERT_TRUE(df != nullptr);
+        ASSERT_EQ(df->cols().size(), 1);
+        ASSERT_EQ(df->getRowCount(), expectedGraphNames.size());
+        const auto& cols = df->cols();
+        auto* actualGraphNames = cols.at(0)->as<ColumnVector<std::string_view>>();
+
+        EXPECT_EQ(expectedGraphNames.getRaw(), actualGraphNames->getRaw());
+    });
```
cyrusknopf, line 1604: I think we should start enforcing checking the result of `queryV2`. If the query fails I think this test will pass because the callback is never executed and there is no assertion/expect after the query. in `WriteQueriesTest.cpp` I have some helpers defined at the top that make checking the result of query a bit less painful.

v2/AST/ListGraphQuery.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+using namespace db::v2;
+
+ListGraphQuery::ListGraphQuery(DeclContext* declContext)
+    : QueryCommand(declContext)
+{
+}
+
+ListGraphQuery::~ListGraphQuery() {
+}
```
cyrusknopf, line 14: I think we should stop this pattern (we have all done it), but it is objectively worse in my opinion. If we instead declare all empty constructors/destructors as `=default` in the header we allow our classes to be trivially cons-/des-tructible, whilst with this pattern I don't think they are.

@rjb32 @LucLabarriere what do you think?

v2/pipeline/processors/ListGraphProcessor.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <spdlog/fmt/fmt.h>
+
+using namespace db::v2;
+
+ListGraphProcessor::ListGraphProcessor()
+{
+}
+
+ListGraphProcessor::~ListGraphProcessor() {
+}
```
cyrusknopf, line 18: Same comment as above on declaring `=default` in header instead, but see what the others think

v2/pipeline/processors/ListGraphProcessor.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include "Processor.h"
+
+#include "interfaces/PipelineBlockInputInterface.h"
+#include "interfaces/PipelineValueOutputInterface.h"
+
+namespace db::v2 {
+
+class ListGraphProcessor : public Processor {
```
cyrusknopf: Class and methods can be final

v2/pipeline/PipelineBuilder.cpp
```diff
@@ -602,6 +603,20 @@ PipelineBlockOutputInterface& PipelineBuilder::addWrite(const WriteProcessor::De
     return output;
 }
 
+PipelineValueOutputInterface& PipelineBuilder::addListGraph() {
+    ListGraphProcessor* loadGraph = ListGraphProcessor::create(_pipeline);
+
+    PipelineValueOutputInterface& output = loadGraph->output();
+    Dataframe* df = output.getDataframe();
+    NamedColumn* graphNameValue = allocColumn<ColumnVector<types::String::Primitive>>(df);
+    graphNameValue->rename("graph_name");
```
LucLabarriere: Actually, maybe we could setup a convention on the column naming, I used "Change ID" in my PR but I think it's bad, I'll change it. Do we prefer snake case or camel case?

test/v2/queries/QueriesTest.cpp
ghost: Ah yes I missed this, I initially designed the `EXPECT_EQ` to happen outside the callback. Good catch!

v2/AST/ListGraphQuery.cpp
rjb32: This is a coding style that I explicitly mandate in the guidelines. Basically it's when a class has non-trivial members such as STL data structures (std::vector, std::map, std::unordered_map...etc) or anything non-trivial to destruct, I strongly suggest to nearly always put the constructor and destructor in the cpp file to avoid inlining in the binary of a ton of STL code that we don't control.

ghost: I've never fully agreed with this style but have gone along with it where it is used in other similar use cases. I'd like to here Remy and Luc's opinions on this as well

v2/pipeline/processors/ListGraphProcessor.h
rjb32: Be careful of pondering the added benefit of "final" and the additional syntactic noise that it adds. It's not something that I encourage in the guidelines in general.

v2/pipeline/PipelineBuilder.cpp
LucLabarriere: I used camel case in my procedures PR: nodeCount, edgeCount, etc. So graphName? What do you think? Neo4j does camel case too

v2/AST/ListGraphQuery.cpp
LucLabarriere: I have to say it bothers me a lot too in certain cases like this one, ListGraphQuery is basically empty. It's the child of QueryCommand which is just a typed wrapper around DeclContext*. So what should just be an 8 byte copy turns into many virtual function calls everywhere it's constructed/destructed. I profiled a bit the performances when I did the parser and switching from 100% header vs 100% cpp had a measurable impact on the perfs.

In this case it didn't matter too much because bison is so slow that the performances are mostly governed by the parsing, but still, it's painful to write code that we know hurt performances for close to no trade off in this case (no STL structures)

cyrusknopf: As a counter point Luc, it may be difficult to discriminate cases where the class is or is not trivial, particularly when there is a lot of inheritance (for example any Processor as a lot of inherited members), and you have to track all the way up the inheritance hierarchy and check. In that sense it is a safe guideline to always define in the CPP to avoid the larger binary.

And that is not even mentioning the fact that a class' "triviality" can change, perhaps adding an STL container member, and then we would potentially have to alter a lot of ctor/dtor definitions.

v2/pipeline/processors/ListGraphProcessor.h
cyrusknopf: I wasn't aware this was covered by the guidelines; I can stop using/commenting it if you would rather we avoid it

v2/AST/ListGraphQuery.cpp
LucLabarriere: Yeah this is why it's probably fine, that's the kind of things we will need to profile when we do an optimisation pass

v2/pipeline/PipelineBuilder.cpp
ghost: Im happy with camel case, especially if we are following convention. What do you think @rjb32 ?

v2/AST/ListGraphQuery.cpp
rjb32: Ah yes for ListGraphQuery I agree it's not necessary, as it is trivial indeed.

v2/pipeline/PipelineBuilder.cpp
rjb32: Camel case is good indeed

PR #169, opened 2025-12-15
Feature/add create graph

ghost:


v2/plan/nodes/CreateGraphNode.h
```diff
@@ -6,16 +6,16 @@ namespace db::v2 {
 
 class CreateGraphNode : public PlanGraphNode {
 public:
-    explicit CreateGraphNode(const std::string& graphName)
+    explicit CreateGraphNode(const std::string_view graphName)
```
LucLabarriere: remove const specifier

v2/AST/CreateGraphQuery.cpp
```diff
@@ -0,0 +1,24 @@
...
+CreateGraphQuery::CreateGraphQuery(DeclContext* declContext,
+                                   std::string_view name)
+    : QueryCommand(declContext),
+    _graphName(name)
+{
+}
+
+CreateGraphQuery::~CreateGraphQuery() {
+}
+
```
cyrusknopf, line 17:
```suggestion
CreateGraphQuery::~CreateGraphQuery()
{
}
```

v2/pipeline/processors/CreateGraphProcessor.cpp
```diff
@@ -0,0 +1,62 @@
...
+    return fmt::format("CreateGraphProcessor @={}", fmt::ptr(this));
+}
+
+void CreateGraphProcessor::prepare(ExecutionContext* ctxt) {
+    _ctxt = ctxt;
+
+    markAsPrepared();
+}
+
+void CreateGraphProcessor::reset() {
```
cyrusknopf, line 45: Should we have a `markAsReset()` here?

v2/AST/CreateGraphQuery.cpp
LucLabarriere: We put braces on next line only after constructors! ahah

cyrusknopf: Wait not destructors too??

cyrusknopf: I have been doing it for destructors as well this whole time I think 😭 my bad

LucLabarriere: No ahah, I know sometimes I have to double check too

LucLabarriere: Quickly checked in the code base with grep + git blame, we all did it at least once in the codebase (yes including Remy!)

v2/pipeline/processors/CreateGraphProcessor.cpp
ghost: shouldn't really matter in this case, as `markAsReset()` resets the processor for the next cycle.

PR #170, opened 2025-12-15
New query v2: LOAD NEO4J "..." AS ...

LucLabarriere:
Two possibilities:
- `LOAD NEO4J "reactome.dump" AS reactome`
- `LOAD NEO4J "reactome.dump"`: implicitely uses the basename of the file as graph name


v2/pipeline/processors/LoadNeo4jProcessor.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <string_view>
+
+#include "Path.h"
```
rjb32: Fix order of includes: parent class header must come first

PR #172, opened 2025-12-15
V2 LOAD GML

rjb32:
Syntax:
```
LOAD GML 'complaints.gml' AS complaints
```

### DIFF
--- a/samples/plan2-test-generator/main.cpp
+++ b/samples/plan2-test-generator/main.cpp
@@ -56,7 +56,7 @@ void runPlan2(std::string_view query) {
         turingDir.rm();
     }
 
-    TuringConfig config = TuringConfig::createDefault();
+    TuringConfig config;
     config.setTuringDirectory(turingDir);
     TuringDB db(&config);
     db.run();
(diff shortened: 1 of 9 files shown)


v2/parser/CypherParser.y
```diff
@@ -317,12 +320,17 @@ singleQuery
...
     | changeQuery { $$ = $1; }
+    | loadGML { $$ = $1; }
     ;
 
 loadGraph
     : LOAD GRAPH ID { $$ = LoadGraphQuery::create(ast, $3); LOC($$, @$); }
     ;
 
+loadGML
+    : LOAD GML STRING_LITERAL AS ID { $$ = LoadGMLQuery::create(ast, $5, $3); LOC($$, @$); }
```
LucLabarriere: In my PR I made the AS keyword optional:
```
- LOAD GML "..."
- LOAD GML "..." AS ...
```

Using basename of the file as the graph name by default, I made the `LoadGraphQuery` constructor and `create` methods take only a single argument by default (the file path), the graph name being set by a setGraphName method

LucLabarriere: It only requires something like this in CypherAnalyzer:
```c++
std::string_view graphName = loadNeo4j->getGraphName();
if (graphName.empty()) {
    graphName = loadNeo4j->getFilePath().basename();
}
```

rjb32: Good choice thanks

PR #178, opened 2025-12-17
Make turingdb -cli by default

rjb32:
`./turingdb` Runs interactively in CLI mode by default and runs the HTTP server
`./turingdb -demon` Spawns the server as a daemon in the background

### DIFF
--- a/regress/serialization/main.py
+++ b/regress/serialization/main.py
@@ -20,8 +20,7 @@ def spawn_turingdb():
 
 def stop_turingdb(proc):
     print(f"- {GREEN}Stopping turingdb{NC}")
-    proc.send_signal(signal.SIGTERM)
-    proc.wait()
+    subprocess.check_call('pkill turingdb', shell=True)
 
 
 def wait_ready(client):


LucLabarriere:
I'm really not convinced about this change, so close to the next release, this can break many scripts to spawn instances here and there. It breaks the regression tests that I'm working on. Plus, I don't even think -cli was a good idea in the first place, it makes us interact with turingdb in a mode that clients will barely ever used (it's mostly for testing of having fun from time to time, if we think about SQL, nobody uses the console except for debug). If we want to test in an interact shell environment, turingsh is the way to go since it actually tests the pipeline the clients go through. I assume this change makes sense in the context of embedded mode turingdb, but I don't see how we can achieve that for the next release anyway

rjb32:
I made the change to run the HTTP server as well when we do ./turingdb and we get the CLI as well to inject directly queries in the server if we want to.

rjb32:
As we discussed, this is to fix the surprising behavior for a first time user that tries ./turingdb and just sees something disappearing in the background because of the demonize by default. Now the CLI opens and the server is launched.
If the user wants a daemon explicitly, they can do ./turingdb -demon

PR #185, opened 2025-12-30
Implement greater/less than (or equal) operators for filters

cyrusknopf:
Context:
- Filters use `ExprProgram` which currently does not support `<`, `>`, `<=`, `>=`

Changes:
- Add column operators for `<`, `>`, `<=`, `>=` through a generic predicate interface
- Add support for `<`, `>`, `<=`, `>=` in `ExprProgram`
- Add tests for integer filters using the new operands

- Remove dead code leftover from v1, refactor functions to minimise repetition

### DIFF
--- a/storage/columns/ColumnOperators.h
+++ b/storage/columns/ColumnOperators.h
@@ -435,7 +435,7 @@ private:
     template <typename Pred, typename T, typename U>
         requires OptionallyComparable<T, U>
               && std::predicate<Pred, unwrap_optional_t<T>, unwrap_optional_t<U>>
-    static std::optional<bool> optionalPredicate(const T& a, const U& b) {
+    inline static std::optional<bool> optionalPredicate(const T& a, const U& b) {
         if constexpr (is_optional_v<T>) {
             if (!a.has_value()) {
                 return std::nullopt;


storage/columns/ColumnOperators.h
```diff
@@ -43,173 +45,132 @@ struct unwrap_optional<std::optional<U>> {
 template <typename T>
 using unwrap_optional_t = typename unwrap_optional<T>::underlying_type;
 
+// Types that may be compared, but one or both may be wrapped in optional
 template <typename T, typename U>
 concept OptionallyComparable =
+    (Stringy<unwrap_optional_t<T>, unwrap_optional_t<U>>
+     || std::totally_ordered_with<unwrap_optional_t<T>, unwrap_optional_t<U>>);
+
+// Types that are semantically equivalent, but one or both may bew rapped in optional
```
rjb32: Typo in the comment: "may bew"

storage/columns/ColumnOperators.h
```diff
@@ -473,9 +427,15 @@ class ColumnOperators {
...
+    /**
+     * @brief Generic function to apply a predicate to two possibly-optional operands,
+     * where either operand being nullopt results in the final result being nullopt, and the
+     * result of applying the operator otherwise.
+     */
+    template <typename Pred, typename T, typename U>
         requires OptionallyComparable<T, U>
-    static std::optional<bool> optionalEq(const T& a, const U& b) {
+              && std::predicate<Pred, unwrap_optional_t<T>, unwrap_optional_t<U>>
+    static std::optional<bool> optionalPredicate(const T& a, const U& b) {
```
rjb32: It would be a good idea to put all of them as inline just in case, because they are called in each of the "functionName" instantiations

storage/columns/ColumnOperators.h
cyrusknopf: Thanks fixed

storage/columns/ColumnOperators.h
cyrusknopf: Marked all as inline

PR #190, opened 2026-01-06
Add wine_ontology regress case

rjb32:
https://turingbio.atlassian.net/browse/TURING-911


rjb32:
FYI @maxvincent24 I added your wine ontology as a test case

PR #194, opened 2026-01-07
TURING-911 Fix stack-smashing error when creating too many labels

cyrusknopf:
Relevant Jira: https://turingbio.atlassian.net/browse/TURING-911

Context:
- When creating labels whose bit representation exceeded the capacity of a label set, we write OoB
- This causes stack smashing

Changes:
- Added bounds check on the bit index of the label to create, and guarded `LabelSet::computBitShift` with assertions

### DIFF
--- a/query/pipeline/processors/WriteProcessor.cpp
+++ b/query/pipeline/processors/WriteProcessor.cpp
@@ -307,7 +307,7 @@ void WriteProcessor::createNodes(size_t numIters) {
 
         std::vector<NodeID>& raw = col->getRaw();
         // Old size should always be 0 (asserted above), but add for safety 
-        size_t oldSize = raw.size();
+        const size_t oldSize = raw.size();
         raw.resize(oldSize + numIters);
         std::iota(raw.begin() + oldSize, raw.end(), nextNodeID);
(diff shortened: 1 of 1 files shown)


query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -301,13 +301,15 @@ void WriteProcessor::createNodes(size_t numIters) {
...
-
         // Populate the output column for this node with the index in the CWB which it
         // appears. These indexes are later transformed into "fake IDs" (an estimate as to
         // what the NodeID will be when it is committed) in @ref postProcessTempIDs.
-        raw.resize(raw.size() + numIters);
-        std::iota(col->begin(), col->end(), nextNodeID);
+
+        std::vector<NodeID>& raw = col->getRaw();
+        // Old size should always be 0 (asserted above), but add for safety 
+        size_t oldSize = raw.size();
```
rjb32: oldSize could be const. Where is the assertion you said was above?

cyrusknopf: Fixed. Assertion is on line 263

PR #196, opened 2026-01-07
Implement `IS (NOT) NULL` for filters

cyrusknopf:
Changes:
- Add a generic `PropertyNull` class which wraps `std::nullopt_t` as `std::nullopt`
- Allocate `ColumnConst<PropertyNull>` to denote comparisons to NULL
- Add specialisations for `equal` and `notEqual` which are used to calculate if NULL

### DIFF
--- a/storage/metadata/PropertyNull.h
+++ b/storage/metadata/PropertyNull.h
@@ -2,6 +2,11 @@
 
 #include <optional>
 
+/**
+ * @brief Default-initialisable wrapper for std::nullopt_t, for use in @ref
+ * ColumnConst<PropertyNull>, required for IS NULL or IS NOT NULL operations.
+ */
 struct PropertyNull {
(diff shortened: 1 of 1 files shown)


query/parser/CypherParser.y
```diff
@@ -736,8 +736,8 @@ comparisonSign
     | GT { $$ = BinaryOperator::GreaterThan; }
     | LT { $$ = BinaryOperator::LessThan; }
     | NOT_EQUAL { $$ = BinaryOperator::NotEqual; }
-    | IS { $$ = BinaryOperator::Equal; }
-    | IS_NOT { $$ = BinaryOperator::NotEqual; }
+    | IS { $$ = BinaryOperator::Equal; } // FIXME
```
rjb32: Remove FIXME

cyrusknopf: Fixed

PR #198, opened 2026-01-08
Turing-929: Fixed neo4j import bug by splitting the import in two changes

LucLabarriere:
# The issue

Currently when we import a graph, we:

- Create a new change
- For each node "chunk", create a new `DataPartBuilder` and create the nodes there
- Build all node dataparts
- For each edge "chunk", create a new `DataPartBuilder` and create the edges there
- Build all edge dataparts
- Submit the change

The problem is that the dataparts built in parallel are "invalid", they are not aware of other dataparts. e.g. the 5th node datapart has firstNodeID = 0.

Ideally, I think that if we want to write in parallel, we need to do that in separate changes. The problem is that at the time of edge creation, we need to use the `_tmpToFinalNodeIDs` map (in `DataPart`) which is not rebased, and thus, invalid.

> [!WARNING]
> Previous implementation: import the nodes then the edges in two separate changes, and disable parallel import so that we don't need to rebase, meaning that the `_tmpToFinalNodeIDs` don't become invalid.

New implementation: @rjb32  import nodes and edges in a single commit (single datapart). This allows to get rid of the tmpToFinalNode/EdgeIDs maps. CKG cannot be imported on my machine (40GB of RAM required during the buiding up of the `DataPartBuilder` then the process gets killed once we call `DataPart::load`. The memory usage is probably ~90GB at this stage.

### DIFF
--- a/import/neo4j/Neo4jImporter.cpp
+++ b/import/neo4j/Neo4jImporter.cpp
@@ -401,7 +401,7 @@ bool Neo4jImporter::importJsonDirImpl(JobSystem& jobSystem,
     {
         ChangeAccessor accessor = change->access();
 
-        if (auto res = change->submit(jobSystem); !res) {
+        if (auto res = accessor.submit(jobSystem); !res) {
             spdlog::error("Could not submit change: {}", res.error().fmtMessage());
             return false;
         }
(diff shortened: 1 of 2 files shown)


storage/DataPart.h
```diff
@@ -80,9 +80,6 @@ class DataPart {
     NodeID _firstNodeID {0};
     EdgeID _firstEdgeID {0};
 
-    // TODO Remove for the new Unique ID system
```
rjb32, line 83: Very good

rjb32 (approved):

PR #206, opened 2026-01-09
Implement more unit tests for filters with predicates

rjb32:
Add more unit tests for filters with: 
* WHERE predicates and notably nested parenthesis
* IS NULL and IS NOT NULL predicates


cyrusknopf:
The general structure of the tests looks good

PR #211, opened 2026-01-12
TURING-931 Fixed bug with WriteProcessor when handling multiple chunks of input

cyrusknopf:
[Jira](https://turingbio.atlassian.net/browse/TURING-931)

Context:
- Previous incorrect assumption about chunking behaviour of the `WriteProcessor` resulted in failing assertions

Changes:
- Add logic to clear the output of the `WriteProcessor` before each call to `execute()`, meaning we can always write up to a chunk (since the input is at most one chunk, we never write more than a chunk as output)
- Add similar logic to `FilterProcessor`, so empty chunks are only emitted if input is closed
- Add a unit test to verify behaviour

### DIFF
--- a/query/pipeline/processors/WriteProcessor.cpp
+++ b/query/pipeline/processors/WriteProcessor.cpp
@@ -24,7 +24,6 @@
 #include "versioning/Transaction.h"
 #include "dataframe/Dataframe.h"
 #include "WriteProcessorTypes.h"
-#include "views/PropertyView.h"
 #include "writers/MetadataBuilder.h"
 
 #include "FatalException.h"
@@ -229,11 +228,12 @@ void WriteProcessor::createNodes(size_t numIters) {
     LabelSet lblset;
     for (const WriteProcessorTypes::PendingNode& node : _pendingNodes) {
         // This is checked for validity in the call to @ref setup
-        auto* col = outDf->getColumn(node._tag)->as<ColumnNodeIDs>();
+        auto* createdNodeCol = outDf->getColumn(node._tag)->as<ColumnNodeIDs>();
 
         // Always empty on first call, or has been cleared by @ref setup.
-        bioassert(col->size() == 0, "Pending nodes column should be empty but is size {}",
-                  col->size());
+        bioassert(createdNodeCol->size() == 0,
+                  "Pending nodes column should be empty but is size {}",
+                  createdNodeCol->size());
 
         const std::span labels = node._labels;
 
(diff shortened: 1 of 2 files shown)


query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -246,21 +228,12 @@ void WriteProcessor::createNodes(size_t numIters) {
...
-                            node._tag.getValue()));
-        }
-        auto* col = _output.getDataframe()->getColumn(node._tag)->as<ColumnNodeIDs>();
-        if (!col) { // @ref as<> performs dynamic_cast
-            throw FatalException(fmt::format("Column {} was marked as a column of"
-                                             " pending nodes, but is not a NodeID"
-                                             " column.", node._tag.getValue()));
-        }
+        // This is checked for validity in the call to @ref setup
+        auto* col = outDf->getColumn(node._tag)->as<ColumnNodeIDs>();
```
rjb32: Choose better naming for this column

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -306,11 +279,12 @@ void WriteProcessor::createNodes(size_t numIters) {
         // what the NodeID will be when it is committed) in @ref postProcessTempIDs.
 
         std::vector<NodeID>& raw = col->getRaw();
-        // Old size should always be 0 (asserted above), but add for safety 
         const size_t oldSize = raw.size();
         raw.resize(oldSize + numIters);
         std::iota(raw.begin() + oldSize, raw.end(), nextNodeID);
         nextNodeID += numIters;
+
+        _wroteRows |= raw.size() > oldSize;
```
rjb32: What does this mean? Please explain or simplify.

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -321,21 +295,12 @@ void WriteProcessor::createEdges(size_t numIters) {
...
 
     for (const WriteProcessorTypes::PendingEdge& edge : _pendingEdges) {
-        if (!outDf->hasColumn(edge._tag)) {
-            throw FatalException(
-                fmt::format("Could not get column in WriteProcessor "
-                            "output dataframe for PendingEdge with tag {}.",
-                            edge._tag.getValue()));
-        }
-
+        // This is checked for validity in the call to @ref setup
```
rjb32, line 298: Same remark, choose better variable name for col

query/pipeline/processors/WriteProcessor.cpp
cyrusknopf: Whether or not the processor has written rows on this call to execute. Would `_writtenToOutput` be clearer?

query/pipeline/processors/WriteProcessor.cpp
cyrusknopf: Renamed to `createdNodeCol`

query/pipeline/processors/WriteProcessor.cpp
cyrusknopf: Renamed to `createdEdgeCol`

PR #212, opened 2026-01-12
Make compatibility changes for building on macos

rjb32:
Quite a lot of adaptations everywhere to make turingdb build on macos with the help of Claude Code.

* net/ http server: use kpoll on macos, enrich the Utils class for managing sockets and socket flags to support both platforms
* io/fs: make adaptations for Apple using ifdef/ifndef
* AlignedBuffer.h: fix alignment because we need 16k alignement on Apple Silicon
* Various header includes fixes and compilation fixes to pass compilation with clang on macos
* More type annotations are needed by Clang
* OpenMP needs to be provided explicitly to FAISS because it's not provided by default with Clang
* Fix bug in FilePageReader and FilePageWriter
* Install LLVM from Homebrew instead of relying on system Clang
* Use newer version of spdlog as a submodule
* Use newer version of GoogleTest
* Force explicitly to use the path of libc++ from homebrew LLVM if on Apple
* Use UInt64 primitive in CountProcessor instead of size_t

Regress:
* Pkill and process killing is not reliable on Apple: use netcat to check for port being free and add wait loops

### DIFF
--- a/test/query/queries/MatchCreateTest.cpp
+++ b/test/query/queries/MatchCreateTest.cpp
@@ -2,6 +2,7 @@
 #include <optional>
 #include <cstdint>
 
+#include "EdgeRecord.h"
 #include "TuringDB.h"
 #include "Graph.h"
 #include "SimpleGraph.h"
@@ -13,6 +14,7 @@
 #include "reader/GraphReader.h"
 #include "dataframe/Dataframe.h"
 
+#include "LineContainer.h"
 #include "TuringTestEnv.h"
 #include "TuringTest.h"
 
@@ -61,6 +63,35 @@ protected:
         _graph = _env->getSystemManager().getGraph(std::string {name});
         ASSERT_TRUE(_graph);
     }
+
+    LabelID getLabelID(std::string_view labelName) {
+        auto labelOpt = read().getMetadata().labels().get(labelName);
+        EXPECT_TRUE(labelOpt.has_value()) << "Label not found: " << labelName;
+        return labelOpt.value_or(LabelID{0});
+    }
+
+    EdgeTypeID getEdgeTypeID(std::string_view edgeTypeName) {
+        auto edgeTypeOpt = read().getMetadata().edgeTypes().get(edgeTypeName);
+        EXPECT_TRUE(edgeTypeOpt.has_value()) << "EdgeType not found: " << edgeTypeName;
+        return edgeTypeOpt.value_or(EdgeTypeID{0});
+    }
+
+    bool nodeHasLabel(NodeID nodeID, LabelID labelID) {
+        return read().getNodeLabelSet(nodeID).hasLabel(labelID);
+    }
+
+    EdgeTypeID getEdgeType(EdgeID edgeID) {
+        return read().getEdgeTypeID(edgeID);
+    }
+
+    static NamedColumn* findColumn(const Dataframe* df, std::string_view name) {
+        for (auto* col : df->cols()) {
+            if (col->getName() == name) {
+                return col;
+            }
+        }
+        return nullptr;
+    }
 };
 
 // =============================================================================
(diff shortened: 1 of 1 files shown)


io/fs/AlignedBuffer.h
```diff
@@ -18,7 +19,8 @@ class AlignedBufferIterator;
 
 class AlignedBuffer {
 public:
-    static constexpr size_t ALIGNMENT {4096ul};
+    // Alignment must be a multiple of system page size (16KB on Apple Silicon, 4KB on Intel)
+    static constexpr size_t ALIGNMENT {16384ul};
```
rjb32, line 23: It was necessary to use 16KB as it is a multiple of 4KB and it is the minimum for Apple Silicon.

io/fs/FilePageReader.cpp
```diff
@@ -8,7 +8,11 @@
 using namespace fs;
 
 Result<FilePageReader> FilePageReader::open(const Path& path, size_t pageSize) {
+#ifdef __APPLE__
```
rjb32, line 11: Apple does not support O_DIRECT which is Linux-only

net/http_server/HTTPServer.cpp
```diff
@@ -24,13 +23,18 @@ HTTPServer::~HTTPServer() {
 }
 
 FlowStatus HTTPServer::initialize() {
-    _serverSocket = ::socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, IPPROTO_TCP);
+    _serverSocket = ::socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
```
rjb32, line 26: The non-blocking is managed afterwards by utils::setNonBlock

net/http_server/Utils.cpp
```diff
@@ -76,15 +98,103 @@ bool listen(ServerSocket s) {
 }
 
 bool epollAdd(EpollInstance instance, Socket fd, EpollEvent& event) {
-    return epoll_ctl(instance, EPOLL_CTL_ADD, fd, &event) != -1;
+#ifdef __APPLE__
+    struct kevent kev;
```
rjb32, line 102: memzero kev struct to initialise properly

net/http_server/Utils.cpp
```diff
@@ -76,15 +98,103 @@ bool listen(ServerSocket s) {
...
+    uint16_t flags = EV_ADD;
+    if (event.events & EVENT_ET) flags |= EV_CLEAR;
```
rjb32, line 104: Don't use contracted if, use if with brackets and go on next line

net/http_server/Utils.cpp
```diff
@@ -76,15 +98,103 @@ bool listen(ServerSocket s) {
...
+#else
+    epoll_event ev { .events = event.events, .data = {.ptr = event.data} };
+    return epoll_ctl(instance, EPOLL_CTL_ADD, fd, &ev) != -1;
+#endif
 }
 
 bool epollMod(EpollInstance instance, Socket fd, EpollEvent& event) {
-    return epoll_ctl(instance, EPOLL_CTL_MOD, fd, &event) != -1;
+#ifdef __APPLE__
+    struct kevent kev;
```
rjb32, line 116: Initialize properly kev with memzero

net/http_server/Utils.cpp
```diff
@@ -76,15 +98,103 @@ bool listen(ServerSocket s) {
...
+    epoll_event ev { .events = event.events, .data = {.ptr = event.data} };
+    return epoll_ctl(instance, EPOLL_CTL_MOD, fd, &ev) != -1;
+#endif
 }
 
 bool epollDel(EpollInstance instance, Socket fd, EpollEvent& event) {
-    return epoll_ctl(instance, EPOLL_CTL_DEL, fd, &event) != -1;
+#ifdef __APPLE__
+    (void)event;
+    struct kevent kev;
```
rjb32, line 129: Initialize properly kev with memzero

net/http_server/Utils.cpp
```diff
@@ -76,15 +98,103 @@ bool listen(ServerSocket s) {
...
+}
+
+int eventWait(EpollInstance instance, EpollEvent* events, int maxEvents, int timeout) {
+    constexpr int MAX_EVENTS = 64;
+    if (maxEvents > MAX_EVENTS) maxEvents = MAX_EVENTS;
+
+#ifdef __APPLE__
+    struct kevent kevents[MAX_EVENTS];
+    struct timespec* ts = nullptr;
+    struct timespec timeout_ts;
```
rjb32, line 154: Initialize properly timespec with memzero

net/http_server/Utils.cpp
```diff
@@ -76,15 +98,103 @@ bool listen(ServerSocket s) {
...
+    if (timeout >= 0) {
+        timeout_ts.tv_sec = timeout / 1000;
+        timeout_ts.tv_nsec = (timeout % 1000) * 1000000;
+        ts = &timeout_ts;
+    }
+    int n = kevent(instance, nullptr, 0, kevents, maxEvents, ts);
```
rjb32, line 160: n can be const

net/http_server/Utils.cpp
```diff
@@ -76,15 +98,103 @@ bool listen(ServerSocket s) {
...
+        if (kevents[i].filter == EVFILT_READ || kevents[i].filter == EVFILT_SIGNAL)
+            events[i].events |= EVENT_IN;
+        if (kevents[i].flags & EV_EOF)
+            events[i].events |= EVENT_RDHUP | EVENT_HUP;
+    }
+    return n;
+#else
+    // Convert our EpollEvent array to epoll_event array
+    epoll_event epollEvents[MAX_EVENTS];
+    int n = epoll_wait(instance, epollEvents, maxEvents, timeout);
```
rjb32, line 173: n can be const

net/http_server/Utils.h
```diff
@@ -1,10 +1,14 @@
 #pragma once
 
+#ifdef __APPLE__
+#include <sys/event.h>
+#else
+#include <sys/epoll.h>
+#endif
+
 #include <array>
 #include <cstdint>
```
rjb32, line 10: Prefer stdint.h instead of cstdint

net/http_server/Utils.h
```diff
@@ -13,11 +17,31 @@ using Socket = int;
...
 using DataSocket = int;
 using EpollInstance = int;
-using EpollEvent = ::epoll_event;
 using EpollSignal = int;
 using StringAddress = std::array<char, ADDR_LEN>;
 using UserData = void*;
 
+// Cross-platform event structure
+struct EpollEvent {
+    uint32_t events;
```
rjb32, line 26: Default initialise each member of the struct, events to 0 and data to {nullptr}

io/fs/FilePageReader.cpp
LucLabarriere: Interesting, does that mean that we cannot bypass the macOS page cache?

storage/iterators/ChunkWriter.h
```diff
@@ -1,11 +1,11 @@
 #pragma once
 
 #include "columns/ColumnVector.h"
+#include "views/NodeView.h"
 
 namespace db {
 
 class Iterator;
-class NodeView;
```
LucLabarriere, line 8: Is it necessary to include the file here? It will be included everywhere

io/fs/FilePageReader.cpp
rjb32: Yes it's done with F_NOCACHE passed to fcntl later on line 26 https://stackoverflow.com/questions/2299402/how-does-one-do-raw-io-on-mac-os-x-ie-equivalent-to-linuxs-o-direct-flag

storage/iterators/ChunkWriter.h
rjb32: I tested to remove it in https://github.com/turing-db/turingdb/pull/212/commits/404e6f80010b9cf9e43071db664b329c240f459e the build failed:
```cpp
/opt/homebrew/Cellar/llvm/21.1.8/bin/../include/c++/v1/__vector/vector.h:388:48: error: arithmetic on a pointer to an incomplete type 'db::NodeView'
  388 |     return static_cast<size_type>(this->__end_ - this->__begin_);
      |                                   ~~~~~~~~~~~~ ^
/opt/homebrew/Cellar/llvm/21.1.8/bin/../include/c++/v1/__vector/vector.h:542:28: note: in instantiation of member function 'std::vector<db::NodeView>::size' requested here
  542 |     size_type __old_size = size();
      |                            ^
/opt/homebrew/Cellar/llvm/21.1.8/bin/../include/c++/v1/__vector/vector.h:251:16: note: in instantiation of member function 'std::vector<db::NodeView>::clear' requested here
  251 |         __vec_.clear();
      |                ^
/opt/homebrew/Cellar/llvm/21.1.8/bin/../include/c++/v1/__vector/vector.h:262:67: note: in instantiation of member function 'std::vector<db::NodeView>::__destroy_vector::operator()' requested here
  262 |   _LIBCPP_CONSTEXPR_SINCE_CXX20 _LIBCPP_HIDE_FROM_ABI ~vector() { __destroy_vector (*this)(); }
      |                                                                   ^
/Users/runner/work/turingdb/turingdb/storage/columns/ColumnVector.h:60:5: note: in instantiation of member function 'std::vector<db::NodeView>::~vector' requested here
   60 |     ~ColumnVector() override = default;
      |     ^
/Users/runner/work/turingdb/turingdb/storage/iterators/ChunkWriter.h:59:33: note: in defaulted destructor for 'db::ColumnVector<db::NodeView>' first required here
   59 |     { it.setNodeViewsColumn(new ColumnVector<NodeView>) };
```

as it is true that NodeView is used as a template parameter of ColumnVector in ChunkWriter.h

LucLabarriere: OK I think it's fine, we can have a compilation time optimisation step at some point, but it's not a big deal for now.

PR #215, opened 2026-01-12
Remove boost

rjb32:


import/neo4j/Neo4jInstance.cpp
```diff
@@ -1,10 +1,10 @@
 #include "Neo4jInstance.h"
 
-#include <boost/process.hpp>
 #include <chrono>
+#include <cstdlib>
 #include <filesystem>
 #include <spdlog/spdlog.h>
```
rjb32: Order of includes, this should not be among system includes

import/neo4j/Neo4jInstance.cpp
```diff
@@ -1,10 +1,10 @@
 #include "Neo4jInstance.h"
 
-#include <boost/process.hpp>
 #include <chrono>
+#include <cstdlib>
```
rjb32: Prefer stdlib.h instead of cstdlib

PR #223, opened 2026-01-14
Bitwise implementation of ColumnKinds, PairColumnKind and Expression kinds

LucLabarriere:
Main logic is stored in the following header files:
- InternalKind.h: Code that represents the internal type of the container (size_t, string, etc)
- ContainerKind.h: Code that represents the container type (vector, const, etc)
- ColumnKind.h: Code that represents the pair container/internal
- PairColumnKind.h: Code that represents the pair column (e.g. vector<size_t>+const<size_t>)
- ExpressionKind.h: code that represents the expression (Operator + ColumnPair or single column for unary operations)

### DIFF
--- a/storage/columns/InternalKind.h
+++ b/storage/columns/InternalKind.h
@@ -36,7 +36,16 @@ class Column;
 class InternalKind {
 private:
     using Types = KindTypes<
-        size_t,
+        signed char,
+        unsigned char,
+        short int,
+        unsigned short int,
+        int,
+        unsigned int,
+        long int,
+        unsigned long int,
+        long long int,
+        unsigned long long int,
         EntityID,
         NodeID,
         EdgeID,
@@ -53,7 +62,6 @@ private:
         PropertyType,
         PropertyNull,
         bool,
-        types::Int64::Primitive,
         types::Double::Primitive,
         types::Bool::Primitive,
         types::String::Primitive,


storage/columns/OperationKind.h
```diff
@@ -0,0 +1,50 @@
...
+    static constexpr Code BitCount = sizeof(Code) * 8;
+
+    static constexpr Code MaxValue = (2ull << (BitCount - 1ull)) - 1ull;
+
+    // Checking if we can fit operator kind + pair column type into ExprOpKind
+    static_assert(BitCount >= sizeof(ColumnOperator) * 8 + PairColumnKind::BitCount);
+
+    // Binary expressions
+    template <ColumnOperator Op, typename Lhs, typename Rhs>
+    static consteval Code code() {
```
rjb32, line 26: I think it's quite elegant

storage/columns/InternalKind.h
```diff
@@ -0,0 +1,116 @@
+#pragma once
+
+#include <concepts>
+#include <cstdint>
```
rjb32: Prefer the stdint.h include style instead of cstdint

storage/columns/ColumnKind.h
```diff
@@ -1,161 +1,73 @@
...
+    static constexpr Code Invalid = std::numeric_limits<Code>::max();
 
-        _SIZE,
-    };
+    /// @brief The number of bits required to represent the column type codes
+    static constexpr Code BitCount = sizeof(Code) * 8;
 
-    using LabelSet = TemplateLabelSet<uint64_t, 4>;
+    /// @brief The maximum value of the column type codes
+    static constexpr Code MaxValue = (2 << (BitCount - 1)) - 1;
```
cyrusknopf: Here MaxValue == Invalid, does that matter? Would we ever need to use MaxValue as an actual value of a Column kind?

query/pipeline/processors/ExprProgram.cpp
```diff
@@ -234,15 +196,14 @@ void ExprProgram::evalInstr(const Instruction& instr) {
     switch (getOperatorType(op)) {
         case ColumnOperatorType::OPTYPE_BINARY:
             evalBinaryInstr(instr);
-        break;
+            break;
 
         case ColumnOperatorType::OPTYPE_UNARY:
             evalUnaryInstr(instr);
-        break;
```
cyrusknopf, line 246: I think @rjb32 wants the old style? With the `break` aligned with `case` unless I'm misremembering

rjb32: Yes please align the break statement with the case

rjb32:
In principle the encoding looks good to me

rjb32:
@LucLabarriere I got issues in trying to make the macos build work with the `__COUNTER__ `macro that seems to behave differently across compilers. Claude suggested to make a giant enum whenever necessary to have get a well defined code.

storage/columns/ColumnKind.h
LucLabarriere: That's a good point yeah thanks!

LucLabarriere: I'll make it MaxValue - 1, reserve the actual MaxValue for Invalid, add a static_assert(MaxValue != Invalid)

LucLabarriere:
> @LucLabarriere I got issues in trying to make the macos build work with the `__COUNTER__ `macro that seems to behave differently across compilers. Claude suggested to make a giant enum whenever necessary to have get a well defined code.

Oh that's right I forgot that it was compiler specific, I'll see how I change it

storage/columns/ContainerKind.h
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// @brief Value indicating an invalid container type code
+    static constexpr Code Invalid = std::numeric_limits<Code>::max();
+
+    /// @brief The number of bits required to represent the container type codes
+    static constexpr size_t BitCount = 8;
+
+    /// @brief The maximum value of the container type codes
+    static constexpr size_t MaxValue = (2 << (BitCount - 1)) - 2;
```
LucLabarriere: @cyrusknopf doing -2 for the max value, and reserving the actual max value of Invalid should fix it

storage/columns/InternalKind.h
```diff
@@ -0,0 +1,107 @@
...
+    static constexpr size_t BitCount = 8;
+
+    /// @brief The maximum value of the internal type codes
+    static constexpr size_t MaxValue = (2 << (BitCount - 1)) - 2;
+
+    /// @brief The number of internal type codes
+    static constexpr size_t Count = Types::count();
+
+    static_assert(MaxValue != Invalid);
+};
```
LucLabarriere, line 113: @rjb32 @cyrusknopf 

Tuple based implementation. There is a static_assert that ensures all types are unique. When we request a type, a static_assert ensures it can be found in the tuple.

This gets rid of the __COUNTER__, while allowing to add types easily

cyrusknopf: Yes I was really hoping we could end up with something like this! Having all the internal types defined in a tuple somewhere will be much easier to maintain

storage/columns/ContainerKind.h
```diff
@@ -0,0 +1,71 @@
...
+            constexpr auto code = (Code)Types::indexOf<Outer>();
+            return code;
+        }
+    }
+
+    /// @brief Value indicating an invalid container type code
+    static constexpr Code Invalid = std::numeric_limits<Code>::max();
+
+    /// @brief The number of bits required to represent the container type codes
+    static constexpr size_t BitCount = 8;
```
cyrusknopf: Maybe worth replacing with `sizeof(Code) * 8` to avoid hard-coding 8?

storage/columns/InternalKind.h
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /// @brief Value indicating an invalid internal type code
+    static constexpr Code Invalid = std::numeric_limits<Code>::max();
+
+    /// @brief The number of bits required to represent the internal type codes
+    static constexpr size_t BitCount = 8;
+
+    /// @brief The maximum value of the internal type codes
+    static constexpr size_t MaxValue = (2 << (BitCount - 1)) - 2;
```
cyrusknopf: Should this be `Code MaxValue` instead of `size_t MaxValue`?

storage/columns/OperationKind.h
```diff
@@ -0,0 +1,55 @@
...
+    static constexpr Code BitCount = sizeof(Code) * 8;
+
+    // Checking if we can fit operator kind + pair column type into ExprOpKind
+    static_assert(BitCount >= OpCodeBitCount + PairColumnKind::BitCount);
+
+    // Binary expressions
+    template <ColumnOperator Op, typename Lhs, typename Rhs>
+    static consteval Code code() {
+        const Code c1 = std::to_underlying(Op);
+        const Code c2 = PairColumnKind::code<Lhs, Rhs>();
```
cyrusknopf, line 28: Maybe I'm misunderstanding, but is this getting a `PairColumnKind::Code` but instantiating it as a `OperationKind::Code`?

LucLabarriere: Yes, it's basically casting it (to a bigger integer so that we can do the bit shift after

cyrusknopf: Ah yes I thought `sizeof(PairColumnKind::Code) > sizeof(OperationKind::Code)`, but I was wrong, looks good

LucLabarriere: I need to double check, I think there's a mistake

storage/columns/InternalKind.h
rjb32: Thanks it looks good

storage/columns/OperationKind.h
LucLabarriere: Nevermind, we should be good, byte layout is: `__ _a | bb bb ` with first 3 bytes unused, byte `a` is the operation, and the `b` bytes correspond to the column pair code

rjb32 (approved):

rjb32:
You can merge as the macos CI is done

PR #229, opened 2026-01-15
Create nightly docker image

ghost:


.github/workflows/ci_ubuntu.yml
```diff
@@ -58,3 +58,8 @@ jobs:
     - name: Run regress
       shell: bash
       run: cd ${{ github.workspace }}/build && source ${{ github.workspace }}/setup.sh && make run_regress
+    - name: Upload turing_install
+      uses: actions/upload-artifact@v3
+      with:
+        name: turingdb-binaries       # artifact name
+        path: ${{ github.workspace }}/build/turing_install  # upload the whole folder
```
rjb32: Please make it a .tar.gz archive first, that would be best for a github release

.github/workflows/ci_docker.yaml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          context: .
+          file: Dockerfile
+          push: false
+          tags: turingdbai/turingdb:nightly
+
+      # Test Docker image
+      - name: Test Docker image
+        run: |
+          docker run --rm turingdbai/turingdb:nightly turingdb --help
```
rjb32: I think we actually test the docker image minimally right here that's good.

.github/workflows/ci_docker.yaml
```diff
@@ -0,0 +1,65 @@
+name: Nightly Docker Build
+
+on:
+  schedule:
+    - cron: "0 2 * * *" # every night at 02:00 UTC
+  workflow_dispatch: # manual trigger
```
rjb32: We should also be able to trigger this workflow when there is a "v*" git tag. Look at what we did in https://github.com/turing-db/turingdb-python/blob/main/.github/workflows/ci_ubuntu.yml

rjb32:
Also push a git tag to this branch once you made the ci_docker workflow to be triggered by a tag so that we can check that the workflow works in this PR

.github/workflows/ci_docker.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+      # Download turingdb archive from main CI
+      - name: Download archive
+        uses: actions/download-artifact@v3
+        with:
+          name: turing-install-archive
+          path: docker/artifacts
+      
+      # Extract archive into Docker context
+      - name: Extract turing_install
```
rjb32: Instead of using the archive it would actually be a better and simpler idea to just use the turingdb pip package:
pip install turingdb

rjb32 (approved):

PR #240, opened 2026-01-15
#258 Implement filtering on Node/EdgePatterns

cyrusknopf:
Context:
- Previously we did not have Node/EdgePattern comparisons, so we could not do:
```
MATCH (n), (m) WHERE n <> m RETURN n
```
  to find pairs of nodes that are distinct.

Changes:
- Implement `SymbolExpr` evaluation in `ExprProgramGenerator`
- Add cases for column operators to take Node/EdgeIDs

Verified with:
 - Enabling of previously disabled unit tests which use `<>` and `=` on Node/EdgePatterns

### DIFF
--- a/test/query/queries/AutoJoinTest.cpp
+++ b/test/query/queries/AutoJoinTest.cpp
@@ -767,8 +767,7 @@ TEST_F(JoinFeatureTest, undirectedEdgeJoin) {
 // =============================================================================
 
 // Test 15: Self-join with equality filter (a = b)
-// DISABLED: Comparison on NodePattern not yet supported
-TEST_F(JoinFeatureTest, DISABLED_selfJoinWithEquality) {
+TEST_F(JoinFeatureTest, selfJoinWithEquality) {
     constexpr std::string_view QUERY = R"(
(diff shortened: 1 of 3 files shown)


query/plan/ExprProgramGenerator.cpp
```diff
@@ -231,6 +232,53 @@ Column* ExprProgramGenerator::generateLiteralExpr(const LiteralExpr* literalExpr
...
+    const auto& incomingStream = _pendingOut.getInterface()->getStream();
+        const bool incStreamContainsVar = (isNode && incomingStream.isNodeStream())
+                                       || (!isNode && incomingStream.isEdgeStream());
+        if (!incStreamContainsVar) {
+            throw FatalException(
+                fmt::format("Could not find column associated with symbol variable {}.",
+                            exprVarDecl->getName()));
+        }
+
+        // XXX: Is this correct? May it be an edge stream but we need extract tgt ids tag?
```
rjb32: It's correct if it's an edge variable then it's a comparison on EdgeIDs

cyrusknopf:
```suggestion
```

Issue #297, opened 2026-01-15
Segmentation fault in GMLGraphSax


rjb32:
Segmentation fault in GMLGraphSax with the following unit test:
```c++
TEST_F(GMLParserTest, EdgePropertyBeforeSourceTargetCrash) {
    // This GML has edge property "label" BEFORE "source" and "target"
    // GMLGraphSax::onEdgeProperty() will dereference _currentEdge which is nullptr
    constexpr std::string_view crashGml =
        "graph [\n"
        "  node [ id 0 ]\n"
        "  node [ id 1 ]\n"
        "  edge [ label \"crash\" source 0 target 1 ]\n"
        "]";

    auto graph = Graph::create();
    GMLGraphSax sax(graph.get());
    GMLGraphParser parser(sax);

    sax.prepare();
    // This will crash due to null pointer dereference in onEdgeProperty
    auto res = parser.parse(crashGml);
    ASSERT_TRUE(res);
}
```

rjb32:
Fixed in #298

PR #298, opened 2026-01-15
#297 Fix GMLGraphSax segfault

rjb32:
Fixes the GMLGraphSax segfault in https://github.com/turing-db/turingdb/issues/297

### DIFF
--- a/import/gml/GMLGraphParser.h
+++ b/import/gml/GMLGraphParser.h
@@ -207,43 +207,7 @@ public:
     bool onEdgeEnd() { return true; }
 
 private:
-    void applyBufferedEdgeProperties() {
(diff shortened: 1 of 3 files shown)


import/gml/GMLGraphParser.h
```diff
@@ -168,12 +200,51 @@ class GMLGraphSax {
...
         _edgeSourceID = NodeID {};
         _edgeTargetID = NodeID {};
+        _bufferedEdgeProps.clear();
         return true;
     }
 
     bool onEdgeEnd() { return true; }
 
 private:
+    void applyBufferedEdgeProperties() {
```
rjb32: Put this big function in the cpp file

Issue #255, opened 2026-01-15
Incorrect iterator handling in HashJoinProcessor.cpp


rjb32:
**Reported by:** @rjb32

In HashJoinProcessor.cpp line 203:

```
const auto it = findInMap(_leftMap, rightCol, i); it != _rightMap.end()
```

The find is executed against the _leftMap but the resulting iterator is checked against the _rightMap

rjb32:
Fixed by #341

PR #309, opened 2026-01-20
#255 Fix Bug In Hash Join

ghost:


rjb32:
Actually the submodule test/googletest is marked as deleted in this PR. I added it in main, you may need to run pull.sh locally and then push the fix

rjb32:
Fixing #255

rjb32:
Please remove the changes from the test/googletest directory in this PR

PR #311, opened 2026-01-20
Column dispatch + operation evaluation rework

LucLabarriere:
- Single dispatch
- Double dispatch

Both using: switch on the container types first (vector, const, set, mask), then jump table based on the internal kind (int, double, etc).

+ @cyrusknopf Cleaner implementation of Column operators

### DIFF
--- a/query/pipeline/processors/EvalBinaryExpr.cpp
+++ b/query/pipeline/processors/EvalBinaryExpr.cpp
@@ -1,0 +1,215 @@
+#include "EvalBinaryExpr.h"
+
+#include "columns/ColumnOperationExecutor.h"
+
+#include "Panic.h"
+
+using namespace db;
+
+namespace {
+
+template <ColumnOperator Op>
+struct EqualEval {
+    Column* _res {nullptr};
+
+    template <typename T, typename U>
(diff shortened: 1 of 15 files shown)


query/pipeline/processors/EvalBinaryExpr.cpp
```diff
@@ -0,0 +1,215 @@
...
+            // using ResultType = ColumnCombinations<Equal, T, U>::ResultType;
+            // execOperation<Equal>(static_cast<ResultType*>(_res), lhs, rhs);
+            fmt::println("EQUAL {}, {}", typeid(*lhs).name(), typeid(*rhs).name());
+        } else if constexpr (Op == OP_NOT_EQUAL) {
+            // using ResultType = ColumnCombinations<NotEqual, T, U>::ResultType;
+            // execOperation<NotEqual>(static_cast<ResultType*>(_res), lhs, rhs);
+            fmt::println("NOT EQUAL {}, {}", typeid(*lhs).name(), typeid(*rhs).name());
+        } else {
+            COMPILE_ERROR("Invalid operator for Equal");
+        }
```
LucLabarriere, line 29: @cyrusknopf This is where you should be able to plug your system in

PR #312, opened 2026-01-20
Fix copy past error in PipelineGenerator for WriteProcessor::PendingEdges reserve

rjb32:
Fix copy-paste error in PipelineGenerator.
We had for penEdges at line 1024:
```c++
WriteProcessor::PendingEdges penEdges;
penNodes.reserve(node->pendingEdges().size());
{
    for (const WriteNode::PendingEdge& pendingPlanEdge : node->pendingEdges()) {
``` 
which reserved in penNodes instead of penEdges. It looks like a copy-paste error.
Instead we should do:
```c++
penEdges.reserve(node->pendingEdges().size());
```


cyrusknopf (approved):
My mistake, thanks!

PR #316, opened 2026-01-21
Specification and design for vector DB integration in the query language

rjb32:


docs/vector/VECTOR_INTEGRATION.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Query commands
+
+The query commands that we want can be possibly decomposed into several steps:
+
+1. Create a vector index for a graph inside TuringDB.
+This is independent of versioning.
+
+```
+CREATE VECTOR INDEX vectordb
```
rjb32, line 13: * Set dimension 
* Set metric

docs/vector/VECTOR_INTEGRATION.md
```diff
@@ -0,0 +1,60 @@
...
+or
+
+```
+LOAD VECTOR FROM "myfilepath" IN vectordb
+```
+
+4. Get a column of the numerical values associated to the n nearest vectors for a given query vector
+
+```
+VECTOR SEARCH vectordb "embedding vector"
```
rjb32, line 32: We can pass a list of floats as literal and we can take it in plan graph/pipeline generation to not have to support lists as values

docs/vector/VECTOR_INTEGRATION.md
```diff
@@ -0,0 +1,60 @@
+# Vector DB integration in TuringDB
+
+We want to integrate in the query language the vector DB implementation in the vector/ directory and exemplified by the sample program in samples/vector-db directory.
+
+## Query commands
+
+The query commands that we want can be possibly decomposed into several steps:
```
rjb32, line 7: * We may actually want to put the vector database at the root level, not per graph.
* We should actually not use the term of "index" as this is rather more a separate vector store, not necessarily synchronized with all the nodes and edges automatically, and also usable to index other kind of data in the future.

rjb32: * Add a create vector library command and refer to vector library afterwards

Issue #317, opened 2026-01-21
Some endpoints crash the server if the requested node is invalid


rjb32:
Reported by @LucLabarriere on 10/03/2025.

The REST API server crashes if we ask for NodeIDs that don't exist in some endpoints such as /get_nodes

This is an analysis of what happens for /get_nodes
```
DBServerProcessor::get_nodes() (server/DBServerProcessor.cpp:1036-1039)                                                                                                                                                                                                                                                                                                         
- Iterates over requested node IDs without validation                                                                                                                                                                                                                                                                                                                            
- Calls reader.getNodeView(nodeID) for each                                                                                                                                                                                                                                                                                                                                      

GraphReader::getNodeView() (storage/reader/GraphReader.cpp:171-191)                                                                                                                                                                                                                                                                                                             
- When node doesn't exist, returns a default-constructed NodeView                                                                                                                                                                                                                                                                                                                
- The returned view has _labelset as a default LabelSetHandle (null pointer internally)                                                                                                                                                                                                                                                                                          

PayloadWriter::write(const NodeView&) (server/PayloadWriter.h:249)                                                                                                                                                                                                                                                                                                              
- Calls n.labelset().decompose(labelIDs) without checking validity                                                                                                                                                                                                                                                                                                               

LabelSetHandle::decompose() → hasLabel() (storage/metadata/LabelSetHandle.h:182-183)                                                                                                                                                                                                                                                                                            
bool hasLabel(LabelID id) const {                                                                                                                                                                                                                                                                                                                                                  
 const auto* integers = _labelset->data();  // CRASH: _labelset is nullptr 
```

PR #320, opened 2026-01-21
#317 Check that NodeIDs and EdgeIDs are valid in DBServerProcessor

rjb32:
Fix for #317 reported by @LucLabarriere 

Some server endpoints such as /get_nodes, /get_edges and /get_node_edges were using NodeView or EdgeView on requested NodeIDs/EdgeIDs without checking if the view is valid before using it.

I added the check of views first and return an error if one of the IDs is invalid without starting to write the response before the check, so that introduces a bit of memorization of the views before using them.

### DIFF
--- a/regress/CMakeLists.txt
+++ b/regress/CMakeLists.txt
@@ -35,6 +35,7 @@ add_subdirectory(read_script)
 add_subdirectory(neo4j_import)
 add_subdirectory(wine_ontology)
 add_subdirectory(bulk_create_crash)
+add_subdirectory(get_nodes_invalid_id)
 
 # Python SDK version to be used by uv
 set(PYTURINGDB "turingdb")
--- a/regress/get_nodes_invalid_id/CMakeLists.txt
+++ b/regress/get_nodes_invalid_id/CMakeLists.txt
@@ -1,0 +1,2 @@
+
+regress_test(get_nodes_invalid_id run.sh main.py)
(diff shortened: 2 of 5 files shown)


rjb32:
> **Claude finished @rjb32's task** —— [View job](https://github.com/turing-db/turingdb/actions/runs/21216880411)
> 
> ### Code Review Complete
> * [x]  Read CODING_STYLE.md to understand all rules
> * [x]  Get the PR diff for changed lines
> * [x]  Analyze C++ files for coding style violations
> * [x]  Report findings
> 
> ### Coding Style Review
> I found the following violation in the changed C++ code:
> 
> **[Move semantics and rvalue references]**: CODING_STYLE.md states: _"Please do not use RVO, move semantics or anything relying on rvalue references and other related mechanisms in C++."_
> 
> The PR uses `std::move()` in two places:
> 
> * [`server/DBServerProcessor.cpp:1043`](https://github.com/turing-db/turingdb/blob/ff2d94560/server/DBServerProcessor.cpp#L1043): `nodes.push_back(std::move(node));`
> * [`server/DBServerProcessor.cpp:1472`](https://github.com/turing-db/turingdb/blob/ff2d94560/server/DBServerProcessor.cpp#L1472): `edges.push_back(std::move(edge));`
> 
> **Fix**: Replace `std::move(node)` with just `node` and `std::move(edge)` with just `edge`. The push_back will perform a copy, which is the expected behavior per coding guidelines.

Here I think we don't want to do a copy of the entire view so std::move is necessary.

PR #327, opened 2026-01-22
JSON import of embeddings

LucLabarriere:


test/vector/JsonImporterTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+        ASSERT_TRUE(res);
+    }
+
+    batch.clear(8);
+
+    {
+        const std::string json = R"(
+            [
+                {
+                    "id": 20,
```
rjb32, line 154: I like it it's simple enough yet easy to parse

vector/import/ImportResult.cpp
```diff
@@ -0,0 +1,14 @@
+#include "import/ImportResult.h"
```
rjb32, line 1: Why is ImportResult.cpp not in the same directory as ImportResult.h

vector/import/ImportResult.h
```diff
@@ -0,0 +1,66 @@
...
+    JSONMissingIDField,
+    JSONMissingVectorField,
+
+    _SIZE,
+};
+
+using ImportErrorTypeDescription = EnumToString<ImportErrorCode>::Create<
+    EnumStringPair<ImportErrorCode::Unknown, "Unknown">,
+
+    EnumStringPair<ImportErrorCode::NoVectors, "No vectors found to import">,
```
rjb32, line 29: So it uses ImportErrorCode and also throwError inside. What's the rationale to divide between them?

vector/import/ImportResult.cpp
LucLabarriere: Oops mistake!

vector/import/ImportResult.h
LucLabarriere: Yeah it's a bit annoying, that's because nlohmann's sax interface is a bit restrictive for errors, we kind of have to throw. So I throw an error that I catch in JsonImporter.cpp and return the error code there

rjb32:
> **Claude finished @LucLabarriere's task** —— [View job](https://github.com/turing-db/turingdb/actions/runs/21247907865)
> 
> ### Code Review Complete
> * [x]  Read CODING_STYLE.md to understand rules
> * [x]  Get PR diff and analyze changed files
> * [x]  Submit review findings
> 
> #### Violations Found
> **[Consecutive blank lines]**: There should never be 2 or more consecutive blank lines in the code.
> 
> * [vector/import/JsonSaxParser.h:347-348](https://github.com/turing-db/turingdb/blob/bb69fcda4/vector/import/JsonSaxParser.h#L347-L348) - Two consecutive blank lines between the `else` block and the next function
> 
> **[Move semantics]**: CODING_STYLE.md states "Please do not use RVO, move semantics or anything relying on rvalue references and other related mechanisms in C++." The following locations use `std::move()` or rvalue references:
> 
> * [vector/import/ImportException.h:20-21](https://github.com/turing-db/turingdb/blob/bb69fcda4/vector/import/ImportException.h#L20-L21) - Returns `ImportError&&` (rvalue reference) and uses `std::move(_e)`
> * [test/vector/JsonImporterTest.cpp:26](https://github.com/turing-db/turingdb/blob/bb69fcda4/test/vector/JsonImporterTest.cpp#L26) - Uses `std::move(res.value())`
> * [samples/vector-db/main.cpp:26](https://github.com/turing-db/turingdb/blob/bb69fcda4/samples/vector-db/main.cpp#L26) - Uses `std::move(dbRes.value())`
> 
> Consider returning by pointer or using output references instead of move semantics.

The std::move usage in these cases is appropriate.

PR #332, opened 2026-01-22
Implement deterministic code review script in python

rjb32:
Implement a deterministic code review script in python to avoid spending too much on Claude.
It has been itself generated from Claude once (hopefully) from CODING_STYLE.md under human supervision and testing.
False positives happen and are to be reported to be fixed.

See manual in scripts/README.md

To run the entire codebase:
```shell
uv run --project scripts/codereview/ codereview --all
``` 

To run on the files changed in the current branch compared to main:
```shell
uv run --project scripts/codereview/ codereview --diff origin/main
``` 

Lists of checks:
* Using namespace in header files
* Include order
* Initialisation of every member of class or struct of pointer type to nullptr
* Detection of STL data structures returned by value
* Detection of non-trivial classes returned by value
* Missing advisor (warning)
* Constructor and destructor brackets
* Bracket positioning

### DIFF
--- a/io/s3/TuringS3Client.cpp
+++ b/io/s3/TuringS3Client.cpp
@@ -294,7 +294,7 @@ S3ClientResult<void> TuringS3Client<T>::uploadDirectory(const std::string& direc
             const Aws::String relativePath = entry.relativePath;
 
             ssKey << prefix << relativePath;
-            Aws::String keyName = ssKey.str();
+            const Aws::String keyName = ssKey.str();
             if (auto res = uploadFile(entry.path, bucketName, keyName); !res) {
                 directoryRes = res;
                 return false;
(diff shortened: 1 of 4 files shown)


import/gml/GMLParser.h
```diff
@@ -6,8 +6,6 @@
 #include "BasicResult.h"
 #include "GMLError.h"
 
-using namespace db;
```
rjb32, line 9: All the changes in this file are to remove `using namespace db` from header

query/AST/DiagnosticsManager.cpp
```diff
@@ -13,9 +14,9 @@ DiagnosticsManager::DiagnosticsManager(SourceManager* sourceManager)
 DiagnosticsManager::~DiagnosticsManager() {
 }
 
-std::string DiagnosticsManager::createErrorString(std::string_view msg, const void* obj) const {
+void DiagnosticsManager::createErrorString(std::string_view msg, const void* obj,
```
rjb32: To fix alignment of arguments

PR #333, opened 2026-01-23
Add unit test showing segfault for dependencies in CALL YIELD + MATCH

rjb32:
I stumbled upon a segmentation fault when implementing VECTOR SEARCH + MATCH that appears also in CALL YIELD + MATCH which seems to be due to the variables of the YIELD clause not being properly handled as dependencies in PlanGraphGenerator.
```cypher
CALL db.labels() YIELD id MATCH (n) WHERE n.age = id RETURN n
```

### DIFF
--- a/test/query/queries/QueriesTest.cpp
+++ b/test/query/queries/QueriesTest.cpp
@@ -2598,6 +2598,34 @@ TEST_F(QueriesTest, indirectLabelFilter) {
     }
 }
 
+// Test that CALL + MATCH + WHERE with YIELD variable reference works correctly.
+// This tests a bug where YIELD variables were not registered as VarNodes in the
+// plan graph, causing a crash in placePredicateJoins when processing WHERE clauses
+// that reference the YIELD variable.
+//
+// The crash occurs because:
+// 1. `CALL db.labels() YIELD id` creates a variable `id`
+// 2. `MATCH (n) WHERE n.age = id` references `id` in the WHERE clause
+// 3. In `ExprDependencies::genExprDependencies`, when processing the SymbolExpr
+//    for `id`, it calls `variables.getVarNode(symbol->getDecl())`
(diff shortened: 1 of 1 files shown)


cyrusknopf:
@LucLabarriere @rjb32 Is this fixed?

LucLabarriere:
> @LucLabarriere @rjb32 Is this fixed?

I think it should be yeah, there is a unit test. It does not segfault anymore, but right now it prints an error saying we cannot compare those types. I assume we have three strategies to gracefully handle this now:
- Have a better error message (Cannot compare LabelID and Int64)
- Allow comparing LabelID to Int64 (I think it's acceptable personally, but debatable)
- Require to cast the value to int first, with a function call `WHERE toInteger(id) = n.age`

cyrusknopf:
> > @LucLabarriere @rjb32 Is this fixed?
> 
> I think it should be yeah, there is a unit test. It does not segfault anymore, but right now it prints an error saying we cannot compare those types. I assume we have three strategies to gracefully handle this now:
> 
> * Have a better error message (Cannot compare LabelID and Int64)
> * Allow comparing LabelID to Int64 (I think it's acceptable personally, but debatable)
> * Require to cast the value to int first, with a function call `WHERE toInteger(id) = n.age`

The unit tests which got added is not passing though haha

rjb32:
Yes that unit test should be modified it's a bit artificial. It was just to produce the segfault.
Are you adding one in your PR @cyrusknopf ?

cyrusknopf:
> Yes that unit test should be modified it's a bit artificial. It was just to produce the segfault. Are you adding one in your PR @cyrusknopf ?

Yes in #353 I have this exact test, as well as others which return values

rjb32:
Ok then I will close this PR

rjb32:
Covered by tests in #353

PR #334, opened 2026-01-26
Vector database query integration

rjb32:
## Summary

Integrates the vector database with the query language. Adds support for creating, loading, searching, and managing vector indexes.

### Create a vector index
```
CREATE VECTOR INDEX <name> WITH DIMENSION <dim> METRIC <metric>
```
Where `metric` is `EUCLID` or `COSINE`.

### Load vectors from file
```
LOAD VECTOR FROM "<filepath>" IN <index_name>
```

### Search for k-nearest neighbors
```
VECTOR SEARCH IN <index_name> FOR <k> [<vector>] YIELD ids
```

Can be combined with `MATCH`:
```
VECTOR SEARCH IN embeddings FOR 10 [0.5, 0.3, 0.8, 0.1] YIELD ids
MATCH (n:Document) WHERE n.id = ids
RETURN n.title, n.content
```

### Delete a vector index
```
DELETE VECTOR INDEX <name>
```

### List all vector indexes
```
SHOW VECTOR INDEXES
```

### DIFF
--- a/vector/VectorDatabase.cpp
+++ b/vector/VectorDatabase.cpp
@@ -160,7 +160,7 @@ VecLibAccessor VectorDatabase::getLibrary(std::string_view libName) {
 }
 
 VecLibWriteAccessor VectorDatabase::getLibraryForWrite(const VecLibID& libID) {
-    std::shared_lock lock {_mutex};
+    std::shared_lock lock(_mutex);
 
     auto it = _vecLibs.find(libID);
     if (it == _vecLibs.end()) {
@@ -171,7 +171,7 @@ VecLibWriteAccessor VectorDatabase::getLibraryForWrite(const VecLibID& libID) {
 }
 
 VecLibWriteAccessor VectorDatabase::getLibraryForWrite(std::string_view libName) {
-    std::shared_lock lock {_mutex};
+    std::shared_lock lock(_mutex);
 
     auto it = _vecLibIDs.find(libName);
     if (it == _vecLibIDs.end()) {
@@ -182,7 +182,7 @@ VecLibWriteAccessor VectorDatabase::getLibraryForWrite(std::string_view libName)
 }
 
 void VectorDatabase::listLibraryNames(std::vector<std::string>& out) const {
-    std::shared_lock lock {_mutex};
+    std::shared_lock lock(_mutex);
 
     out.clear();
     out.reserve(_vecLibs.size());
--- a/query/plan/ExprProgramGenerator.cpp
+++ b/query/plan/ExprProgramGenerator.cpp
@@ -377,7 +377,7 @@ Column* ExprProgramGenerator::generateFuncInvocationExpr(const FunctionInvocatio
 
         Column* argCol = generateExpr(args->front());
 
-        ColumnOperator convOp;
+        ColumnOperator convOp = OP_NOOP;
         Column* resCol = nullptr;
 
         if (funcName == "toInteger") {
--- a/samples/tfl/main.cpp
+++ b/samples/tfl/main.cpp
@@ -75,13 +75,13 @@ int main(int argc, const char** argv) {
         "LOAD CSV '" + std::string(csvPath.get())
         + "' WITH HEADERS AS row "
         "RETURN row.station1 AS s1, row.station2 AS s2, "
-        "row.line AS line, row.time AS time";
+        "row.line AS line, row._time AS time";
 
     struct Connection {
         std::string station1;
         std::string station2;
         std::string line;
-        double time;
+        double _time = 0.0;
     };
 
     std::vector<Connection> connections;
@@ -177,8 +177,8 @@ int main(int argc, const char** argv) {
             " line: \"{}\"}}]->({})"
             ",\n({})-[:CONNECTED_TO {{time: {:.1f},"
             " line: \"{}\"}}]->({})",
-            v1, conn.time, conn.line, v2,
-            v2, conn.time, conn.line, v1);
+            v1, conn._time, conn.line, v2,
+            v2, conn._time, conn.line, v1);
         edgeCount += 2;
     }
 
@@ -209,7 +209,7 @@ int main(int argc, const char** argv) {
     // Walk edges (all platforms of s1 <-> all platforms of s2)
     for (const auto& conn : connections) {
         if (conn.line != "Walk") continue;
-        double walkWeight = conn.time + penalty;
+        double walkWeight = conn._time + penalty;
         for (const auto& l1 :
              stationLines[conn.station1]) {
             for (const auto& l2 :
--- a/test/query/queries/VectorQueriesTest.cpp
+++ b/test/query/queries/VectorQueriesTest.cpp
@@ -23,7 +23,7 @@ namespace {
 
 // Helper struct to hold a vector with its ID
 struct TestVector {
-    uint64_t id {0};
+    uint64_t _id {0};
     std::vector<float> values;
 };
 
@@ -47,7 +47,7 @@ void findKNearestNeighbors(std::vector<uint64_t>& result,
     std::vector<std::pair<float, uint64_t>> distances;
     for (const auto& vec : vectors) {
         const float dist = euclideanDistanceSquared(vec.values, query);
-        distances.emplace_back(dist, vec.id);
+        distances.emplace_back(dist, vec._id);
     }
 
     // Sort by distance (ascending)
@@ -369,7 +369,7 @@ TEST_F(VectorQueriesTest, vectorSearchReturnsCorrectResults) {
     {
         std::ofstream out(vectorFile);
         for (const auto& vec : testVectors) {
-            out << vec.id;
+            out << vec._id;
             for (float v : vec.values) {
                 out << "," << v;
             }
@@ -438,7 +438,7 @@ TEST_F(VectorQueriesTest, vectorSearchWithDifferentK) {
     {
         std::ofstream out(vectorFile);
         for (const auto& vec : testVectors) {
-            out << vec.id;
+            out << vec._id;
             for (float v : vec.values) {
                 out << "," << v;
             }
@@ -507,7 +507,7 @@ TEST_F(VectorQueriesTest, vectorSearchWithHighPrecisionFloats) {
         std::ofstream out(vectorFile);
         out << std::fixed << std::setprecision(6);
         for (const auto& vec : testVectors) {
-            out << vec.id;
+            out << vec._id;
             for (float v : vec.values) {
                 out << "," << v;
             }
@@ -624,7 +624,7 @@ TEST_F(VectorQueriesTest, vectorSearchWithMatch) {
     {
         std::ofstream out(vectorFile);
         for (const auto& vec : testVectors) {
-            out << vec.id;
+            out << vec._id;
             for (const float v : vec.values) {
                 out << "," << v;
             }
--- a/vector/ShardCache.cpp
+++ b/vector/ShardCache.cpp
@@ -113,7 +113,7 @@ void ShardCache::evictLibraryShards(VecLibID libID) {
 }
 
 void ShardCache::flush() {
-    std::unique_lock lock {_mutex};
+    std::unique_lock lock(_mutex);
 
     for (auto& [id, it] : _accessedMap) {
         it->second._shard->save();


query/analyzer/CypherAnalyzer.cpp
```diff
@@ -320,6 +334,29 @@ void CypherAnalyzer::analyze(S3TransferQuery* s3Transfer) {
     s3Transfer->setS3Prefix(s3URL);
 }
 
+void CypherAnalyzer::analyze(const CreateVectorIndexQuery* query) {
+    if (query->getDimension() == 0) {
+        throwError("Vector index dimension must be greater than 0", query);
+    }
+
+    std::string_view indexName = query->getIndexName();
```
rjb32: can be const

query/analyzer/CypherAnalyzer.cpp
```diff
@@ -320,6 +334,29 @@ void CypherAnalyzer::analyze(S3TransferQuery* s3Transfer) {
...
+    if (indexName.empty()) {
+        throwError("Vector index name cannot be empty", query);
+    }
+}
+
+void CypherAnalyzer::analyze(const LoadVectorQuery* query) {
+    std::string_view filePath = query->getFilePath();
```
rjb32: can be const

query/analyzer/CypherAnalyzer.cpp
```diff
@@ -320,6 +334,29 @@ void CypherAnalyzer::analyze(S3TransferQuery* s3Transfer) {
...
+    if (filePath.empty()) {
+        throwError("LOAD VECTOR file path cannot be empty", query);
+    }
+
+    std::string_view indexName = query->getIndexName();
```
rjb32: indexName can be const

query/analyzer/ReadStmtAnalyzer.cpp
```diff
@@ -378,6 +394,58 @@ void ReadStmtAnalyzer::analyze(EdgePattern* edgePattern) {
...
+    }
+
+    // Validate vector has elements
+    const ListExpr* queryVector = stmt->getQueryVector();
+    if (!queryVector || queryVector->empty()) {
+        throwError("VECTOR SEARCH query vector cannot be empty", stmt);
+    }
+
+    // Analyze each element in the query vector
+    for (Expr* elem : queryVector->getElements()) {
```
rjb32, line 434: elem can be const

query/analyzer/ReadStmtAnalyzer.cpp
```diff
@@ -378,6 +394,58 @@ void ReadStmtAnalyzer::analyze(EdgePattern* edgePattern) {
...
+        }
+    }
+
+    // Process YIELD clause - create VarDecl for 'ids' variable
+    const YieldClause* yield = stmt->getYield();
+    if (!yield) {
+        throwError("VECTOR SEARCH requires YIELD clause", stmt);
+    }
+
+    YieldItems* yieldItems = yield->getItems();
```
rjb32: yieldItems, yieldItemExpr, yieldItem can be const

query/interpreter/InterpreterContext.h
```diff
@@ -17,11 +21,13 @@ class InterpreterContext {
     InterpreterContext(LocalMemory* mem,
                        QueryCallbackV2 callback,
                        const ProcedureBlueprintMap* procedures,
+                       vec::VectorDatabase* vectorDatabase = nullptr,
```
rjb32: Don't put a default value for the vectorDatabase pointer

query/pipeline/PipelineBuilder.cpp
```diff
@@ -820,6 +825,91 @@ PipelineBlockOutputInterface& PipelineBuilder::addShowProcedures() {
     return output;
 }
 
+PipelineValueOutputInterface& PipelineBuilder::addCreateVectorIndex(std::string_view indexName,
+                                                                     vec::Dimension dimension,
```
rjb32: Fix argument alignment, they must aligned once below each other

query/pipeline/PipelineBuilder.cpp
```diff
@@ -820,6 +825,91 @@ PipelineBlockOutputInterface& PipelineBuilder::addShowProcedures() {
...
+    output.setValue(countCol);
+
+    _pendingOutput.setInterface(&output);
+    return output;
+}
+
+PipelineValuesOutputInterface& PipelineBuilder::addVectorSearch(std::string_view indexName,
+                                                                 uint64_t k,
+                                                                 std::vector<float> queryVector) {
+    VectorSearchProcessor* proc = VectorSearchProcessor::create(
```
rjb32: Keep the arguments on one line it's less ugly

query/pipeline/PipelineBuilder.cpp
```diff
@@ -820,6 +825,91 @@ PipelineBlockOutputInterface& PipelineBuilder::addShowProcedures() {
...
+    countCol->rename("count");
+    output.setValue(countCol);
+
+    _pendingOutput.setInterface(&output);
+    return output;
+}
+
+PipelineValuesOutputInterface& PipelineBuilder::addVectorSearch(std::string_view indexName,
+                                                                 uint64_t k,
+                                                                 std::vector<float> queryVector) {
```
rjb32: Dont pass vector by value! Use a reference

query/pipeline/PipelineBuilder.h
```diff
@@ -54,7 +55,10 @@ class PipelineBuilder {
 
     const PendingOutputView& getPendingOutput() const { return _pendingOutput; }
     PendingOutputView& getPendingOutput() { return _pendingOutput; }
-    PipelineOutputInterface* getPendingOutputInterface() const { return _pendingOutput.getInterface(); }
+
```
rjb32: No need to change this

query/pipeline/processors/CreateVectorIndexProcessor.cpp
```diff
@@ -0,0 +1,75 @@
...
+void CreateVectorIndexProcessor::reset() {
+}
+
+void CreateVectorIndexProcessor::execute() {
+    vec::VectorDatabase* vectorDb = _ctxt->getVectorDatabase();
+    if (!vectorDb) {
+        throw PipelineException("VectorDatabase not available");
+    }
+
+    vec::VectorResult<vec::VecLibID> result = vectorDb->createLibrary(
```
rjb32: result can be const

query/pipeline/processors/DeleteVectorIndexProcessor.cpp
```diff
@@ -0,0 +1,67 @@
...
+void DeleteVectorIndexProcessor::reset() {
+}
+
+void DeleteVectorIndexProcessor::execute() {
+    vec::VectorDatabase* vectorDb = _ctxt->getVectorDatabase();
+    if (!vectorDb) {
+        throw PipelineException("VectorDatabase not available");
+    }
+
+    vec::VectorResult<void> result = vectorDb->deleteLibrary(std::string(_indexName));
```
rjb32: result can be const

query/pipeline/processors/LoadVectorProcessor.cpp
```diff
@@ -0,0 +1,125 @@
...
+        if (values.size() != dimension) {
+            throw PipelineException(fmt::format(
+                "Vector dimension mismatch: expected {}, got {}",
+                dimension, values.size()));
+        }
+
+        batch.addPoint(id, std::span<const float>(values));
+    }
+
+    vec::VectorResult<void> result = accessor.addEmbeddings(batch);
```
rjb32: result can be const

query/pipeline/processors/ShowVectorIndexesProcessor.cpp
```diff
@@ -0,0 +1,73 @@
...
+        throw PipelineException("VectorDatabase not available");
+    }
+
+    using ColumnString = ColumnVector<types::String::Primitive>;
+    using ColumnUInt = ColumnVector<types::UInt64::Primitive>;
+
+    ColumnString* colNames = _outNames.getValues()->as<ColumnString>();
+    ColumnUInt* colDimensions = _outDimensions.getValues()->as<ColumnUInt>();
+
+    std::vector<std::string> indexNames = vectorDb->listLibraryNames();
```
rjb32: It should not need a copy, use a const reference instead

vector/VecLib.cpp
```diff
@@ -85,13 +85,17 @@ VectorResult<std::unique_ptr<VecLib>> VecLib::Loader::load(VecLibStorage& storag
 }
 
 VectorResult<void> VecLib::addEmbeddings(const BatchVectorCreate& batch) {
+    // Iterate with explicit index since the batch stores data at indices
```
rjb32: Explain more what was the problem before and why this change is necessary here

vector/VectorDatabase.cpp
```diff
@@ -156,6 +157,41 @@ VecLibAccessor VectorDatabase::getLibrary(std::string_view libName) {
...
+
+    auto it = _vecLibIDs.find(libName);
+    if (it == _vecLibIDs.end()) {
+        return VecLibWriteAccessor {};
+    }
+
+    return getLibraryForWrite(it->second);
+}
+
+std::vector<std::string> VectorDatabase::listLibraryNames() const {
```
rjb32: Do not return vector by value

query/pipeline/processors/LoadVectorProcessor.cpp
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        std::istringstream iss(line);
+        std::string token;
+
+        // Read ID
+        if (!std::getline(iss, token, ',')) {
+            throw PipelineException("Invalid vector file format: missing ID");
+        }
+        uint64_t id = std::stoull(token);
```
rjb32: id can be const

query/pipeline/processors/LoadVectorProcessor.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+        // Read vector values
+        std::vector<float> values;
```
rjb32: values can be reused and hoistered out of the loop

query/AST/expr/ListExpr.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace db {
+
+class CypherAST;
+
+class ListExpr : public Expr {
+public:
+    using Elements = std::vector<Expr*>;
+
+    void addItem(Expr* expr) { _elements.push_back(expr); }
```
rjb32: Put addItem in cpp file

query/pipeline/processors/CreateVectorIndexProcessor.cpp
```diff
@@ -0,0 +1,75 @@
...
+void CreateVectorIndexProcessor::reset() {
+}
+
+void CreateVectorIndexProcessor::execute() {
+    vec::VectorDatabase* vectorDb = _ctxt->getVectorDatabase();
+    if (!vectorDb) {
+        throw PipelineException("VectorDatabase not available");
+    }
+
+    const vec::VectorResult<vec::VecLibID> result = vectorDb->createLibrary(
```
rjb32: Put each argument one below the other

vector/VectorDatabase.h
```diff
@@ -42,6 +43,13 @@ class VectorDatabase {
     [[nodiscard]] VecLibAccessor getLibrary(const VecLibID& libID);
     [[nodiscard]] VecLibAccessor getLibrary(std::string_view libName);
 
+    [[nodiscard]] VecLibWriteAccessor getLibraryForWrite(const VecLibID& libID);
+    [[nodiscard]] VecLibWriteAccessor getLibraryForWrite(std::string_view libName);
+
+    void listLibraryNames(std::vector<std::string>& out) const;
+
+    std::shared_mutex& getMutex() { return _mutex; }
```
rjb32: This is particularly disgusting to return a mutex. Don't do that

rjb32:
@LucLabarriere I rebased the branch against the latest changes and cleaned it up. It's all clean now I think.
See the docs/ first and the tests.
I also added a deterministic seed at vector database initialisation.

query/analyzer/ReadStmtAnalyzer.cpp
```diff
@@ -404,6 +418,60 @@ void ReadStmtAnalyzer::analyze(EdgePattern* edgePattern) {
...
+    }
+
+    // Validate vector has elements
+    const ListExpr* queryVector = stmt->getQueryVector();
+    if (!queryVector || queryVector->empty()) {
+        throwError("VECTOR SEARCH query vector cannot be empty", stmt);
+    }
+
+    // Analyze each element in the query vector
+    for (Expr* const elem : queryVector->getElements()) {
```
LucLabarriere: Do you want to keep the const on the pointer here?

query/pipeline/processors/LoadVectorProcessor.cpp
```diff
@@ -0,0 +1,131 @@
...
+    // Get write accessor for exclusive access
+    vec::VecLibWriteAccessor accessor = vectorDb->getLibraryForWrite(std::string(_indexName));
+    if (!accessor.isValid()) {
+        throw PipelineException(fmt::format("Vector index '{}' not found", _indexName));
+    }
+
+    // Parse file and load embeddings
+    // Expected format: CSV with id,dim1,dim2,...,dimN
+    const std::string filePath{_filePath};
+    std::ifstream file{filePath};
```
LucLabarriere: style maybe, space between varname and braces

query/pipeline/processors/VectorSearchProcessor.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    // Execute search
+    vec::VectorSearchQuery query(accessor.metadata()._dimension);
+    query.setVector(_queryVector);
+    query.setMaxResultCount(_k);
+
+    vec::VectorSearchResult result;
+    vec::VectorResult<void> searchResult = accessor.search(query, result);
+
+    if (!searchResult.has_value()) {
```
LucLabarriere: Just `if (!searchResult)` would be clearer, this result doesn't contain a value

regress/vector_serialization/run.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+# Set TURING_HOME to the install root (two levels up from regress/test_name/)
+export TURING_HOME=$(cd "$SCRIPT_DIR/../.." && pwd)
```
LucLabarriere: This seems risky, is this needed for anything? It should already be set by the run_regress script, here it might override it

regress/vector_serialization/main.py
```diff
@@ -0,0 +1,207 @@
...
+
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+VECTORS_FILE = os.path.join(SCRIPT_DIR, "vectors.csv")
+
+
+def spawn_turingdb():
+    print(f"- {GREEN}Starting turingdb{NC}")
+    turing_home = os.environ.get('TURING_HOME', '')
+    turingdb_bin = os.path.join(turing_home, 'bin', 'turingdb') if turing_home else 'turingdb'
+    return subprocess.Popen(f"exec {turingdb_bin} -demon -turing-dir .turing", shell=True)
```
LucLabarriere: No sure this is needed, the run_regress script already ensures turingdb is in path

regress/vector_serialization/main.py
```diff
@@ -0,0 +1,207 @@
...
+
+        dimensions = list(indexes['dimension'])
+        idx = names.index('test_vectors')
+        if dimensions[idx] != 8:
+            raise Exception(f"Expected dimension 8, got {dimensions[idx]}")
+
+        # Load vectors from file
+        print(f"- {BLUE}Loading vectors from file{NC}")
+        client.query(f'LOAD VECTOR FROM "{VECTORS_FILE}" IN test_vectors')
+        print("  Vectors loaded successfully")
```
LucLabarriere, line 111: I don't understand, how does turingdb find the vectors file? It should have been put in the .turing/data folder

vector/VecLib.cpp
```diff
@@ -85,13 +85,20 @@ VectorResult<std::unique_ptr<VecLib>> VecLib::Loader::load(VecLibStorage& storag
...
-    for (const auto& data : batch) {
+    for (auto it = batch.begin(); it != batch.end(); ++it, ++signature) {
+        const auto& data = *it;
         if (data._externalIDs.empty()) {
             continue;
         }
 
-        auto& shard = _shardCache->getShard(_metadata, signature++);
+        // Use the actual index (signature) where data is stored
+        auto& shard = _shardCache->getShard(_metadata, signature);
```
LucLabarriere: Oh that was a bug

regress/vector_serialization/main.py
LucLabarriere: I see, it actually searches for the file in the current directory. I think this is risky, if creates a vulnerability on the server if it can escape the data dir. Also it doesn't make too much sense in a 'server <-> client' interaction mode since the python client will not be able to find the file because the server will be looking for it relative to its own workind dir, instead of that of the python client. I think it should be searched for in .turing/data for now, and we can add a config to allow the server to look for files outside the data dir in the future

query/pipeline/processors/LoadVectorProcessor.cpp
rjb32: The initialisation style is also wrong. It should be parenthesis.

query/analyzer/ReadStmtAnalyzer.cpp
rjb32: No indeed

query/pipeline/processors/VectorSearchProcessor.cpp
rjb32: Yes indeed

regress/vector_serialization/run.sh
rjb32: Yes that should not be done

regress/vector_serialization/main.py
rjb32: Yes setting TURING_HOME should not be done here

regress/vector_serialization/main.py
rjb32: Yes indeed

docs/vector/VECTOR_INTEGRATION.md
```diff
@@ -0,0 +1,110 @@
...
+to make users observe and rely too much on TuringDB entity IDs as they are not guaranteed to be stable.
+
+* The numerical ID could also be a foreign key which is referencing data in an external database as well,
+such as documents or other records.
+
+-> Thus we don't know the ID to be used is the user's choice. It may be an arbitrary property.
+
+-> Users may want to come with their own embeddings
+
+-> We obviously to generate embeddings at each commit
```
cyrusknopf, line 29: Does this mean each time there is a commit, some embeddings are computed?
Or that any time embeddings are entered into the DB, that makes a new commit?

query/analyzer/ExprAnalyzer.cpp
```diff
@@ -68,6 +68,10 @@ void ExprAnalyzer::analyzeExpr(Expr* expr) {
         case Expr::Kind::INDEX:
             analyzeIndexExpr(static_cast<IndexExpr*>(expr));
             break;
+        case Expr::Kind::LIST:
+            // List expressions are handled specially in vector search context
+            // For now, just mark them as having no specific type
+            break;
```
cyrusknopf, line 74: Once we support lists later, will we be able to tell easily when we are in a "vector search context" and when we are in a regular list context?

query/analyzer/ReadStmtAnalyzer.cpp
```diff
@@ -68,6 +70,10 @@ void ReadStmtAnalyzer::analyze(Stmt* stmt) {
             analyze(static_cast<LoadCSVStmt*>(stmt));
             break;
 
+        case Stmt::Kind::VECTOR_SEARCH:
+            analyze(static_cast<const VectorSearchStmt*>(stmt));
+            break;
```
cyrusknopf:
```suggestion
        case Stmt::Kind::VECTOR_SEARCH:
            analyze(static_cast<const VectorSearchStmt*>(stmt));
        break;
```

query/analyzer/ReadStmtAnalyzer.cpp
```diff
@@ -404,6 +418,60 @@ void ReadStmtAnalyzer::analyze(EdgePattern* edgePattern) {
...
+        throwError("VECTOR SEARCH YIELD clause cannot be empty", stmt);
+    }
+
+    for (SymbolExpr* yieldItemExpr : *yieldItems) {
+        const Symbol* yieldItem = yieldItemExpr->getSymbol();
+
+        if (yieldItem->getName() != "ids") {
+            throwError(fmt::format("VECTOR SEARCH only supports YIELD ids, got '{}'",
+                                   yieldItem->getName()), stmt);
+        }
```
cyrusknopf, line 460: Why is it necessary to hardcode the name of the yield item as `ids`?

query/AST/expr/ListExpr.h
```diff
@@ -0,0 +1,37 @@
...
+    Elements::const_iterator begin() const { return _elements.begin(); }
+    Elements::const_iterator end() const { return _elements.end(); }
+
+private:
+    Elements _elements;
+
+    ListExpr()
+        : Expr(Kind::LIST)
+    {
+    }
```
cyrusknopf, line 32: Should be in .cpp file?

query/pipeline/processors/DeleteVectorIndexProcessor.cpp
```diff
@@ -0,0 +1,67 @@
...
+                                            _indexName, result.error().fmtMessage()));
+    }
+
+    using ColumnString = ColumnConst<types::String::Primitive>;
+    ColumnString* colName = _outName.getValue()->as<ColumnString>();
+    colName->set(_indexName);
+
+    _outName.getPort()->writeData();
+    finish();
+}
```
cyrusknopf, line 67: Does FAISS handle concurrency between libraries? What if one client starts reading from a library which is then deleted by another client during the read?

query/pipeline/processors/LoadVectorProcessor.cpp
```diff
@@ -0,0 +1,143 @@
...
+    vec::Dimension dimension = accessor.metadata()._dimension;
+    vec::BatchVectorCreate batch;
+    accessor.prepareCreateBatch(&batch);
+
+    std::string line;
+    std::string token;
+    std::vector<float> values;
+    values.reserve(dimension);
+
+    for (;;) {
```
cyrusknopf: Would we prefer a `while` here?

query/pipeline/processors/LoadVectorProcessor.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+        while (std::getline(iss, token, ',')) {
+            values.push_back(std::stof(token));
+        }
+
+        if (values.size() != dimension) {
+            throw PipelineException(fmt::format(
+                "Vector dimension mismatch: expected {}, got {}",
+                dimension, values.size()));
+        }
```
cyrusknopf, line 122: If `stoull` or `PipelineException` throws here, is the FAISS vector DB in a valid state? do the calls to `batch.addPoint` need to be undone?

query/pipeline/processors/ShowVectorIndexesProcessor.cpp
```diff
@@ -0,0 +1,74 @@
...
+    ColumnUInt* colDimensions = _outDimensions.getValues()->as<ColumnUInt>();
+
+    vectorDb->listLibraryNames(_indexNames);
+    for (const std::string& name : _indexNames) {
+        vec::VecLibAccessor accessor = vectorDb->getLibrary(name);
+        if (accessor.isValid()) {
+            colNames->push_back(name);
+            colDimensions->push_back(accessor.metadata()._dimension);
+        }
+    }
```
cyrusknopf, line 69: In what cases is `accessor` not valid?

docs/vector/VECTOR_INTEGRATION.md
rjb32: Embeddings and the vector index are outside the commit system exactly for this reason:
* To avoid to have a new commit when embeddings change because the user may have changed their pipelines that compute them , in a way that has nothing to do with the graph per se
* A vector index is at a high level just a map (numerical ID -> vector). So it is not strictly speaking only for nodes. It can be anything that has a numerical ID
* A vector index is just a facility on the side with respect to the graph
* We want users to be able to create graphs and subgraphs and not have to recompute embeddings when a subgraph is just a subset of an existing graph with existing embeddings

Yes this doc is a bit wrong now I have to update it.

LucLabarriere: It means that embeddings are kind of "tied" to a commit, in the sense that, if we update a node after you computed its embeddings, then the embeddings become "invalid". The vector database is literally just a `some id -> vector` mapping, it's completely agnostic of what it is computed against. `some id` may be a NodeID, and EdgeID, or edge a node property, an edge property, etc. The vector can be anything, including a 3d point in space if you want. The user creates a vector index and decides what it should represent, if it is tied to a commit, or something else. It is completely decoupled from the graph storage

query/analyzer/ExprAnalyzer.cpp
rjb32: When we support it later, we will just need to check if it is a list literal or an more complex list expression to apply the treatment directly at generation time if it is a list literal

query/analyzer/ReadStmtAnalyzer.cpp
LucLabarriere: Do you mean why we can only do YIELD ids and not YIELD myids? If that's what you mean, that's the Cypher syntax, the YIELD expr uses the named return value (hard coded), if you want to rename it you use the AS keyword:
```
YIELD ids AS myids
```

query/pipeline/processors/DeleteVectorIndexProcessor.cpp
LucLabarriere: Normally, everything is wrapped properly in the VectorDatabase class, the concurrency is handled by us with shared mutexes

docs/vector/VECTOR_INTEGRATION.md
cyrusknopf: I don't understand how 
> "if we update a node after you computed its embeddings, then the embeddings become 'invalid'"

and

> "embeddings are kind of 'tied' to a commit"

are compatible with

> "the user creates a vector index and decides what it should represent, if it is tied to a commit, or something else"


If we assume we do have some way to update a node, then why does that invalidate the embedding? (aside from the obvious case where an embedding points to a TuringDB NodeID, and then that NodeID changes as a result of the update; I'm talking more about an embedding pointing to a user-defined ID that need not change after we "update" the node. Unless of course, the "update" to the node is an update such that rerunning the embedding process on that node would produce a different result, but is this always the case a priori?).

And how are embeddings tied to a commit, if the user can chose what the index represents, which is possibly not a commit?

query/pipeline/processors/LoadVectorProcessor.cpp
LucLabarriere: I was wondering too, I got the error while testing it trying to use a non integer id, it raised an exception, caught by TuringDB's query execution -> so fine. In the VectorDatabase code, nothing is done until we submit the vector batch (line 131)

docs/vector/VECTOR_INTEGRATION.md
LucLabarriere: Yeah I explained poorly, the vector database is really its own thing, completely separate from the graph dbms. The user can even create vector indexes without even having any entities in the database. The user decides what they want to compute embeddings for and send the embeddings, attached to some numerical id.

So he could add a property "vector_id" to nodes, monotically increasing, then compute the embeddings on his side in python, using a template string like this:

```python
client.query("CREATE VECTOR INDEX myindex WITH DIMENSION 1024 METRIC COSINE")

templatestr = """
Node id: {n.vector_id}
Node properties:
- Title: {n.title}
- Description: {n.desc}
- Content: {n.content}
"""

data = client.query('MATCH (n) RETURN n.vector_id, n.title, n.desc, n.content')
for i in range(data.shape[0]):
    toencode = templatestr.format(data['n.vector_id'][i],
                                  data['n.title'][i],
                                  data['n.desc'][i],
                                  data['n.content'][i])
    embeddings = compute_embeddings(toencode)
    csvfile.write([data['n.vector_id']] + embeddings) # Write the csv containing
                                                      # the embeddings
client.query('LOAD VECTOR FROM "vectors.csv" IN myindex')

#Then use it with some query:

result = client.query(f"""
    VECTOR SEARCH IN myindex FOR 8 [0.552, -1.241, ...] YIELD ids
    MATCH (n) WHERE n.vector_id = ids RETURN n.vector_id, n.title
    """)
```

query/AST/expr/ListExpr.h
rjb32: Yes indeed

query/pipeline/processors/DeleteVectorIndexProcessor.cpp
rjb32: deleteLibrary acquires a unique lock over the entire VectorDatabase because indeed it has global consequences for VectorDatabase

query/pipeline/processors/LoadVectorProcessor.cpp
rjb32: Yes indeed

query/pipeline/processors/ShowVectorIndexesProcessor.cpp
rjb32: Because VecLibAccessor has a default constructor which is invalid by default

PR #335, opened 2026-01-26
Implement Shortest Path Functionality

ghost:
V1 of shortest path processor using uni-directional djistraka.

This works by wiring the inputs of the plan graph branches containing the source and target variables to a shortest path node. For now this means:

- Common Ancestor Joins will not work well with ShortestPath nodes
- We can't access the values in the source or target branches in our return projection.

I feel that there will be some iteration on the design of the input and outputs of the processor so the above two points are also likely to be affected in the near future.


query/AST/CypherASTDumper.cpp
```diff
@@ -184,6 +185,13 @@ void CypherASTDumper::dump(std::ostream& out, const SinglePartQuery* query) {
                     dump(out, setStmt);
                 }
                 break;
+
+                case Stmt::Kind::SHORTESTPATH: {
+                    const ShortestPathStmt* shortestPathStmt = static_cast<const ShortestPathStmt*>(stmt);
+                    out << "    _" << std::hex << query << " ||--o{ _" << std::hex << shortestPathStmt<< " : \"\"\n";
+                    //dump(out, shortestPathStmt);
```
cyrusknopf:
```suggestion
```

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include "dataframe/Dataframe.h"
+
+using namespace db;
+
+template <SupportedType T>
+ShortestPathProcessor<T>::ShortestPathProcessor(LocalMemory* mem,
+                                             ColumnTag sourceTag,
+                                             ColumnTag targetTag,
+                                             const PropertyType& edgeType) 
+    :_mem(mem),
```
cyrusknopf:
```suggestion
    : _mem(mem),
```

query/pipeline/processors/ShortestPathProcessor.h
```diff
@@ -0,0 +1,103 @@
...
+    ColumnIndices* _outputIndices {nullptr};
+    std::unique_ptr<GetOutEdgesChunkWriter> _getOutEdgesWriter;
+
+    ColumnIndices* _propertyIndices {nullptr};
+    ColumnVector<edgePropType>* _properties {nullptr};
+    std::unique_ptr<GetPropertiesChunkWriter<EdgeID, T>> _getPropertiesWriter;
+
+    std::unordered_set<NodeID> _targetNodes;
+    std::priority_queue<DjistrakaNode<edgePropType>, std::vector<DjistrakaNode<edgePropType>>, DjistrakaNodeCompartor<edgePropType>> _heap;
+    std::unordered_map<NodeID, std::tuple<NodeID, EdgeID, edgePropType>> _heapValueMap;
```
cyrusknopf: I think these two should be aliased

query/pipeline/PipelineBuilder.cpp
```diff
@@ -840,3 +881,16 @@ template PipelineValuesOutputInterface& PipelineBuilder::addGetPropertiesWithNul
...
+                                                                                           ColumnTag,
+                                                                                           const PropertyType&,
+                                                                                           NamedColumn*&,
+                                                                                           NamedColumn*&);
+template PipelineBlockOutputInterface& PipelineBuilder::addShortestPath<db::types::UInt64>(PipelineOutputInterface*,
+                                                                                           ColumnTag,
+                                                                                           ColumnTag,
+                                                                                           const PropertyType&,
+                                                                                           NamedColumn*&,
+                                                                                           NamedColumn*&);
```
cyrusknopf, line 906: Do we no support it for `types::Int64`?

system/SystemManager.cpp
```diff
@@ -462,9 +462,11 @@ ChangeResult<Transaction> SystemManager::openTransaction(std::string_view graphN
     // if hash != head: Requesting a read on a specific commit (either pending or frozen)
     if (auto tx = change->openReadTransaction(commitHash); tx.isValid()) {
         // Reading pending commit
+        spdlog::info("opening read");
```
cyrusknopf:
```suggestion
 ```

File: system/SystemManager.cpp [hunk 6]
```diff
@@ -462,9 +462,11 @@ ChangeResult<Transaction> SystemManager::openTransaction(std::string_view graphN
...
         return tx;
     } else if (auto tx = graph->openTransaction(commitHash); tx.isValid()) {
         // Reading frozen commit
+        spdlog::info("opening frozen");
```
cyrusknopf:
```suggestion
```

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,199 @@
...
+                                                                                 _outputEdges);
+    _getPropertiesWriter->setOutput(_properties);
+    _getPropertiesWriter->setIndices(_propertyIndices);
+
+    markAsPrepared();
+}
+
+template <SupportedType T>
+void ShortestPathProcessor<T>::reset() {
+    markAsReset();
```
cyrusknopf, line 92: I'm not sure when this is called, but `GetOutEdgesProcessor` resets its iterator here. Perhaps you want to do a similar thing?

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    if (_source.getPort()->hasData()) {
+        const Dataframe* leftDf = _source.getDataframe();
+        auto* col = leftDf->getColumn<ColumnNodeIDs>(_sourceColumn);
+
+        for (const auto val : *col) {
+            _heap.push({val, NodeID(), EdgeID(), 0});
+            _heapValueMap.insert({
+                val, {NodeID(), EdgeID(), 0}
+            });
```
cyrusknopf, line 123: What are these values?

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+            _heap.pop();
+
+            // We only need to confirm if the distance value is stale
+            // as we only a new value to the heap when we find a shorter
+            // path to a node
+            if (std::get<2>(_heapValueMap[val.id]) != val.distance) {
+                // remove stale value
+                continue;
+            }
```
cyrusknopf: How does this remove the stale value?

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,199 @@
...
+        auto edge = _heap.top().edge;
+        pathVec.push_back(_heap.top().id.getValue());
+
+        while (lastNode.isValid()) {
+            pathVec.push_back(edge.getValue());
+            pathVec.push_back(lastNode.getValue());
+
+            const auto& pathInfo = _heapValueMap[lastNode];
+            lastNode = std::get<0>(pathInfo);
+            edge = std::get<1>(pathInfo);
```
cyrusknopf: Did you consider a struct instead of a tuple as map key? the calls to `get` are a bit ambiguous

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    // execute the algorithm once we have all the data
+    if (_source.getPort()->isClosed() &&
+        _target.getPort()->isClosed() &&
+        !_heap.empty()) {
+        while (!_heap.empty()) {
+            const auto val = _heap.top();
+
+            if (_targetNodes.contains(val.id)) {
+                // target found
```
cyrusknopf: So as soon as we find a target, we break and write output? So the shortest path only ever returns a single path?

query/pipeline/PipelineBuilder.cpp
ghost: Djistraka doesn't really work with negative weighted edges so we shouldn't support signed integers.

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,199 @@
...
+template <SupportedType T>
+void ShortestPathProcessor<T>::reset() {
+    markAsReset();
+}
+
+template <SupportedType T>
+void ShortestPathProcessor<T>::execute() {
+    if (_target.getPort()->hasData()) {
+        const Dataframe* rightDf = _target.getDataframe();
+        auto* col = rightDf->getColumn<ColumnNodeIDs>(_targetColumn);
```
cyrusknopf: `getColumn` can return `nullptr` in a few cases. Probably worth checking at all callsites

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+        for (const auto val : *col) {
```
cyrusknopf:
```suggestion
        for (const NodeID val : *col) {
```

query/pipeline/processors/ShortestPathProcessor.cpp
ghost: These are the default prevNode prevEdge and distance values for the source nodes inserted into the heap.

query/pipeline/processors/ShortestPathProcessor.cpp
ghost: we have already popped the value from the heap. I think the comment is a bit misleading I will change it.

query/pipeline/PipelineBuilder.cpp
cyrusknopf: Doubles are also signed though?

query/pipeline/processors/ShortestPathProcessor.cpp
ghost: I was thinking about that it just felt a bit weird to have 3/4 value in the DijkstraNode struct in a separate struct.

query/pipeline/processors/ShortestPathProcessor.cpp
ghost: yep, I think a future direction is to allow users to specify the top K nodes they want, and we break only after the Kth path has been found.

query/pipeline/PipelineBuilder.cpp
ghost: hmm good point actually! Might be worth adding in Int64s. What do you think @rjb32 ?

query/pipeline/processors/ShortestPathProcessor.cpp
ghost: Shouldn't matter as reset() is called after the processor has been marked as finished. This processor is only marked as finished once the query is over as it buffers for all the target and source nodes before performing the Dijkstra search.

query/pipeline/PipelineBuilder.cpp
rjb32: Yes indeed it's worth supporting int64 weights

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,214 @@
...
+                // target found
+                break;
+            }
+
+            _heap.pop();
+
+            // We only need to confirm if the distance value is stale
+            // as we only a new value to the heap when we find a shorter
+            // path to a node
+            if ((_heapValueMap[val.id].distance) != val.distance) {
```
rjb32: The use of the operator[] for a map read may introduce unintentional modifications to the map. We should prefer using find for a map read.

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+            _input->clear();
+            _input->push_back(val.id);
+            _getOutEdgesWriter->reset();
+            _getOutEdgesWriter->fill(SIZE_MAX);
+
+            _getPropertiesWriter->reset();
+            _getPropertiesWriter->fill(SIZE_MAX);
+
+            for (size_t i = 0; i < _properties->size(); ++i) {
```
rjb32: Add comments to delimit a bit these phases

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,214 @@
...
+                        _heap.push({outputNodeId, val.id, outputEdgeId, dist});
+                        _heapValueMap[outputNodeId] = {val.id, outputEdgeId, dist};
+                    }
+                }
+            }
+        }
+
+        const Dataframe* outDf = _out.getDataframe();
+
+        ColumnVector<edgePropType>* distCol = outDf->getColumn<ColumnVector<edgePropType>>(_distTag);
```
rjb32: Same from there, explain more what you are doing with comments

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        distCol->push_back(_heap.top().distance);
+        auto& pathVec = pathCol->emplace_back();
+
+        auto lastNode = _heap.top().prevNode;
+        auto edge = _heap.top().edge;
+        pathVec.push_back(_heap.top().id.getValue());
+
+        while (lastNode.isValid()) {
```
rjb32: Add comments for this loop

test/query/queries/ShortestPathQueryTest.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    auto query(std::string_view query) {
+        auto callback = [](const Dataframe*) {};
+        auto res = _db->query(query, _graphName, &_env->getMem(), callback,
+                              CommitHash::head(), ChangeID::head());
+        return res;
+    }
+};
+
+// =============================================================================
```
rjb32, line 166: Please add non-error tests

tools/turingdb/TuringShell.cpp
```diff
@@ -367,6 +369,28 @@ void tabulateWrite(tabulate::RowStream& rs, const std::optional<T>& value) {
     }
 }
 
+void tabulateWrite(tabulate::RowStream& rs, db::Path path) {
```
rjb32: Should not path be taken as a const reference

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+        _source.getPort()->consume();
+    }
+
+    // execute the algorithm once we have all the data
+    if (_source.getPort()->isClosed() &&
+        _target.getPort()->isClosed() &&
+        !_heap.empty()) {
+        while (!_heap.empty()) {
+            const auto val = _heap.top();
```
rjb32: It's quite important for readability to not use auto here and write explicitly that it is a DijkstraNode

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+        ColumnVector<Path>* pathCol = outDf->getColumn<ColumnVector<Path>>(_pathTag);
+        if (!pathCol) {
+            throw TuringException("Could not find path column");
+        }
+
+        distCol->push_back(_heap.top().distance);
+        auto& pathVec = pathCol->emplace_back();
+
+        auto lastNode = _heap.top().prevNode;
```
rjb32: Same do not use auto here for readability

storage/GraphPath.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include "ID.h"
+
+namespace db {
+using Path = std::vector<EntityID>;
```
cyrusknopf, line 7: Does this mean we construct EntityIDs from NodeIDs and EdgeIDs (alternating)?

ghost: Yep the path is represented as alternate Nodes and edges.

query/pipeline/processors/ShortestPathProcessor.cpp
cyrusknopf: What do you mean?

query/pipeline/processors/ShortestPathProcessor.cpp
cyrusknopf: Are they used to designate some invalid state? since they are default initialised? I can't see calls to `isValid`. Also is it not customary to initialise distances to infinity in Dijkstra instead of 0? (maybe im misunderstanding what this heap is for)

query/pipeline/processors/ShortestPathProcessor.cpp
ghost: Just worried that I would be repeating myself, but I have changed this.

query/pipeline/processors/ShortestPathProcessor.cpp
ghost: We do call to isValid() at line 236. In traditional Dijkstra with a table we mark all non-source nodes as infinity but the source is marked as 0. When working with a heap (and without knowing the entire graph) we don't need to mark the undiscovered nodes at infinity. 

In the code snippet above we are inserting the source nodes into the heap. So the distance should be 0 and the previous nodeID and EdgeID should be invalid (for semantics as well as to help us know when we have reached the source when we are recreating the shortest paths.)

cyrusknopf: Oh makes sense I missed the fac this was just pushing source nodes

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,254 @@
...
+                val, {NodeID(), EdgeID(), 0}
+            });
+        }
+
+        _source.getPort()->consume();
+    }
+
+    // execute the algorithm once we have all the data
+    if (_source.getPort()->isClosed() &&
+        _target.getPort()->isClosed()) {
```
cyrusknopf: Why not reverse the condition and early return to reduce the indentation on this whole block

query/pipeline/processors/ShortestPathProcessor.cpp
cyrusknopf: Is it not always marked as finished? if the condition on 130 is not true it falls through to an unconditional finish() call

ghost: Ah sorry I was a bit confused, you are right that the processor is marked as finish on every call. But actually the GetOutEdges processor marks as finished once it has outputted all its values. So it is ready to get reset the iterator for the next set of inputs.

The shortest path has to buffer all the inputs until it runs its main execution loop only 1 time. So we shouldn't mark the iterator to reset as it doesn't really do anything as it will always be invalid until we start adding inputs to it in the main execution loop.

Hope that makes sense!

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,254 @@
...
+            }
+
+            // Set up the input columns and iterators
+            _input->clear();
+            _input->push_back(val.id);
+            _getOutEdgesWriter->reset();
+            _getOutEdgesWriter->fill(SIZE_MAX);
+
+            _getPropertiesWriter->reset();
+            _getPropertiesWriter->fill(SIZE_MAX);
```
cyrusknopf: Sry I can't find the definition, what is SIZE_MAX?

query/pipeline/processors/ShortestPathProcessor.cpp
cyrusknopf: Okay ya got it, I was a bit mistaken for what `reset` should do, thanks

query/pipeline/processors/ShortestPathProcessor.cpp
```diff
@@ -0,0 +1,254 @@
...
+                // target found
+                break;
+            }
+
+            _heap.pop();
+
+            // We only need to confirm if the distance value is stale
+            // as we only add a new value to the heap when we find a shorter
+            // path to a node.
+            const auto& it = _heapValueMap.find(val.id);
```
rjb32: const auto it

query/pipeline/processors/ShortestPathProcessor.cpp
ghost: This is a macro that defines that maximum size of a size_t - it is used here to ensure that we expand all the results of the iterator in one chunk. It is used quite a bit in C but I just checked and it is a part of the C++ standard as well.

cyrusknopf: Might be worth adding a comment as to why we are using SIZE_MAX as the argument here. Everywhere else in the codebase I believe the argument is always CHUNK_SIZE

PR #338, opened 2026-01-27
Return wildcard feature

### LINKED ISSUE #260: Analyzer error with CREATEs and wildcard
**Reported by:** @cyrusknopf

```
$ turingdb -in-memory -reset-default
```

```
change new
checkout change-0
CREATE (n:Node) RETURN *
```

```
[2026-01-08 12:14:15] [error] ANALYZE_ERROR: Unexpected exception: std::get: wrong index for variant
```

`gdb` backtrace:

```
Thread 1 "turingdb" hit Catchpoint 1 (exception thrown), 0x00007384b71974a1 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007384b71974a1 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x000058cecf8fe709 in std::__throw_bad_variant_access (__what=__what@entry=0x58cecfbc7078 "std::get: wrong index for variant") at /usr/include/c++/11/variant:1314
#2  0x000058cecf906f8e in std::__throw_bad_variant_access (__valueless=false) at /usr/include/c++/11/variant:1322
#3  std::get<0ul, std::vector<db::Expr*, std::allocator<db::Expr*> >, db::Projection::All> (__v=std::variant<std::vector<db::Expr*>, db::Projection::All> [index 1] = {...}) at /usr/include/c++/11/variant:1703
#4  std::get<std::vector<db::Expr*, std::allocator<db::Expr*> >, std::vector<db::Expr*, std::allocator<db::Expr*> >, db::Projection::All> (__v=std::variant<std::vector<db::Expr*>, db::Projection::All> [index 1] = {...}) at /usr/include/c++/11/variant:1141
#5  db::Projection::items (this=0x58cef67cf820) at /home/dev/turingdb/query/AST/Projection.h:55
#6  db::CypherAnalyzer::analyze (this=0x7ffccd40a270, returnSt=<optimized out>) at /home/dev/turingdb/query/analyzer/CypherAnalyzer.cpp:157
#7  0x000058cecfa2261f in db::CypherAnalyzer::analyze (this=this@entry=0x7ffccd40a270) at /home/dev/turingdb/query/analyzer/CypherAnalyzer.cpp:57
#8  0x000058cecfa213e8 in db::QueryInterpreterV2::execute (this=this@entry=0x7ffccd40a9d0, ctxt=..., query=..., graphName="default") at /home/dev/turingdb/query/interpreter/QueryInterpreterV2.cpp:76
#9  0x000058cecf963e69 in db::TuringDB::query(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, db::LocalMemory*, std::function<void (db::Dataframe const*)>, db::TemplateCommitHash<0>, db::TemplateCommitHash<1>) (this=this@entry=0x7ffccd40af90, query=..., graphName="default",
    mem=0x7ffccd40b010, callback=..., commit=..., change=...) at /home/dev/turingdb/db/TuringDB.cpp:86
#10 0x000058cecf93462f in db::TuringShell::processLine (this=0x7ffccd40b5a0, line="CREATE (n:Node) RETURN *") at /usr/include/c++/11/string_view:137
#11 0x000058cecf934e45 in db::TuringShell::startLoop (this=this@entry=0x7ffccd40b5a0) at /home/dev/turingdb/tools/turingdb/TuringShell.cpp:303
#12 0x000058cecf920903 in main (argc=<optimized out>, argv=<optimized out>) at /home/dev/turingdb/tools/turingdb/TuringDBTool.cpp:127
(gdb)
```

LucLabarriere:
Support for `RETURN *` feature. Possible cases:

- `MATCH ... RETURN *`: returns all named variables
- `MATCH ... RETURN *, n`: `n` is already returned by `*` so the explicit `RETURN n` has no effect
- `MATCH ... RETURN *, n.name`: `n.name` is an additional return field
- `MATCH ... RETURN n, *`: forbidden by the parser
- `CREATE ... RETURN *`: same rules as `MATCH` queries
- `CALL ... YIELD ... RETURN *`: same rules as `MATCH queries`

Fixes #260 and #358

## Unit tests

```cypher
// 1.
CALL db.propertyTypes() YIELD propertyType, valueType RETURN *

// 2.
MATCH (n)
CALL db.propertyTypes() YIELD propertyType, valueType RETURN *, n.name

// 3.
CALL db.propertyTypes() YIELD *

// 4.
CALL db.propertyTypes() YIELD * RETURN * // fails

// 5.
MATCH (n) RETURN n, * // fails

// 6.
MATCH (n:Person)-->(m:Interest)
CALL db.labels() YIELD label, id
RETURN *, m.name

// 7.
MATCH (n) RETURN *

// 8.
MATCH (n) RETURN *, n.name

// 9.
MATCH (n:Person)
MATCH (m:Interest)
RETURN *
```

### DIFF
--- a/query/AST/Projection.h
+++ b/query/AST/Projection.h
@@ -85,6 +85,8 @@ private:
     bool _returningAll {false};
 
     Items _items;
+
+    // Maps a VarDecl*/Expr* to its column name
     std::unordered_map<uintptr_t, std::string_view> _names;
     std::unordered_set<std::string_view> _namesSet;
 


LucLabarriere:
@rjb32

query/analyzer/CypherAnalyzer.h
```diff
@@ -55,6 +56,7 @@ class CypherAnalyzer {
     CypherAST* _ast {nullptr};
     GraphView _graphView;
     const GraphMetadata& _graphMetadata;
+    const QueryCommand* _currentQuery {nullptr};
```
rjb32: Why is it necessary now?

query/analyzer/ReadStmtAnalyzer.cpp
```diff
@@ -115,7 +115,7 @@ void ReadStmtAnalyzer::analyze(const CallStmt* callStmt) {
 
     // Step 3. Analyze YIELD clause
     const YieldClause* yield = callStmt->getYield();
-    if (yield == nullptr) {
+    if (yield == nullptr || yield->getItems() == nullptr) {
         // No yield OR YIELD *.
```
rjb32: Should that comment be changed now?

query/AST/Projection.h
```diff
@@ -43,21 +50,29 @@ class Projection {
...
     Skip* getSkip() const { return _skip; }
     OrderBy* getOrderBy() const { return _orderBy; }
 
     const Items& items() const {
-        return std::get<Items>(_items);
+        return _items;
     }
 
-    void add(Expr* Expr);
+    void pushBackExpr(Expr* expr);
```
rjb32: Is this necessary?

query/AST/Projection.h
```diff
@@ -43,21 +50,29 @@ class Projection {
...
+    void pushFrontDecl(VarDecl* decl);
+
+    void setReturnAll() { _returningAll = true; }
 
-    void setAll() { _items = All {}; }
+    const std::string_view* getName(const Expr* item) const;
```
rjb32: Why return a const std::string_view*?

query/analyzer/CypherAnalyzer.h
LucLabarriere: We need to retrieve all the named variables in the current context, which is the context of the current query command (single part query, e.g. `MATCH ... MATCH ... RETURN ...`)

LucLabarriere: I could store the current context actually, might be better

query/analyzer/ReadStmtAnalyzer.cpp
LucLabarriere: No but it's misleading, I'll change it, it means that 'YIELD *' is equivalent to having no YIELD at all in our implementation (neo4j distinguishes them by saying that YIELD * also returns deprecated columns)

query/AST/Projection.h
LucLabarriere: That's only because we now also have `pushFrontDecl`, I can change it back, right now it looks like this in the code:

```c++
    void pushBackExpr(Expr* expr);
    void pushFrontDecl(VarDecl* decl);
```

query/AST/Projection.h
LucLabarriere: Only because we might not find a name for the item, I could return std::optional<std::string_view> instead

query/AST/Projection.h
rjb32: I prefer the addXYZ naming

query/AST/Projection.h
rjb32: I think in this case it will look better

query/AST/Projection.h
```diff
@@ -1,21 +1,29 @@
...
 class CypherAST;
 
 class Projection {
 public:
     friend CypherAST;
     struct All {};
-    using Items = std::vector<Expr*>;
+
+    using ReturnItem = std::variant<Expr*, VarDecl*>;
+    using Items = std::list<ReturnItem>;
```
cyrusknopf, line 26: Why a `std::list` now?

query/AST/Projection.h
```diff
@@ -66,8 +82,11 @@ class Projection {
     bool _distinct {false};
     bool _aggregate {false};
     bool _hasGroupingKeys {false};
+    bool _returningAll {false};
 
-    std::variant<Items, All> _items;
+    Items _items;
+    std::unordered_map<uintptr_t, std::string_view> _names;
```
rjb32, line 90: Can we document what these sets are for?

query/parser/CypherParser.y
```diff
@@ -481,7 +481,7 @@ orderByItem
 
 orderBySSt
     : ORDER BY orderByItem { $$ = OrderBy::create(ast); $$->addItem($3); LOC($$, @$); }
-    | orderBySSt COMMA orderByItem { $$ = $1; $$->addItem($3); }
+    | orderBySSt COMMA orderByItem { $$ = $1; $$->addItem($3); LOC($$, @$); }
```
rjb32, line 484: Good catches

query/AST/Projection.h
LucLabarriere: Because I append items at the front to handle RETURN *, ... now

cyrusknopf: Did you consider `std::deque`? It is minutiae though, I doubt it matters, just wondering

LucLabarriere: Not at all ahah, could have worked probably, overall I think I should use std::deque more often... Here though I think std::list is better because std::deque is a much bigger object to allocate compared to a list

PR #339, opened 2026-01-27
CALL procedures support for arguments + dependencies

LucLabarriere:
- Added a `db.describeCommit(commit)` procedure to test
- Works with

```cypher
CALL db.history() YIELD commit AS c
CALL db.describeCommit(c) YIELD nodeCount
RETURN n, nodeCount
```

```cypher
CHANGE NEW
checkout change-0
CREATE (n:Commit { commitString: "aef14d6ea" })
COMMIT;
MATCH (n)
CALL db.describeCommit(n.commitString) YIELD nodeCount
RETURN n, n.commitString, nodeCount
```

- Fixes this query (from issue #333):

```cypher
CALL db.labels() YIELD id MATCH (n) WHERE n.age = id RETURN n
```

Except that it needs @cyrusknopf operators to work

- Requires much more unit tests

- Right now, the algorithm used to place the procedure call is, for each dependency, evaluate the path between the procedure evaluation and the dependency and:

```c++
const auto [path, ancestorNode] = _topology->getShortestPath(n, dep._var);

switch (path) {
    case PlanGraphTopology::PathToDependency::SameVar:
        return;
    }

    case PlanGraphTopology::PathToDependency::UndirectedPath:
        PlanGraphNode* depBranchTip = _topology->getBranchTip(dep._var);
        depBranchTip->connectOut(n);
        return;
    }
}
```

So for `MATCH (n), (m) CALL proc(n, m) YIELD ... RETURN ...`, we get:

```
Scan -> n --|
            |
            |--> ProcedureEval
            |
Scan -> m --|
```

Not only does that work for simple cases like the `describeCommit` procedure, but could also used for the shortest path calculation @suhassn-turing-bio. However, some procedures will require a cartesian product/join which will make the algorithm fallback to the same one we use for predicates (`insertDataFlowNode` function)

```c++
switch (path) {
    case PlanGraphTopology::PathToDependency::SameVar:
        return;
    }

    case PlanGraphTopology::PathToDependency::UndirectedPath: {
        const auto* varDecl = static_cast<VarNode*>(ancestorNode)->getVarDecl();
        JoinNode* join = _tree->create<JoinNode>(varDecl,
                                                 varDecl,
                                                 JoinType::COMMON_ANCESTOR);
        PlanGraphNode* depBranchTip = _topology->getBranchTip(dependency);
        depBranchTip->connectOut(join);
        return;
    }

        // If nodes are on two different islands
        // Cartesian product. This can be optimized in the future into a ValueHashJoin
        CartesianProductNode* join = _tree->insertBefore<CartesianProductNode>(filter);
        PlanGraphNode* depBranchTip = _topology->getBranchTip(dependency);
        depBranchTip->connectOut(join);
        return;
    }
}
```

If we want to do that someday, we will need to collapse the Join/CartesianProd + Procedure call pairs into a single Procedure call node for procedures that don't work on cartesian products (like the shortest path if we decide to port it).

Example of what it could look it for `shortestPath`:

```cypher
MATCH (a), (b) CALL db.shortestPath(a, b) YIELD path, distance RETURN a, b, path, distance
```

### DIFF
--- a/test/query-test-suite/tests/call-dependencies-0.json
+++ b/test/query-test-suite/tests/call-dependencies-0.json
@@ -1,0 +1,16 @@
+{
+  "enabled": true,
+  "graph": "simpledb",
+  "query": "MATCH (n)\nCALL db.describeCommit(n.name) YIELD nodeCount, edgeCount\nRETURN n, nodeCount, edgeCount",
+  "expect": {
+    "plan": "flowchart TD\n    0[\"`\n        __SCAN_NODES__\n    `\"]\n    1[\"`\n        __VAR__\n        __name__: n\n    `\"]\n    2[\"`\n        __FILTER_NODE__\n    `\"]\n    3[\"`\n        __PROCEDURE_EVAL__\n        __func__ db.describeCommit\n        __yield__\n        __yield_item__: nodeCount\n        __yield_item__: edgeCount\n        __args__\n        __arg__: unnamed\n    `\"]\n    4[\"`\n        __GET_PROPERTY_WITH_NULL__\n        __var__ n\n        __prop__ name\n    `\"]\n    5[\"`\n        __PRODUCE_RESULTS__\n    `\"]\n    0-->4\n    1-->3\n    2-->1\n    3-->5\n    4-->2",
+    "result": "n,nodeCount,edgeCount\n0,0,0\n1,0,0\n2,0,0\n3,0,0\n4,0,0\n5,0,0\n6,0,0\n7,0,0\n8,0,0\n9,0,0\n10,0,0\n11,0,0\n12,0,0\n13,0,0\n14,0,0\n15,0,0\n16,0,0\n17,0,0"
+  },
+  "tags": [
+    "call",
+    "dependencies",
+    "args",
+    "match"
+  ],
+  "write-required": false
+}
--- a/test/query-test-suite/tests/call-dependencies-1.json
+++ b/test/query-test-suite/tests/call-dependencies-1.json
@@ -1,0 +1,17 @@
+{
+  "enabled": false,
+  "graph": "simpledb",
+  "query": "CALL db.labels() YIELD label, id\nMATCH (n) WHERE n.name = label\nRETURN n, label",
+  "expect": {
+    "plan": "flowchart TD\n    0[\"`\n        __PROCEDURE_EVAL__\n        __func__ db.labels\n        __yield__\n        __yield_item__: label\n        __yield_item__: id\n    `\"]\n    1[\"`\n        __SCAN_NODES__\n    `\"]\n    2[\"`\n        __VAR__\n        __name__: n\n    `\"]\n    3[\"`\n        __FILTER_NODE__\n        __has predicate__\n        __dep__ _n_._name_ (String)\n    `\"]\n    4[\"`\n        __GET_PROPERTY_WITH_NULL__\n        __var__ n\n        __prop__ name\n    `\"]\n    5[\"`\n        __CARTESIAN_PRODUCT__\n    `\"]\n    6[\"`\n        __PRODUCE_RESULTS__\n    `\"]\n    0-->5\n    1-->4\n    2-->6\n    3-->2\n    4-->5\n    5-->3",
+    "result": "EXEC ERROR\nUnsupported mask application for kind 273"
+  },
+  "tags": [
+    "call",
+    "dependencies",
+    "where",
+    "match"
+  ],
+  "write-required": false,
+  "disabled-reason": "Unsupported mask application for kind 273"
+}
--- a/test/query-test-suite/tests/call-dependencies-2.json
+++ b/test/query-test-suite/tests/call-dependencies-2.json
@@ -1,0 +1,16 @@
+{
+  "enabled": true,
+  "graph": "simpledb",
+  "query": "CALL db.history() YIELD nodeCount, edgeCount\nMATCH (n) WHERE n.age > nodeCount OR n.age > edgeCount\nRETURN n.name, n.age",
+  "expect": {
+    "plan": "flowchart TD\n    0[\"`\n        __PROCEDURE_EVAL__\n        __func__ db.history\n        __yield__\n        __yield_item__: nodeCount\n        __yield_item__: edgeCount\n    `\"]\n    1[\"`\n        __SCAN_NODES__\n    `\"]\n    2[\"`\n        __VAR__\n        __name__: n\n    `\"]\n    3[\"`\n        __FILTER_NODE__\n        __has predicate__\n        __dep__ _n_._age_ (Integer)\n        __dep__ _n_._age_ (Integer)\n    `\"]\n    4[\"`\n        __GET_PROPERTY_WITH_NULL__\n        __var__ n\n        __prop__ age\n    `\"]\n    5[\"`\n        __CARTESIAN_PRODUCT__\n    `\"]\n    6[\"`\n        __GET_PROPERTY_WITH_NULL__\n        __var__ n\n        __prop__ name\n    `\"]\n    7[\"`\n        __PRODUCE_RESULTS__\n    `\"]\n    0-->5\n    1-->4\n    2-->6\n    3-->2\n    4-->5\n    5-->3\n    6-->7",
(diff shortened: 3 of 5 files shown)


query/plan/PlanGraphVariables.h
```diff
@@ -34,17 +35,24 @@ class PlanGraphVariables {
         return _varNodes;
     };
 
+    void setProducer(const VarDecl* varDecl, PlanGraphNode* producer);
```
rjb32, line 38: I think that is very interesting, especially also for my vector search command that may also have a YIELD clause

rjb32:
@LucLabarriere Please add the CALL describe commit + MATCH query as a unit test

query/plan/PipelineGenerator.cpp
```diff
@@ -946,7 +947,43 @@ PipelineOutputInterface* PipelineGenerator::translateProcedureEvalNode(Procedure
...
+    size_t i = 0;
+    for (const auto* argExpr : *argExprs) {
+        const Column* col = nullptr;
+
+        const VarDecl* argDecl = argExpr->getExprVarDecl();
+
+        if (!argDecl) {
+            ExprProgram* exprProg = ExprProgram::create(_pipeline);
+            ExprProgramGenerator exprGen(this, exprProg, _builder.getPendingOutput());
+            col = exprGen.generateExpr(argExpr);
```
rjb32, line 976: Here you get the resulting column from ExprProgramGenerator but nowhere there is a call to the eval function, not even in the DatabaseProcedureProcessor? So who actually evaluates the expressions?

query/plan/ExprProgramGenerator.cpp
```diff
@@ -237,10 +237,11 @@ Column* ExprProgramGenerator::generateSymbolExpr(const SymbolExpr* symbolExpr) {
...
 
-    if (type != EvaluatedType::NodePattern && type != EvaluatedType::EdgePattern) {
-        throw PlannerException(
-            "Attempted to generate SymbolExpr which was neither Node nor EdgePattern.");
-    }
+    // TODO check, this probably breaks stuff
+    //if (type != EvaluatedType::NodePattern && type != EvaluatedType::EdgePattern) {
+    //    throw PlannerException(
+    //        "Attempted to generate SymbolExpr which was neither Node nor EdgePattern.");
+    //}
```
LucLabarriere, line 246: @cyrusknopf I commented out this check since now, symbols can represent other things than just entities (result of yield: `YIELD nodeCount`, or later with statements: `WITH 5 AS intLiteral`). It might break something in the future though so we'll have to be careful.

cyrusknopf: Hmm okay

query/plan/ReadStmtGenerator.cpp
```diff
@@ -537,6 +500,39 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+                deps.genExprDependencies(*_variables, arg);
+
+                for (ExprDependencies::VarDependency& dep : deps.getVarDeps()) {
+                    generateDependency(dep._producerNode, dep._expr);
+                    const auto [path, ancestorNode] = _topology->getShortestPath(n, dep._producerNode);
+
+                    switch (path) {
+                        case PlanGraphTopology::PathToDependency::SameVar:
+                        case PlanGraphTopology::PathToDependency::BackwardPath: {
+                            return;
```
ghost, line 523: Maybe we should throw an error here? If we find a Procedure node to be the same var or on the same branch as any other var there is definitely something broken right?

query/plan/ReadStmtGenerator.cpp
```diff
@@ -537,6 +500,39 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+                        }
+
+                        case PlanGraphTopology::PathToDependency::UndirectedPath:
```
ghost: I wonder what the right thing to do in this case is. Say we have
`match (n)-->(m), (n)-->(p), call db.procedure(m,p)`

with the current algorithm we connect the procedure node to m, and then for the dependency on var node p - find that there is an undirected path between the procedure node and the p var node and then connect the nodes directly. But I think the intended behaviour here should be a Common Ancestor Join and then connecting that to the procedure node.

ghost: I've run into this issue with my shortest path processor implementation and it's pretty hard to have a good solution there, but I think here there can be a nice solution!

query/plan/ReadStmtGenerator.cpp
LucLabarriere: That's true yeah! I4ll do that

query/plan/ReadStmtGenerator.cpp
LucLabarriere: I'm not sure actually, I was thinking that in this case we would have wanted to receive the columns with possibly different dimensions. Isn't it what you need for the shortestPath?

ghost: Hmmm how I saw it was if someone had written a join pattern and then want some variables from the pattern as an input to a procedure they would expect there to be some filtering through a join.

LucLabarriere: But that wouldn't work for the shortest path right ? Because I agree with you but right now we don't have such case, and if we want to do that, I'll think we'll also have to add CartesianProduct since the user could also expect to work on a cartesian product as input (it's quite dependent on the procedure that is called). If you confirm that I'm right that the shortest path expects the inputs to not be joined/cartesian product, then here's what I suggest:

- Right now keep it as is (just connect the tip of the branches). This means that the procedures need to expect the inputs to not have the same dimensions

- Later, fix the switch to add a Join/CartProd depending on the case + add a step in the Optimization phase that finds all Join/CartProd + ProcedureEval pairs, and check whether the procedure needs that to be optimized away or not

My understanding is that a procedure like a shortest path: `CALL db.shortestPath(n, m)` could be implemented as a simple Join/CartProd + shortestPath on all generated pairs (what neo4j does), but your implementation is an optimized version of it that skips the need for a cartesian product, so it would make sense to have the optimizer to remove those (I was thinking of doing that later)

rjb32: I agree that we don't want to do cartesian product in general right now for the inputs of a procedure because we don't care about having input columns to a procedure with the same dimension.

However I think that in the case of a common ancestor:
```
n------->m.      CALL proc(m,p)
 |-----> p
``` 
We expect to receive m,p "filtered" implicitly by a join on the common ancestor n so that m and p derive from the same n values. Otherwise the set of inputs to the procedure can just be wrong. In this case I think it's not about the dimension matter, it's about the implicit filtering made by the join on n that needs to be done.

LucLabarriere: Oh I get it now ! it took me quite some time to understand sorry

query/plan/ReadStmtGenerator.cpp
LucLabarriere: I added an error on SameVar, but I kept the BackwardPath to be ok, since I think it would this would trigger it:

```cypher
MATCH (n)-->(m) CALL sth(n, m) ...
```

Here `n` and `m` are two dependencies, the first dependency evaluation connects the procedure evaluation to the tip of the branch (`n->m`), then we evaluate the `m` dependency and see that we have a backward path that does not require to do anything

query/plan/ReadStmtGenerator.cpp
```diff
@@ -537,6 +500,52 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+
+                        case PlanGraphTopology::PathToDependency::UndirectedPath: {
+                            // Join
+                            const auto* varDecl = static_cast<VarNode*>(ancestorNode)->getVarDecl();
+                            JoinNode* join = _tree->create<JoinNode>(varDecl,
+                                                                     varDecl,
+                                                                     JoinType::COMMON_ANCESTOR);
+                            PlanGraphNode* depBranchTip = _topology->getBranchTip(dep._producerNode);
+                            depBranchTip->connectOut(join);
+                            return;
```
LucLabarriere, line 534: I think this should fix the issue you had @suhassn-turing-bio, do you agree?

Sorry it took me so long to understand!

PR #344, opened 2026-01-27
Unit test suite with a Web app that can list/run/create unit tests

LucLabarriere:
# Unit test suite

image

- Full blown unit test suite that allows to create unit tests easily

image

- Can run the unit tests and compare the generated plan graph + query output to the expected ones

- Shows a description of the test, its output and plan graph as an interactive mermaid graph

image

- Can update the query test output/plangraph by clicking on a button

image

- Can share unit tests links
- Can disable certain unit tests if some features are not supported

- Supports write transactions (creates changes dynamically if required)

### DIFF
--- a/test-suite/frontend/src/App.tsx
+++ b/test-suite/frontend/src/App.tsx
@@ -26,7 +26,7 @@ import {
   SidebarTrigger,
   useSidebar
 } from "@/components/ui/sidebar";
-import { FilePlus, GitCompareArrows, PanelLeft, Play, PlayCircle, Plus, Share2, Trash2 } from "lucide-react";
+import { AlertTriangle, Ban, CheckCircle2, Clock3, FilePlus, GitCompareArrows, PanelLeft, Play, PlayCircle, Plus, Share2, Trash2 } from "lucide-react";
 
 type TestMeta = {
   name: string;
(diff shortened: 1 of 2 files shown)


test/query-test-suite/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+set(QUERY_TEST_SUITE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/tests)
+
+target_compile_definitions(test_query_test_suite
+    PRIVATE QUERY_TEST_SUITE_DIR="${QUERY_TEST_SUITE_DIR}")
+
+target_compile_definitions(query_test_suite_cli
+    PRIVATE QUERY_TEST_SUITE_DIR="${QUERY_TEST_SUITE_DIR}")
+
+install(RUNTIME_DEPENDENCY_SET turingdb-test-cli
+    RUNTIME DESTINATION bin)
```
LucLabarriere: @rjb32 Most critical part of the code I think, it adds a turingdb-test-cli binary to the $TURING_HOME/bin folder that can be used by the Rest API. Maybe this is not ideal

rjb32: Can we not put it in the install?

LucLabarriere: That's what it does here, but it means that clients may have access to it in their path which is weird (especially since they don't have the sources so they don't have the unit test files)

rjb32: Ok then it's better to just have it as a binary in the test suite directory but not Install it

PR #346, opened 2026-01-27
Fix codereview CI for external PRs

rjb32:
Fix the codereview CI to make it work for external PRs by contributors who do not have write access on the repository.

* Run the codereview tool on the PR to get the errors and warnings
* Publish the comments on main branch which is trusted code


rjb32:
## Investigation Complete

**Finding:** The Code Style Review workflow was not triggering because this PR modifies the workflow file itself (`.github/workflows/code_review.yml`). This is expected GitHub security behavior - when workflow files are in the PR diff, path filter behavior changes.

**Verification:** Created test PR #347 with only a `.cpp` file (no workflow changes) and the Code Style Review workflow triggered and ran successfully.

**Resolution:** Removed the test `.cpp` file from this PR. This PR now only contains the workflow fix. Since there are no `.cpp`/`.h`/`.hpp` files in this PR, the Code Style Review workflow is expected to not trigger (and that's correct behavior - there's nothing to review).

The workflow fix itself will be testable once this is merged to main, when a real PR with `.cpp` files is opened.

PR #347, opened 2026-01-27
Test: Verify Code Style Review trigger

rjb32:
## Summary
Testing if the Code Style Review workflow triggers when only `.cpp` files are changed (without workflow file modifications).

This is a test to verify that PR 346's Code Style Review not triggering is due to the workflow file being modified in the same PR.

## Test Plan
- Observe if Code Style Review workflow triggers for this PR
- If it does: confirms the theory that workflow file modifications block the trigger
- If it doesn't: need further investigation

This PR will be closed after observation.

### DIFF
--- a/storage/test_style_check.cpp
+++ b/storage/test_style_check.cpp
@@ -1,0 +1,23 @@
+// Test file with intentional style violations
+// This file is used to verify that the Code Style Review workflow triggers
+
+#include <iostream>
+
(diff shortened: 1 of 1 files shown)


rjb32:
Test confirmed: Code Style Review workflow triggers when only .cpp files are changed. The issue with PR 346 is that modifying the workflow file in the same PR blocks the trigger (expected GitHub security behavior).

PR #349, opened 2026-01-28
Feat/dnf package manager support

roquess:
This PR adds support for dnf package manager to enable building on Red Hat, Fedora, and CentOS systems.

## Changes
- Auto-detect package manager (apt-get or dnf) at script runtime
- Add dnf installation commands as alternative to apt-get
- Maintain full backward compatibility with Debian/Ubuntu systems

## Testing
Tested on Fedora 43

Resolves the issue where the build script only supported Debian-based distributions.

### DIFF
--- a/dependencies.sh
+++ b/dependencies.sh
@@ -15,20 +15,20 @@ BUILD_DIR=$DEPENDENCIES_DIR/build
 mkdir -p $DEPENDENCIES_DIR
 mkdir -p $BUILD_DIR
 
-# Detect package manager (apt-get or dnf)
-if command -v apt-get &> /dev/null; then
-    PKG_MANAGER="apt-get"
-    PKG_INSTALL="install -qqy"
-elif command -v dnf &> /dev/null; then
-    PKG_MANAGER="dnf"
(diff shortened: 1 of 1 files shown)


dependencies.sh
```diff
@@ -10,17 +9,28 @@ else
...
 if command -v apt-get &> /dev/null; then
-    echo "Updating apt cache..."
-    sudo apt-get update
+    PKG_MANAGER="apt-get"
+    PKG_INSTALL="install -qqy"
+elif command -v dnf &> /dev/null; then
+    PKG_MANAGER="dnf"
+    PKG_INSTALL="install -y"
+else
+    echo "Neither apt-get nor dnf found. Please install dependencies manually."
```
rjb32: @roquess See macos CI, we enter in this else case when on macos because neither apt-get nor dnf exists, so the condition has to be a little bit different

roquess: You're right, the package manager detection should only be done on Linux. On macOS, neither apt-get nor dnf exist, so the script falls into the else case and exits with an error when it should continue normally.

PR #351, opened 2026-01-28
Throw error for unsupported operators in return, disable test for now

### LINKED ISSUE #343: Returning arbitrary expressions raises an exception
If we attempt to return an arbitrary expression, we get an exception:

```cypher
MATCH (n) RETURN n.duration + 5
```

Output:

```
PLAN ERROR
Unexpected exception: _Map_base::at
```

This should be fixed in one of @cyrusknopf's PRs

cyrusknopf:
Fixes #343


test/query/plan/success-reads/aggregate-3
rjb32, line 1: Why removing it instead of marking it as an expected fail?

cyrusknopf: I'm not sure if there is a way to mark an individual test case as expected to fail with this system. @LucLabarriere do you know of a way?

LucLabarriere: You can maybe update the unit test but it's quite tedious here, it will be replaced by the test-suite web app PR + it's fixed in your future PR on expressions so I'm not sure it's worth spending time on it here. This PR just changes the error message of a unit test that will not even fail anymore in a few days

rjb32: Ok then let's merge this

rjb32 (approved):

PR #354, opened 2026-01-30
Build TuringDB using UV and bring Python SDK in repo

rjb32:
* Build turingDB using the UV package manager
* Bring the Python SDK finally in the TuringDB repo

This has and will have several advantages:
* Don't have to maintain coordination with the turingdb-python repo and submodule update for updating turingdb
* A more meaningful regress: instead of downloading the turingdb package from pypi in each regress test case, we can test directly against the package with the current turingdb being built
* Prepare for in-process turingdb and a python module written in C++ and thus we will need tight synchronisation with the turingdb C++ main codebase

### DIFF
--- a/.github/workflows/ci_docker.yaml
+++ b/.github/workflows/ci_docker.yaml
@@ -2,7 +2,7 @@ name: Docker Nightly / Release Build
 
 on:
   schedule:
-    - cron: "0 2 * * *"   # nightly at 02:00 UTC
+    - cron: "0 4 * * *"   # nightly at 04:00 UTC
   workflow_dispatch:
   push:
     tags:


setup.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def _get_project_root() -> Path:
+    """Get the project root directory."""
+    return Path(__file__).parent.resolve()
+
+
+def _get_build_executable() -> Path:
+    """Get the path to the executable in the build directory."""
+    return _get_project_root() / "build" / "turing_install" / "bin" / "turingdb"
```
LucLabarriere, line 21: I'm not sure we all use the `build/turing_install` path, can we use `TURING_HOME` instead ? By default in the setup.sh script it is set to this path, and it can be overriden.

Or maybe it does not matter since it's only for the uv build

setup.py
```diff
@@ -0,0 +1,102 @@
...
+    python_requires=">=3.10",
+    install_requires=[
+        "boto3>=1.42.22",
+        "boto3-stubs>=1.42.22",
+        "click>=8.3.1",
+        "httpx>=0.28.1",
+        "orjson>=3.11.5",
+        "pandas>=2.3.3",
+        "pandas-stubs>=2.3.3.251219",
+        "prompt-toolkit>=3.0.52",
```
LucLabarriere: I'm a bit surprised by this, isn't it automatically handled by uv?

rjb32: Yes fixed

setup.py
rjb32: Yes fixed

setup.py
```diff
@@ -0,0 +1,96 @@
...
+    if turing_home:
+        return Path(turing_home) / "bin" / "turingdb"
+    return _get_project_root() / "build" / "turing_install" / "bin" / "turingdb"
+
+
+def _run_cmake_build():
+    """Run cmake configure, build, and install."""
+    source_dir = _get_project_root()
+    build_dir = source_dir / "build"
+    build_dir.mkdir(parents=True, exist_ok=True)
```
cyrusknopf, line 28: Would we rather remove any build directory first? Sometimes artefacts left in a stale build directory caused CI problems

LucLabarriere: I'm not sure personally because Maxime will use this probably, and he'll have to rebuild the whole thing everytime if that's the case

rjb32: I think it's best in doubt to keep the removal of the build directory manual for now, just like with cmake

rjb32:
I found a way to expand the compatibility of the macos wheel with a minimal MacOS version that we can supposedly give via an environment variable.

PR #364, opened 2026-01-31
Add Suse/OpenSuse support to dependencies.sh

roquess:
## Summary

Adds `zypper` support to the dependency build script. Package names are aligned to work on both openSUSE Leap and Tumbleweed without any distro-specific detection.

## Changes

- `zypper` is now detected alongside `apt-get` and `dnf`, with `--non-interactive` set by default to avoid GPG key prompts.
- All package names use the metapackage variants (`libopenssl-devel`, `libcurl-devel`, etc.) that resolve correctly on both Leap and Tumbleweed.
- `cmake` and `gcc-c++` are installed automatically if missing.
- Added a fallback for `nproc` on systems where `util-linux` is not installed.
- Fixed a missing "already installed" message for `flex` on macOS.

## Tested on

- openSUSE Tumbleweed


rjb32:
Closed because dependencies.sh changed a lot since this PR

PR #369, opened 2026-02-02
Fix Unordered Iteration Of Source Processors

ghost:
There was a bug induced by our iteration over an unordered set we were using to store all the source processors for our pipeline. Fixing that issue uncovered a bug in the hash join processor that seemingly passed before due to the asymmetry in the code paths of the left and right streams. I am going to see if we can simplify the code a bit.

### DIFF
--- a/query/pipeline/PipelineV2.h
+++ b/query/pipeline/PipelineV2.h
@@ -21,7 +21,7 @@ public:
     friend ExprProgram;
     friend PredicateProgram;
     using Processors = std::vector<Processor*>;
-    using SourcesSet = std::unordered_set<Processor*>;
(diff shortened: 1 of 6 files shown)


query/pipeline/PipelineV2.h
```diff
@@ -21,7 +21,7 @@ class PipelineV2 {
     friend ExprProgram;
     friend PredicateProgram;
     using Processors = std::vector<Processor*>;
-    using SourcesSet = std::unordered_set<Processor*>;
+    using SourcesSet = std::vector<Processor*>;
```
rjb32, line 24: Indeed much better

PR #370, opened 2026-02-02
Fix wheel platform tags for cross-platform compatibility

### LINKED ISSUE #366: Wrong binary version with `uv add 'turingdb==1.20.0'`, getting arm64 instead of x86
# Wrong architecture

- Only with version 1.20
- Appears on ubuntu x86

## Reproduce
- run

```
mkdir testproject
cd testproject
uv init
uv add 'turingdb==1.20.0'
uv run turingdb

# Output
Unexpected error: [Errno 8] Exec format error
```

The binary version is arm64 instead of x86:

```
file ./.venv/lib64/python3.13/site-packages/turingdb/bin/turingdb 
./.venv/lib64/python3.13/site-packages/turingdb/bin/turingdb: Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|WEAK_DEFINES|BINDS_TO_WEAK|PIE|HAS_TLV_DESCRIPTORS>
```

## Expected
Should start turingdb, binary should be x86

rjb32:
## Summary

This fixes issue #366 where Ubuntu users were getting macOS ARM64 binaries when installing turingdb v1.20.0.

**Root cause (two combined issues):**

1. macOS wheels were incorrectly tagged as `-none-any` (platform-independent) instead of proper `macosx_*` platform tags. `delocate-wheel` doesn't rename wheels.

2. Linux wheels required glibc 2.39 (`manylinux_2_39`), only available on Ubuntu 24.04+. Users on older Ubuntu couldn't use these wheels, causing pip/uv to fall back to the `-none-any` wheel containing the macOS binary.

**Changes:**

- **macOS workflow**: Add wheel retagging step to rename `-none-any` wheels to proper `macosx_11_0_arm64` platform tags
- **Ubuntu workflow**: Switch to `manylinux_2_28` container for broader glibc compatibility (2.28 vs 2.39), supporting Ubuntu 20.04+
- **Both workflows**: Add verification step to prevent publishing platform-independent wheels

**Expected wheel names after fix:**
```
turingdb-X.Y.Z-cpXXX-cpXXX-macosx_11_0_arm64.whl
turingdb-X.Y.Z-cpXXX-cpXXX-manylinux_2_28_x86_64.whl
```

Fixes #366

### DIFF
--- a/.github/workflows/ci_release_macos.yml
+++ b/.github/workflows/ci_release_macos.yml
@@ -68,7 +68,8 @@ jobs:
         uv build --wheel --python ${{ matrix.python-version }}
       env:
         # Use LLVM's lld linker to avoid Xcode linker loading wrong libc++ at runtime
-        CMAKE_ARGS: "-DCMAKE_C_COMPILER=${{ env.LLVM_PREFIX }}/bin/clang -DCMAKE_CXX_COMPILER=${{ env.LLVM_PREFIX }}/bin/clang++ -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=lld -DCMAKE_SHARED_LINKER_FLAGS=-fuse-ld=lld -DCMAKE_MODULE_LINKER_FLAGS=-fuse-ld=lld"
+        # Add -fexperimental-library to enable std::jthread support in LLVM 19's libc++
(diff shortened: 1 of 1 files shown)


rjb32:
Closed because I am going to do it some other way, this is way too complex.

PR #375, opened 2026-02-03
Added support for `LOAD JSONL 'mygraph.jsonl' AS mygraph` in Cypher

LucLabarriere:
# Feature

Added the ability to parse `jsonl` files and create graphs from them. The jsonl model is that returned by the Neo4J's apoc procedure: `apoc.export.json.all`

Supports specifying edge sources and targets using the neo4j format (json object), an integer, or a string:

```
{ ... start: { "id": "5, ... }, end: { "id": 10, ... }
{ ... start: "5", end: 10 }
```

### DIFF
--- a/import/jsonl/JsonlImportResult.h
+++ b/import/jsonl/JsonlImportResult.h
@@ -41,10 +41,10 @@ public:
     {
     }
 
-    [[nodiscard]] JsonlImportErrorType getType() const { return _type; }
-    [[nodiscard]] size_t getLine() const { return _line; }
-    [[nodiscard]] std::string_view getMessage() const { return _message; }
-    [[nodiscard]] std::string fmtMessage() const;
+    JsonlImportErrorType getType() const { return _type; }
+    size_t getLine() const { return _line; }
+    std::string_view getMessage() const { return _message; }
+    std::string fmtMessage() const;
 
     template <typename... T>
     static BadResult<JsonlImportError> result(JsonlImportErrorType type,
--- a/import/jsonl/JsonlParser.cpp
+++ b/import/jsonl/JsonlParser.cpp
@@ -194,7 +194,7 @@ JsonlImportResult<void> JsonlParser::parse(ChangeAccessor& change, std::istream&
                     tgtNodeID = parseEntityID(*end);
                 }
 
-                [[maybe_unused]] const EdgeRecord& edge = builder.addEdge(edgeTypeID, srcNodeID, tgtNodeID);
+                const EdgeRecord& edge = builder.addEdge(edgeTypeID, srcNodeID, tgtNodeID);
 
                 if (properties != obj.end()) {
                     for (const auto& [key, value] : properties->items()) {
--- a/import/jsonl/JsonlParser.h
+++ b/import/jsonl/JsonlParser.h
@@ -19,7 +19,7 @@ public:
     JsonlParser& operator=(const JsonlParser&) = delete;
     JsonlParser& operator=(JsonlParser&&) = delete;
 
-    static JsonlImportResult<void> parse(ChangeAccessor& change, std::istream& stream);
+    [[nodiscard]] static JsonlImportResult<void> parse(ChangeAccessor& change, std::istream& stream);
 };
 
 }


LucLabarriere:
> ## ✅ Code Style Review Results
> Found **0 error(s)** and **1 warning(s)**.
> 
> Click to see all violations
> File	Line	Severity	Rule	Message
> `samples/jsonl-import/main.cpp`	62	warning	Const	Local variable 'db' is never modified. Consider declaring it as const.

@rjb32 The warning comes from here:

```c++
TuringDB db {&config};
SystemManager& sysMan = db.getSystemManager();
```

`db` cannot be const to be able to access `SystemManager` as non const, how do we tag that as a false positive?

rjb32:
ignore errors

rjb32:
> > ## ✅ Code Style Review Results
> > Found **0 error(s)** and **1 warning(s)**.
> > Click to see all violations
> > File	Line	Severity	Rule	Message
> > `samples/jsonl-import/main.cpp`	62	warning	Const	Local variable 'db' is never modified. Consider declaring it as const.
> 
> @rjb32 The warning comes from here:
> 
> ```c++
> TuringDB db {&config};
> SystemManager& sysMan = db.getSystemManager();
> ```
> 
> `db` cannot be const to be able to access `SystemManager` as non const, how do we tag that as a false positive?

It is just a warning it should allow you to proceed

import/jsonl/JsonlImportResult.h
```diff
@@ -0,0 +1,65 @@
...
+    explicit JsonlImportError(JsonlImportErrorType type,
+                              size_t line,
+                              const std::string& message = "")
+        : _type(type),
+        _line(line),
+        _message(message)
+    {
+    }
+
+    [[nodiscard]] JsonlImportErrorType getType() const { return _type; }
```
rjb32: All these nodiscard are good in theory but they add syntactical noise everywhere

import/jsonl/JsonlParser.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+                        if (value.is_number_float()) {
+                            vt = ValueType::Double;
+                        } else if (value.is_boolean()) {
+                            vt = ValueType::Bool;
+                        } else if (value.is_number()) {
+                            vt = ValueType::Int64;
+                        } else {
+                            vt = ValueType::String;
+                        }
```
cyrusknopf, line 128: I presume this is from the nlohmann JSON lib, but is it possible that `value` is neither float, nor bool, nor number, nor string? And then this would fail silently, just continuing as if it were a string?

import/jsonl/JsonlParser.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+                        PropertyType pt = metadataBuilder.getOrCreatePropertyType(ptName, vt);
+                        if (pt._valueType != vt) {
+                            ptName += fmt::format(" ({})", ValueTypeName::value(vt));
```
cyrusknopf, line 132: Do we have to have this `propName (propType)` convention again?

import/jsonl/JsonlParser.cpp
```diff
@@ -0,0 +1,243 @@
...
+                        if (value.is_number_float()) {
+                            builder.addNodeProperty<types::Double>(nodeID, pt._id, value.get<double>());
+                        } else if (value.is_boolean()) {
+                            builder.addNodeProperty<types::Bool>(nodeID, pt._id, value.get<bool>());
+                        } else if (value.is_number()) {
+                            builder.addNodeProperty<types::Int64>(nodeID, pt._id, value.get<int64_t>());
+                        } else if (value.is_string()) {
+                            builder.addNodeProperty<types::String>(nodeID, pt._id, value.get<std::string_view>());
+                        } else {
+                            builder.addNodeProperty<types::String>(nodeID, pt._id, value.dump());
```
cyrusknopf, line 145: In what case of `value` would this else branch hit?

import/jsonl/JsonlParser.cpp
```diff
@@ -0,0 +1,243 @@
...
+                if (end == obj.end()) {
+                    return JsonlImportError::result(JsonlImportErrorType::MISSING_EDGE_TYPE, lineNumber);
+                }
+
+                EdgeTypeID edgeTypeID = metadataBuilder.getOrCreateEdgeType(edgeType->get<std::string_view>());
+
+                NodeID srcNodeID;
+                NodeID tgtNodeID;
+
+                if (start->is_object()) {
```
cyrusknopf, line 173: What does this branch mean?

import/jsonl/JsonlParser.cpp
```diff
@@ -0,0 +1,243 @@
...
+                    if (tgtIDIt == end->end()) {
+                        return JsonlImportError::result(JsonlImportErrorType::MISSING_EDGE_TGT_ID, lineNumber);
+                    }
+
+                    tgtNodeID = nodeIDs.at(parseEntityID(*tgtIDIt));
+                } else {
+                    tgtNodeID = parseEntityID(*end);
+                }
+
+                [[maybe_unused]] const EdgeRecord& edge = builder.addEdge(edgeTypeID, srcNodeID, tgtNodeID);
```
cyrusknopf:
```suggestion
                const EdgeRecord& edge = builder.addEdge(edgeTypeID, srcNodeID, tgtNodeID);
```

import/jsonl/JsonlParser.h
```diff
@@ -19,7 +19,7 @@ class JsonlParser {
     JsonlParser& operator=(const JsonlParser&) = delete;
     JsonlParser& operator=(JsonlParser&&) = delete;
 
-    [[nodiscard]] static JsonlImportResult<void> parse(ChangeAccessor& change, std::istream& stream);
+    static JsonlImportResult<void> parse(ChangeAccessor& change, std::istream& stream);
```
cyrusknopf: I would say this should have `[[nodiscard]]` - not checking this result is actually harmful. The other examples in `JsonlImportResult.h` such as
```
    [[nodiscard]] size_t getLine() const { return _line; }
```
should be removed IMO though: not checking their result is not harmful

import/jsonl/JsonlParser.cpp
LucLabarriere: So in general, we should able to accept any kind of property types, including objects and lists (which neo4j supports), so if encounter anything else than the supported types, we import it as a string. The lists and maps are 'dumped' as is, so if we have `"MyProperty": [ 1, 2, 3]`, it gets dumped into turingdb as a property type "MyProperty" of value type string with value: `"[ 1, 2, 3]"`

import/jsonl/JsonlParser.cpp
LucLabarriere: We have to have some convention for when we encounter conflicts with existing property types. Here I reused the same convention but maybe we can use something else, like `propName_valueType` (that's only if there's a conflict, which is quite rare, it only occurs twice in the latest reactome version for example)

import/jsonl/JsonlParser.cpp
LucLabarriere: I would trigger for `null`, lists `[...]`, and objects `{ key: value }`

import/jsonl/JsonlParser.cpp
LucLabarriere: This is a bit ugly but only due to the way neo4j exports graphs with its `CALL apoc.export.json.all()` procedure. The procedure exports the graph as:

```jsonl
{ "id": "0", "type": "node", "properties": {} }
{ "id": "1", "type": "node", "properties": {} }
{ "id": "2", "type": "node", "properties": {} }
{ "id": "0", "type": "edge", "properties": {},
  "start": { "id": "0", "type": "node", "properties": {}},
  "end": { "id": "1", "type": "node", "properties": {}}
}
```

Which is super weird, but it's to be as compatible as possible with neo4j's output

import/jsonl/JsonlParser.cpp
cyrusknopf: Ah sorry I misread this snippet, I thought it would unconditionally add the (type), but I now see it is only on conflict.

import/jsonl/JsonlParser.cpp
cyrusknopf: Ah I see, do we want to issue a warning on list or map since we currently don't support them?

import/jsonl/JsonlParser.h
LucLabarriere: Yeah I guess there are three design choices:
- Put it everywhere where it caches bugs (if you call a const getter but don't use the return value, there is something fishy): this is my approach, but I don't think you guys like it
- Put it only where we want to remind the dev it should use the return value to handle errors: This is probably your favourite approach
- Put it nowhere because it generates noise: Probably Remy's approach

LucLabarriere: Wait you're right I think I misread Remy's comment, I thought he wrote it on `JsonlImportResult<void> parse()` but I see he wrote it for the getters

LucLabarriere: I readded the [[nodiscard]] hint on the parse method, and removed it from the getters

import/jsonl/JsonlParser.cpp
LucLabarriere: I wonder, the problem is that some imports will trigger many warnings that will fill up the logs and slow down considerably the imports...

import/jsonl/JsonlParser.h
cyrusknopf: Yeah there are a couple approaches for sure, maybe we need a debate then put it in the guidelines 😄

import/jsonl/JsonlParser.cpp
cyrusknopf: Ah okay yeah no worries

PR #376, opened 2026-02-03
Add LOAD CSV specification document

rjb32:
This document specifies the design and architecture for CSV loading functionality in TuringDB, modeled after Neo4j's LOAD CSV clause.

LOAD CSV in neo4j https://neo4j.com/docs/cypher-manual/current/clauses/load-csv/

### DIFF
--- a/docs/CSV.md
+++ b/docs/CSV.md
@@ -164,9 +164,9 @@ LOAD CSV 'path/to/file.csv' WITH HEADERS AS row
 CREATE (n:Person {name: row.name, age: toInteger(row.age)})
 ```
 
-### ON ERROR SKIP Mode
+### Error Handling Mode
 
-By default, malformed lines cause a runtime error. Use `ON ERROR SKIP` to skip malformed lines and continue processing:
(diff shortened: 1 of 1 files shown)


rjb32:
Specification for #382

rjb32:
The scope:
* Goal: mimic neo4j LOAD CSV ergonomics 
* Don't use variants or list types if we can do otherwise
* Neo4j LOAD CSV actually returns strings that have to be converted if necessary using explicit conversion functions, for example to integer..etc
* `LOAD CSV 'file.csv' AS row `the row variable can be used to feed a MATCH clause or a CREATE clause in the same query. Thus LOAD CSV is a read statement from the grammar point of view.
* We will support filtering of rows later using WITH or LOAD CSV... WHERE cond
* LOAD CSV errors out by default if a row is malformed or has not the right number of columns. It is possible to make it skip these rows instead using LOAD CSV... ON ERROR SKIP
* We can get plain ColumnString* using an indexed expression like row[0], row[42], but also with a non-literal index like row[id] potentially

Issue #377, opened 2026-02-03
Query Test Suite does not support string-outputs of ColumnConst


cyrusknopf:
Queries which return constant values, such as
~~~
MATCH (n) RETURN 4
~~~
(supported as part of #360), do not format correctly, because
~~~
std::string columnValueToString(const db::Column* column, size_t row) {
    return db::dispatchColumnVector(column, [&](const auto* typed) {
        using ColumnType = std::decay_t<decltype(*typed)>;
        using ValueType = typename ColumnType::ValueType;
        const ValueType& value = (*typed)[row];
        return valueToString(value);
    });
}
~~~
expects a ColumnVector, resulting in output akin to:

<img width="1518" height="737" alt="Image" src="https://github.com/user-attachments/assets/d853bc80-e6b3-47d9-980c-b66542e017b1" />

One option could be to use `ColumnSingleDispatcher` instead of `dispatchColumnVector`, but their may be issues with specifying the `AllowedKinds`, which need require all `ColumnVector` internal types, some of which are not supported by `ColumnConst`.

LucLabarriere:
Which internal kinds are not supported in both ColumnVector<> and ColumnConst<>? I think that philosophically speaking, ColumnConst<> is just an optimisation that allows to avoid filling huge vectors with a single value, so they should allow the same kinds

cyrusknopf:
> Which internal kinds are not supported in both ColumnVector<> and ColumnConst<>? I think that philosophically speaking, ColumnConst<> is just an optimisation that allows to avoid filling huge vectors with a single value, so they should allow the same kinds

Perhaps I misspoke, I meant that there are some types that currently don't make sense to be instantiated as a ColumnConst, for example ColumnConst<NodeID>. But that could change in the future. In any case I got the new dispatcher working

PR #378, opened 2026-02-03
Use new dispatcher for Test Suite

cyrusknopf:
Implements #377

### DIFF
--- a/storage/columns/AllowedKinds.h
+++ b/storage/columns/AllowedKinds.h
@@ -4,10 +4,11 @@
 #include <optional>
 
 #include "ID.h"
-#include "columns/ColumnOperator.h"
+#include "versioning/ChangeID.h"
+#include "ColumnOperator.h"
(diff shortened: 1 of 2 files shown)


storage/columns/AllowedKinds.h
```diff
@@ -238,4 +239,36 @@ struct TypeRestrictions<OP_NOT> {
     >;
 };
 
+/// Types that are outputted by queries, used in @ref QueryTestRunner
+struct OutputtedTypes {
```
rjb32, line 243: Does it have to be put explicitly? Is it not something that we can infer from somewhere else? Like for example from LocalMemory?

cyrusknopf: LocalMemory contains types that do not make sense to format as output, for example `const CommitBuilder*`. Also there is no tuple which I could use of defined internal types in `LocalMemory`, they are defined inline.

rjb32 (approved):

PR #381, opened 2026-02-04
Specifications for the turingdb stop CLI command

LucLabarriere:
View the page at [specs](https://github.com/turing-db/turingdb/blob/docs/stop-cmd-specs/docs/stop-cmd-specs.md)

### DIFF
--- a/docs/stop-cmd-specs.md
+++ b/docs/stop-cmd-specs.md
@@ -1,0 +1,848 @@
+# Stop command specifications
+
+Design specifications for the stop command. The general idea is that we may
+want to launch multiple instances of TuringDB that would run in separate 
+`turing-dir` directories, and listen on different ports. The stop command
+should be bullet proof and allow:
+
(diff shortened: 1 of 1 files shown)


docs/stop-cmd-specs.md
```diff
@@ -0,0 +1,848 @@
...
+## Stop Command Implementation
+
+### Data Structures
+
+```cpp
+struct LockFileData {
+    pid_t pid;
+    int port;
+    time_t start_time;
+    char hostname[256];
```
rjb32, line 177: It is not clear to me why should we write the hostname in the lock file.

rjb32:
It will be important during implementation to take care of not duplicating process management code just for this feature and to enrich/adapt the existing process-management utilities we have in common/

rjb32:
Fairly simple looks good to me (except hostname in the lock file which does not seem very useful?)

PR #391, opened 2026-02-05
Remove AWS SDK and use minio

rjb32:
* Remove AWS SDK
* Use minio SDK because it is way smaller
* Add ls command to s3testcli to have a "tree" like output


.github/workflows/ci_ubuntu.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest-m
+    runs-on: macos-latest
```
LucLabarriere: Before I forget, is it normal that the ubuntu CI runs on macos?

rjb32: Thanks for catching it. I use now a single ci_build.yml workflow file to mutualize the code with a strategy matrix to run a build job per targeted OS

PR #408, opened 2026-02-09
Normalise Dataframe/Column row logic

cyrusknopf:
Context:
- `Dataframe::getRowCount` just returns the size of the first column, but other columns may have differing sizes

- There is no way to check if a Dataframe is "rectangular" in the sense that all columns are logically the same size

- We wish to treat `ColumnConst<T>` as a space-optimised `ColumnVector<T>`, in the sense that has some size and holds the same value (constant) at each index
- We therefore need to match the API of `ColumnVector<T>

Changes:
- Change `getRowCount` to `getLogicalRowCount` which now returns the largest row count of any column

- Add `Dataframe::isRectangular` to check if a Dataframe is logically rectangular

- Add `[]` operator and `at` method to `ColumnConst` - returning the held constant for any `i` - to match the random-access API of `ColumnVector`

### DIFF
--- a/storage/dataframe/Dataframe.cpp
+++ b/storage/dataframe/Dataframe.cpp
@@ -22,21 +22,37 @@ void Dataframe::addColumn(NamedColumn* column) {
 }
 
 size_t Dataframe::getLogicalRowCount() const {
-    size_t mostRows {0};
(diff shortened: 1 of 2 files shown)


cyrusknopf:
Moved to #416

PR #414, opened 2026-02-10
Separate dependencies and build tool installation scripts

cyrusknopf:
Extracted installation of `cmake` and LLVM to a separate script, to be compatible with those not using `brew` or `apt`

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -43,9 +43,12 @@ jobs:
       uses: actions/cache@v4
       with:
         path: external/dependencies
-        key: ${{ matrix.os }}-dependencies-${{ hashFiles('dependencies.sh', '.gitmodules', 'build.sh') }}
+        key: ${{ matrix.os }}-dependencies-${{ hashFiles('install_build_tools.sh', 'dependencies.sh', '.gitmodules', 'build.sh') }}
 
-    - name: Install dependencies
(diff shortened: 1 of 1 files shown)


cyrusknopf:
@rjb32 What are your thoughts on this? I think this is also what @LucLabarriere requested, since he uses Arch linux and so `dependencies.sh` is not runnable because it assumes `apt` on Linux. With this separation, the user can choose to install `cmake` and `llvm` via the build tool install script, or do it themselves, and then `dependencies.sh` is platform-independent.

Would this change require a lot of tweaks to the CI?

dependencies.sh
rjb32, line 78: You will need to read the MACOS_COMPILER_ARGS from the MACOS_SETENV file first to get them back

rjb32:
> @rjb32 What are your thoughts on this? I think this is also what @LucLabarriere requested, since he uses Arch linux and so `dependencies.sh` is not runnable because it assumes `apt` on Linux. With this separation, the user can choose to install `cmake` and `llvm` via the build tool install script, or do it themselves, and then `dependencies.sh` is platform-independent.
> 
> Would this change require a lot of tweaks to the CI?

Yes sounds like a good idea to me

rjb32:
@cyrusknopf I just rebased against main, reloaded the macos setenv file in dependencies.sh and updated the CI workflows

PR #415, opened 2026-02-10
Remove OpenBLAS

rjb32:
Try to remove OpenBLAS.
FAISS: Use Apple Accelerate framework on macos (default choice).

On macos:
```shell
./apple_lld.sh build/tools/turingdb/turingdb

build/tools/turingdb/turingdb
  /System/Library/Frameworks/Accelerate.framework/Versions/A/Accelerate (dyld shared cache)
  /usr/lib/libSystem.B.dylib (dyld shared cache)
  /System/Library/Frameworks/SystemConfiguration.framework/Versions/A/SystemConfiguration (dyld shared cache)
  /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (dyld shared cache)
  /System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (dyld shared cache)
``` 

@cyrusknopf @suhassn-turing-bio @LucLabarriere It actually does not work on ubuntu as we need some BLAS implementation it seems. On macos it's fine there is the Apple Accelerate framework.


LucLabarriere:
I attempted to build it without OpenBLAS on MacOS yesterday but I go unrelated compile errors (from bison...)

I think it should be fine to require OpenBLAS to be installed globally on the machine, because it's such a performance impactful dependency that it's probably better to let the user install it (usually trivial on linux distros anyway). Faiss does `find_package(BLAS REQUIRED)`, so they'll get an error message, but maybe we could add it to our CMakeLists.txt to display a better error?

LucLabarriere:
Update: Managed to make the build pass, I had a pending Xcode Command Line tools update

PR #423, opened 2026-02-13
Test code review bot

rjb32:
Test code review bot


cyrusknopf:
Would we be able to keep the "dropdown" of the violations, so that the big table is hidden by default?

rjb32:
* Violations table is now foldable
* Added section in prompt to actually verify each violation back to the source file and exclude if unsure

PR #427, opened 2026-02-13
Add check for duplicate named return items

### LINKED ISSUE #406: Unique return item names
This query executes,
```cypher
RETURN 5,5
```
and returns
```
+---+
| 5 |
+---+
| 5 |
+---+
```
However, in neo4j, this does not work:

Image

We need to either
1. Support duplicate return item names
2. Add a parse/analysis error for this

cyrusknopf:
Fixes #406, with tests

### DIFF
--- a/query/analyzer/CypherAnalyzer.cpp
+++ b/query/analyzer/CypherAnalyzer.cpp
@@ -257,6 +257,8 @@ void CypherAnalyzer::analyze(const ReturnStmt* returnSt) {
                 continue;
             }
 
+            // If the projection already contains this variable's name, it means it was
+            // added explicitly above. Skip: do not again again as part of the wildcard
             const std::string_view declName = decl->getName();
             if (projection->hasName(declName)) {
                 continue;


query/analyzer/CypherAnalyzer.cpp
```diff
@@ -234,9 +235,16 @@ void CypherAnalyzer::analyze(const ReturnStmt* returnSt) {
...
+        if (seenNames.contains(std::string {name})) {
+            throwError(
+                fmt::format(
+                    "Return items must have unique names; {} was already defined.", name),
+                item);
+        }
 
         _exprAnalyzer->analyzeRootExpr(item);
 
+        // XXX: when is this valid?
```
rjb32: This should not happen as per the parser.
We can have quoted property names such as 
```
n.`something`
```
but normally not for variable names

query/analyzer/CypherAnalyzer.cpp
```diff
@@ -234,9 +235,16 @@ void CypherAnalyzer::analyze(const ReturnStmt* returnSt) {
...
             name = item->getName();
         }
+        if (seenNames.contains(std::string {name})) {
+            throwError(
+                fmt::format(
+                    "Return items must have unique names; {} was already defined.", name),
+                item);
+        }
 
         _exprAnalyzer->analyzeRootExpr(item);
```
LucLabarriere: You actually actually replace this whole snippet:

```c++
        if (auto* symbolExpr = dynamic_cast<SymbolExpr*>(item)) {
            Symbol* symbol = symbolExpr->getSymbol();
            name = symbol->getName();
        } else {
            name = item->getName();
            if (name.empty()) {
                const std::string_view name = srcMan->getStringRepr(std::bit_cast<std::uintptr_t>(item));
                if (name.empty()) [[unlikely]] {
                    throwError("Failed to generate name for projection item", item);
                }

                item->setName(name);
            }

            name = item->getName();
        }

        _exprAnalyzer->analyzeRootExpr(item);

        if (name.empty()) {
            continue;
        }

        if (projection->hasName(name)) {
            continue; // Already added this name (added by wildcard)
        }

        projection->setName(item, name);
```

By this:

```c++
        if (auto* symbolExpr = dynamic_cast<SymbolExpr*>(item)) {
            Symbol* symbol = symbolExpr->getSymbol();
            name = symbol->getName();
        } else {
            name = item->getName();
            if (name.empty()) {
                const std::string_view name = srcMan->getStringRepr(std::bit_cast<std::uintptr_t>(item));
                if (name.empty()) [[unlikely]] {
                    throwError("Failed to generate name for projection item", item);
                }

                item->setName(name);
            }

            name = item->getName();
        }

        _exprAnalyzer->analyzeRootExpr(item);

        bioassert(!name.empty(), "Variable must have a name");

        if (projection->hasName(name)) {
            throwError(
                fmt::format(
                    "Return items must have unique names; {} was already defined.", name),
                item);
        }

        projection->setName(item, name);

```

It uses the `namesSet` member of projection

cyrusknopf: Looks like this works, I thought it would not because of the comment here:
~~~
if (projection->hasName(name)) {
            continue; // Already added this name (added by wildcard)
}
~~~
meaning if we did something like `MATCH (n), (m) RETURN *, n` then `n` would already have a name from the wildcard, and the explicit `n` would trip the exception. Was my understanding wrong?

LucLabarriere: Actually you're right, it doesn't work... my bad, I think we should try to get it to work somehow, instead of having a separate unordered_set though

LucLabarriere: You should be able to swap the logic, first handle the explicit returns, then handle the wildcard?

```c++
    const SourceManager* srcMan = _ast->getSourceManager();

    for (const Projection::ReturnItem& returnItem : projection->items()) {
        const auto* exprPtr = std::get_if<Expr*>(&returnItem);
        if (!exprPtr) {
            continue;
        }

        Expr* item = *exprPtr;
        std::string_view name;

        if (auto* symbolExpr = dynamic_cast<SymbolExpr*>(item)) {
            Symbol* symbol = symbolExpr->getSymbol();
            name = symbol->getName();
        } else {
            name = item->getName();
            if (name.empty()) {
                const std::string_view name = srcMan->getStringRepr(std::bit_cast<std::uintptr_t>(item));
                if (name.empty()) [[unlikely]] {
                    throwError("Failed to generate name for projection item", item);
                }

                item->setName(name);
            }

            name = item->getName();
        }

        _exprAnalyzer->analyzeRootExpr(item);

        bioassert(!name.empty(), "Variable must have a name");

        if (projection->hasName(name)) {
            throwError(
                fmt::format(
                    "Return items must have unique names; {} was already defined.", name),
                item);
        }

        projection->setName(item, name);
        isAggregate |= item->isAggregate();
        hasGroupingKeys |= !item->isAggregate();

        const bool isBinary = item->getKind() == Expr::Kind::BINARY;
        if (isBinary) {
            throw AnalyzeException(
                "Binary expressions in RETURN clauses are not yet supported.");
        }

        const bool isUnary = item->getKind() == Expr::Kind::UNARY;
        if (isUnary) {
            throw AnalyzeException(
                "Unary expressions in RETURN clauses are not yet supported.");
        }
    }

    if (projection->isReturningAll()) {
        // Return all variables defined in the current query

        bioassert(_ctxt, "Query context is invalid");

        // Iterate the decls in reverse declaration order. Since we call `pushFrontDecl()`
        // The decls end up in order. e.g. MATCH (a), (b), (c) RETURN *, a.name
        // - Initial projection items: ['a.name'];
        // - After first `pushFrontDecl()`: ['c', 'a.name'];
        // - After second `pushFrontDecl()`: ['b', 'c', 'a.name'];
        // - After third `pushFrontDecl()`: ['a', 'b', 'c', 'a.name'];
        for (VarDecl* decl : std::views::reverse(*_ctxt)) {
            if (decl->isUnnamed()) {
                continue;
            }

            const std::string_view declName = decl->getName();
            if (projection->hasName(declName)) {
                continue;
            }

            // Push at the front since '*' is only allowed at the beginning of the return statement
            projection->pushFrontDecl(decl);
            projection->setName(decl, declName);
        }
    }

```

LucLabarriere:
@rjb32 
@cyrusknopf I see you have PlanGenTest unit tests failing, I think those should be removed, they were replaced by the unit test suite (all the tests were ported to the test suite)

cyrusknopf:
> @rjb32 @cyrusknopf I see you have PlanGenTest unit tests failing, I think those should be removed, they were replaced by the unit test suite (all the tests were ported to the test suite)

I can remove them in this PR?

query/analyzer/CypherAnalyzer.cpp
cyrusknopf: Yep think that works, pushed it and added a test

rjb32:
> > @rjb32 @cyrusknopf I see you have PlanGenTest unit tests failing, I think those should be removed, they were replaced by the unit test suite (all the tests were ported to the test suite)
> 
> I can remove them in this PR?

Yes please do

query/analyzer/CypherAnalyzer.cpp
```diff
@@ -263,6 +241,33 @@ void CypherAnalyzer::analyze(const ReturnStmt* returnSt) {
...
+        // - After third `pushFrontDecl()`: ['a', 'b', 'c', 'a.name'];
+        for (VarDecl* decl : std::views::reverse(*_ctxt)) {
+            if (decl->isUnnamed()) {
+                continue;
+            }
+
+            const std::string_view declName = decl->getName();
+            if (projection->hasName(declName)) {
+                continue;
+            }
```
cyrusknopf, line 265:
```suggestion

            // If the projection already contains this variable's name, it means it was
            // added explicitly above. Skip: do not again again as part of the wildcard
            const std::string_view declName = decl->getName();
            if (projection->hasName(declName)) {
                continue;
            }
```

PR #442, opened 2026-02-20
Fix initialisation style

rjb32:
Fix many initialisation style issues throughout the codebase.

* Member variables have to start with underscore
* Member variables must be initialised using braces {42}
* Local variables must be initialised using the assignment style


vector/VectorSearchQuery.h
```diff
@@ -47,7 +47,7 @@ class VectorSearchQuery {
     }
 
 private:
-    const Dimension _dimension;
+    const Dimension _dimension {};
```
cyrusknopf: In `VecLibMetadata.h` we do `_dimension {0}`, but here `_dimension {}`. Note sure which one is the correct style but it is inconsistent

storage/versioning/NonDeletedRanges.h
```diff
@@ -6,8 +6,8 @@
 namespace db {
 
 struct NonDeletedRange {
-    size_t start;
-    size_t size;
+    size_t _start {};
```
rjb32: Should be initialised to zero

PR #407, opened 2026-02-09
Implement expression evaluation in RETURN projections

cyrusknopf:
Implements #359

Overview:
- Adds unnamed variable declarations for binary expressions
- Updates `CountProcessor` to use new `ColumnOperatorDispatcher`
- Implements `ExprEvalProcessor` to evaluate expressions and store the results in the output dataframe
- Modifies `ExprProgramGenerator` with more support for unary expressions
- Adds support in `PipelineGenerator` for the new `ExprEvalProcessor`
- Adds evaluation logic for functions with expressions-as-arguments

PR #445, opened 2026-02-20
Correctly encode ColumnConst in JsonEncoder

cyrusknopf:
Context:
- With the new `JsonEncoder`, `ColumnConst<T>` only outputted a single value

Changes:
- Use `Dataframe::getLogicalRowCount` to determine how many values of `ColumnVector<T>` and `ColumnConst<T>` to output

Note: Cannot verify behaviour of returning `ColumnConst<T>` in this PR because it is not possible to return a `ColumnConst<T>` of logical-length greater than 1 until #407 is merged, however there are passing tests in #407 with the same changes in this PR applied

### DIFF
--- a/query/outputs/JsonEncoder.h
+++ b/query/outputs/JsonEncoder.h
@@ -58,6 +58,10 @@ public:
         }
     }
 
+private:
+    WriterT& _writer;
+    const size_t _logicalRowCount {0};
+
     template <Optional T>
     void encodeValue(const T& value) {
         if (!value.has_value()) {
@@ -99,10 +103,6 @@ public:
     void encodeValue(const T& value) {
         _writer.write(fmt::format("\"{}\"", value));
     }
-
-private:
-    WriterT& _writer;
-    const size_t _logicalRowCount {0};
 };
 
 template <Writer WriterT>


query/outputs/JsonEncoder.h
```diff
@@ -14,25 +14,39 @@ namespace db {
...
     void operator()(const ColumnVector<T>* col) {
         _first = true;
 
-        for (const T& value : *col) {
+        for (size_t row = 0; row < _logicalRowCount; row++) {
             if (!_first) {
                 _writer.write(",");
             }
 
+            const T& value = col->at(row);
```
LucLabarriere: I wonder if we can use `->operator[](row)` here to avoid the bounds check. Or we keep `.at(row)` for safety

cyrusknopf: Yeah we can do either, happy to change

cyrusknopf: Actually what are the implications of throwing an exception here with `at()`? Maybe `[]` is preferred

LucLabarriere: Throwing here is caught by the interpreter, so it doesn't crash the server. [] can segfault. However, I think you already added a bounds check for all columns to calculate the row count of the block?

cyrusknopf: Okay good it does not crash the server. There is no bounds check in `getLogicalRowCount`, it only gets the size of the first ColumnVector (i.e. ignores ColumnConsts) and returns that size. So it is theoretically possible a dataframe could contain a ColumnVector of size 10, and another with size 5, and `getLogicalRowCount` returns 10. From what I understand when we discussed this, @rjb32 did not want the check for rectangularity of the dataframe every time a chunk is outputted due to the overhead for each chunk, but correct me if I am wrong.

LucLabarriere: I think it is quite risky to use [] then... but it would be cheaper to do the check once (we're talking nanoseconds no?)

cyrusknopf: I saw slow downs on the order of seconds for a dataframe with 65k columns, but I think that would only really happen when doing something like `CREATE (n1), (n2), (n3), ... { and entire graph worth of nodes} ... (nK) RETURN *` which is questionable to begin with.

LucLabarriere: Hm `->at()` is evaluated per row, so it should happen on regular `MATCH (n) RETURN n` queries too. Maybe branch prediction saves us for this query though?

cyrusknopf: What I mean is if we do a rectangular check, we can then use `[]` confidently, meaning we skip the cost-per-row of `at()` but pay the cost-per-column of the rectangular check. Under the assumption that in general, `number of columns <<< number of rows`, it would be preferred.

LucLabarriere: Oh I see I misunderstood... Then maybe ->at() is fine with the branch prediction. Tough call ahah

rjb32: Yes indeed let's do an operator [] here

cyrusknopf: @rjb32 Could you confirm if you'd want us to do a check that the dataframe is rectangular before the call to JSON encoder? or just use `[]` without a prior check?

rjb32: @cyrusknopf Yes just use [] here. I think that's an imperfect choice to make here but that's probably best right now

cyrusknopf: @rjb32 I have changed to `[]`.
@LucLabarriere I also removed the `_first` attribute from your `ChunkJsonEncoder` functor. All seems to work but let me know if you can see any issues with what I've done.

query/outputs/JsonEncoder.h
```diff
@@ -159,14 +176,21 @@ class JsonEncoder {
     void writeDataframe(const Dataframe& df) {
         arr();
 
-        ChunkJsonEncoder<WriterT> encoder {_writer};
+        const size_t logicalRowCount = df.getLogicalRowCount();
+        ChunkJsonEncoder<WriterT> encoder {._writer = _writer,
```
rjb32: That's a bit not ideal style-wise because it's a public struct/class and we initialise arguments directly, we may forget some..etc which is why constructors have been invented. We should call a constructor here

cyrusknopf: I've made `ChunkJsonEncoder` a class with a defined constructor to always initialise `_writer` and `_logicalRowCount`

PR #457, opened 2026-02-27
Fix Full Matrix CI: Use std::iota instead of std::ranges::iota 

rjb32:
std::ranges::iota is not present on gcc 11/12 used on Ubuntu 22.04.

Full Matrix build CI was failing https://github.com/turing-db/turingdb/actions/runs/22503755762


rjb32:
Fixed in https://github.com/turing-db/turingdb/actions/runs/22505630563

PR #460, opened 2026-03-02
Fix parsing of integer literals

cyrusknopf:
Context:
- The current logic for parsing hexadecimal and octal literals is incorrect, and parses them as decimals without stripping the prefix
- We currently have no guarantee that a parsed int literal fits into an int64

Changes:
- Fixes hex and oct parsing
- Adds a size check on all integer literals, ensuring they fit within a 64 bit integer
- Adds tests to the test suite to verify behaviour

### DIFF
--- a/query/parser/CypherLexer.l
+++ b/query/parser/CypherLexer.l
@@ -225,9 +225,8 @@ using Token = db::YCypherParser::token;
     const char* begin = yytext;
     const char* end = yytext + yyleng;
 
-    bool negative = false;
-    if (*begin == '-') {
-        negative = true;
+    const bool negative = *begin == '-';
+    if (negative) {
         begin += 3; // skip leading "-0x"
     } else {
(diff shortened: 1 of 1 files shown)


query/parser/CypherLexer.l
```diff
@@ -208,27 +208,108 @@ using Token = db::YCypherParser::token;
...
 -?0[xX][0-9a-fA-F]+ {
+    static constexpr int HEX_BASE = 16;
+
+    const char* begin = yytext;
+    const char* end = yytext + yyleng;
+
+    bool negative = false;
+    if (*begin == '-') {
+        negative = true;
+        begin += 3; // skip leading "-0x"
```
rjb32, line 230: Are we sure of this +3? What if the number is "-0"?

cyrusknopf: `-0` does not match the regex pattern for this rule

PR #467, opened 2026-03-02
Implement extensions

rjb32:
Implement an extension system for procedures.

Each extension is a dynamic library in .so on linux which defines its own namespace and fills it with procedure definitions. 

A procedure "hello" defined in an extension "greeter" can be called using a CALL statement:
```shell
CALL greeter.hello()
+---------------+
| message        |
+---------------+
| Hello, World!  |
+---------------+
``` 

Extensions can be listed with SHOW EXTENSIONS:
```shell
SHOW EXTENSIONS
+---------+
| name    |
+---------+
| greeter |
+--------
``` 

An extension has to be explicitly installed before use with the INSTALL statement:
```
INSTALL greeter
``` 
The extension is first searched in the extension directory of the turing installation, which is calculated by default relatively to the path of the turingdb binary in bin_path/../lib/turingdb/extensions and then in the user extensions directory in the turing dir in extensions/ such as $HOME/.turing/extensions. Greeter is a trivial builtin extension which is always available to test in turing install.

Once installed, the procedures of installed extensions are shown by SHOW PROCEDURES.

### DIFF
--- a/extensions/ExtensionManager.cpp
+++ b/extensions/ExtensionManager.cpp
@@ -1,7 +1,6 @@
 #include "ExtensionManager.h"
 
 #include <dlfcn.h>
-#include <shared_mutex>
 
 #include <spdlog/spdlog.h>
 #include <spdlog/fmt/fmt.h>
--- a/extensions/ExtensionManager.h
+++ b/extensions/ExtensionManager.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <mutex>
 #include <shared_mutex>
 #include <string_view>
 #include <unordered_map>
--- a/query/pipeline/procedures/ProcedureManager.cpp
+++ b/query/pipeline/procedures/ProcedureManager.cpp
@@ -37,7 +37,14 @@ void ProcedureManager::init() {
     ProceduresProcedure::registerProcedure(db);
 }
 
+void ProcedureManager::getNamespaces(Namespaces& result) const {
+    std::shared_lock<std::shared_mutex> lock(_mutex);
+    result = _namespaces;
+}
+
 const Procedure* ProcedureManager::getProcedure(std::string_view fullName) const {
+    std::shared_lock<std::shared_mutex> lock(_mutex);
+
     const size_t dot = fullName.find('.');
     if (dot == std::string_view::npos) {
         return nullptr;
@@ -46,15 +53,17 @@ const Procedure* ProcedureManager::getProcedure(std::string_view fullName) const
     const std::string_view nsName = fullName.substr(0, dot);
     const std::string_view procName = fullName.substr(dot + 1);
 
-    const ProcedureNamespace* ns = getNamespace(nsName);
-    if (!ns) {
+    const auto it = _namespaceMap.find(nsName);
+    if (it == _namespaceMap.end()) {
         return nullptr;
     }
 
-    return ns->getProcedure(procName);
+    return it->second->getProcedure(procName);
 }
 
 ProcedureNamespace* ProcedureManager::getNamespace(std::string_view name) const {
+    std::shared_lock<std::shared_mutex> lock(_mutex);
+
     const auto it = _namespaceMap.find(name);
     if (it == _namespaceMap.end()) {
         return nullptr;
@@ -64,6 +73,8 @@ ProcedureNamespace* ProcedureManager::getNamespace(std::string_view name) const 
 }
 
 ProcedureNamespace* ProcedureManager::createNamespace(std::string_view name) {
+    std::unique_lock<std::shared_mutex> lock(_mutex);
+
     ProcedureNamespace* ns = new ProcedureNamespace(name);
     _namespaces.push_back(ns);
     _namespaceMap[name] = ns;
--- a/query/pipeline/procedures/ProcedureManager.h
+++ b/query/pipeline/procedures/ProcedureManager.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <memory>
+#include <mutex>
+#include <shared_mutex>
 #include <string_view>
 #include <unordered_map>
 #include <vector>
@@ -19,7 +21,7 @@ public:
 
     void init();
 
-    const Namespaces& namespaces() const { return _namespaces; }
+    void getNamespaces(Namespaces& result) const;
 
     const Procedure* getProcedure(std::string_view fullName) const;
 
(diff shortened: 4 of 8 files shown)


system/TuringConfig.cpp
```diff
@@ -1,11 +1,47 @@
...
+    }
+
+    char buf[4096];
+#ifdef __APPLE__
+    uint32_t bufSize = sizeof(buf);
+    int ok = _NSGetExecutablePath(buf, &bufSize) == 0;
+#else
+    ssize_t bufSize = readlink("/proc/self/exe", buf, sizeof(buf) - 1);
+    int ok = bufSize > 0;
+    if (ok) buf[bufSize] = '\0';
```
rjb32: use braces

system/TuringConfig.cpp
```diff
@@ -1,11 +1,47 @@
...
+    const char* envDir = getenv("TURING_EXTENSIONS_DIR");
+    if (envDir && *envDir) {
+        result = fs::Path(std::string(envDir));
+        return;
+    }
+
+    char buf[4096];
+#ifdef __APPLE__
+    uint32_t bufSize = sizeof(buf);
+    int ok = _NSGetExecutablePath(buf, &bufSize) == 0;
```
rjb32: ok can be const

ghost:
It looks pretty good! I think we should add a mutex to extensions manager to deal with concurrent access?

extensions/ExtensionManager.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return _installedMap.find(name) != _installedMap.end();
+}
+
+void ExtensionManager::loadExtensionDef(const TuringExtensionDef* def,
+                                        ExtensionDescriptor::Handle handle) {
+    const char* nsName = def->_namespaceName;
+
+    spdlog::info("Loading extension '{}'", nsName);
+
+    ProcedureNamespace* ns = _procedures->createNamespace(nsName);
```
ghost: This would break if we were to add multiple procedures that belong to the same namespace - we would need a check in ProcedureManager I think.

Also what did you have in mind for loading a library of procedures - would it be it's own singe .so file or would it be a directory of .so files?

rjb32:
> It looks pretty good! I think we should add a mutex to extensions manager to deal with concurrent access?

Indeed very good point

extensions/ExtensionManager.cpp
rjb32: It is for loading a library of procedures? All the content of the namespace is in its own .so

rjb32: We create a namespace and it is passed to the callback of the .so that populates it with whatever they want

rjb32:
Yes indeed it lacks concurrency protection. Thank you I need to look into it more

query/pipeline/processors/InstallExtensionProcessor.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+void InstallExtensionProcessor::prepare(ExecutionContext* ctxt) {
+    _ctxt = ctxt;
+
+    markAsPrepared();
+}
+
+void InstallExtensionProcessor::reset() {
+}
```
cyrusknopf, line 47: Do we need `markAsReset()` here?

extensions/ExtensionManager.cpp
ghost: Ah I see that we can load whatever we want in the init callback. Do we want to allow a user to upload different procedures for a namespace separately? Either way we should add a check before createNamespace() to see if the namespace already exists.

system/TuringConfig.cpp
```diff
@@ -1,11 +1,49 @@
...
+namespace {
+
+void resolveInstallExtensionsDir(fs::Path& result) {
+    const char* envDir = getenv("TURING_EXTENSIONS_DIR");
+    if (envDir && *envDir) {
+        result = fs::Path(std::string(envDir));
+        return;
+    }
+
+    char buf[4096];
```
cyrusknopf: Would be good to give 4096 a meaningful name

system/TuringConfig.cpp
```diff
@@ -1,11 +1,49 @@
...
+    ssize_t bufSize = readlink("/proc/self/exe", buf, sizeof(buf) - 1);
+    const int ok = bufSize > 0;
+    if (ok) {
+        buf[bufSize] = '\0';
+    }
+#endif
+    if (ok) {
+        const fs::Path exePath(std::string{buf});
+        result = exePath.parent().parent() / "lib" / "turingdb" / "extensions";
+        return;
```
cyrusknopf, line 40: Why do we use this directory?

rjb32: Because we have extensions in the install directory of turingdb and extensions in the user turing directory. To be able to ship with extensions. So when turingdb is installed arbitrarily anywhere with the wheel, we need a way to find back where is the right "lib" directory. So one way is to do relatively to the binary. I am open to better ideas

LucLabarriere: It seems to make sense to me, two other possibilities:

- we could also use an environment variable like TURING_EXT_PREFIX that would store a list of paths to search through to discover extensions. So that by default, we look in the install dir, and we also search for alternative prefix paths using that environment variable
- or we also search in the -turing-dir path "extensions/" subfolder (I could imagine a python script that, if it has the right shape, it can be invoked as an extension if it is placed in the extensions/ subfolder)

In any case, personally I like having a lib directory in the install folder by default

LucLabarriere: I see you actually already did the user extensions dir thing, nice!

rjb32: For clarity, here is the search order:
* Installation dir extensions: by default uses <exe path>/../lib  or TURING_EXTENSIONS_DIR env var if present
* User extensions: search in turing dir/extensions

query/pipeline/processors/InstallExtensionProcessor.cpp
rjb32: Thanks done

system/TuringConfig.cpp
rjb32: Thanks done

rjb32:
@suhassn-turing-bio I added locking in ExtensionManager, ProcedureManager and ProcedureNamespace.

PR #471, opened 2026-03-03
Load csv create2

rjb32:


query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -263,14 +227,22 @@ void WriteProcessor::createNodes(size_t numIters) {
...
-            // Add each property; NOTE: for now all ColumnConst, so all same value.
-            // Extract the value first, then add that value to each node to create.
-            // TODO: More cache friendly access patterns
-            for (const CommitWriteBuffer::UntypedProperty& prop : constProps) {
-                for (size_t i = 0; i < numIters; i++) {
-                    auto& pendingNode = _writeBuffer->getPendingNode(numPendingNodesPrior + i);
-                    pendingNode.properties.emplace_back(prop);
-                }
+            for (const auto& [name, type, valueCol] : node._properties) {
+                const PropertyTypeID propID =
```
rjb32: Don't overwrap, put on the same line

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -263,14 +227,22 @@ void WriteProcessor::createNodes(size_t numIters) {
...
+                    using Elem = typename std::remove_pointer_t<decltype(col)>::ValueType;
+
+                    for (size_t i = 0; i < numIters; i++) {
+                        const Elem& elem = (*col)[i];
+                        auto& pending = _writeBuffer->getPendingNode(numPendingNodesPrior + i);
+                        pending.properties.emplace_back(CommitWriteBuffer::UntypedProperty {propID, toPropertyVariant<Elem>(elem)});
+                    }
+                };
+                using Types = OutputtedTypes;
+                ColumnSingleDispatcher<Types::Allowed, decltype(populateNodeProps), Types::Excluded>
```
rjb32: Define a type alias for this ColumnSingleDispatcher instantiation

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -263,14 +227,21 @@ void WriteProcessor::createNodes(size_t numIters) {
...
-            // Add each property; NOTE: for now all ColumnConst, so all same value.
-            // Extract the value first, then add that value to each node to create.
-            // TODO: More cache friendly access patterns
-            for (const CommitWriteBuffer::UntypedProperty& prop : constProps) {
-                for (size_t i = 0; i < numIters; i++) {
-                    auto& pendingNode = _writeBuffer->getPendingNode(numPendingNodesPrior + i);
-                    pendingNode.properties.emplace_back(prop);
-                }
+            for (const auto& [name, type, valueCol] : node._properties) {
+                const PropertyTypeID propID = _metadataBuilder->getOrCreatePropertyType(name, type)._id;
```
cyrusknopf, line 231: As per my comment in the original code, we cannot do this processing of properties after adding the nodes to the write buffer, as this can lead to nodes in invalid state in the following case:

1. add x pending nodes to the write buffer (they are all uninitialised; invalid properties, labels, etc.)
2. try and parse properties, parsing fails, throws exeception
3. we now have x uninitialised, invalid pending nodes in the write buffer, UB when change gets committed.

This separation of adding nodes, then adding properties was done to be more cache friendly when iterating over the properties vector, but it has made it a bit complicated.

Perhaps I should rewrite it. An alternative is to return an error instead of throwing when creating properties, and to "undo" the pending nodes, then throw

PR #472, opened 2026-03-04
Fixed limit bug if processor receives an empty chunk

LucLabarriere:


rjb32:
Let's bring more eyes @cyrusknopf @suhassn-turing-bio

LucLabarriere:
CI doesn't pass so I guess I was too fast with the PR

ghost:
@LucLabarriere what was the bug?

query/pipeline/processors/LimitProcessor.cpp
```diff
@@ -57,11 +57,6 @@ void LimitProcessor::execute() {
 
         // Write rows of inputBlock that are below the limit
         const size_t rowsToWrite = std::min(blockRowCount, remainingCapacity);
-        if (rowsToWrite == 0) {
-            _reachedLimit = true;
-            return;
-        }
```
cyrusknopf, line 63: Does this fix it?

~~~cpp
if (rowsToWrite == 0 && _input.getPort()->isClosed()) {
    _reachedLimit = true;
    return;
}
~~~

LucLabarriere: It seems to be better yeah, I'll test that because if I remove it entirely, the processor seems to execute and output empty chunks

LucLabarriere:
> @LucLabarriere what was the bug?

The issue is that, if the LIMIT processor receives an empty chunk, it assumes the stream is complete and it closes its input port. However, some processors may process output chunks, even though they still have data to process and they may still produce non-empty chunks (for example if a filter processor removes everything and outputs an empty chunk, it doesn't mean that it's done and it may still produce non-empty chunks afterwards).

query/pipeline/processors/LimitProcessor.cpp
cyrusknopf: Yeah maybe we need a `bool _hasWritten` member that means we only output an empty chunk if the entire input stream was empty.

LucLabarriere:
Ok guys @rjb32 @cyrusknopf @suhassn-turing-bio  I realize now that the bug cannot be fixed with the current design for 1 particular case. What we are supposed to do (and can kind of do but not quite):
- `rowsToWrite = std::min(avail, dfRowCount)`
- `writeRows(inputDf, outputDf, rowCount)`
- `writeData()`
- `if (remaining == 0) || (input.closed()) { _input.close(); }`

This works in 99% of cases, except for one case, if we run this query:

```cypher
RETURN 5 limit 0
```

`rowsToWrite = 0`, we run `writeRows(inputDf, outputDf, 0)`. That actually does nothing since the dataframe contains a single column: a `ColumnConst<uint64_t>`. Then in the output lambda, we have to retrieve the logical row count from the data frame. This is done by looping over the columns in the dataframe, finding the first non column vector column, and retrieving its size. If the dataframe only has a single `ColumnConst<uint64_t>`, the dataframe is assumed to be of length 1 (there is no way to have a dataframe with a single columnconst with length != 1). So the query returns a column with 1 value `5` instead of zero values

Right now it works because there is a silent bug in the output query, the output is an empty data array `[]`, (the query doesn't produce a single chunk), which is incorrect, it should produce a single empty chunk

cyrusknopf:
> Ok guys @rjb32 @cyrusknopf @suhassn-turing-bio I realize now that the bug cannot be fixed with the current design for 1 particular case. What we are supposed to do (and can kind of do but not quite):
> 
> * `rowsToWrite = std::min(avail, dfRowCount)`
> * `writeRows(inputDf, outputDf, rowCount)`
> * `writeData()`
> * `if (remaining == 0) || (input.closed()) { _input.close(); }`
> 
> This works in 99% of cases, except for one case, if we run this query:
> 
> ```cypher
> RETURN 5 limit 0
> ```
> 
> `rowsToWrite = 0`, we run `writeRows(inputDf, outputDf, 0)`. That actually does nothing since the dataframe contains a single column: a `ColumnConst<uint64_t>`. Then in the output lambda, we have to retrieve the logical row count from the data frame. This is done by looping over the columns in the dataframe, finding the first non column vector column, and retrieving its size. If the dataframe only has a single `ColumnConst<uint64_t>`, the dataframe is assumed to be of length 1 (there is no way to have a dataframe with a single columnconst with length != 1). So the query returns a column with 1 value `5` instead of zero values
> 
> Right now it works because there is a silent bug in the output query, the output is an empty data array `[]`, (the query doesn't produce a single chunk), which is incorrect, it should produce a single empty chunk

I see, what if we specialise `writeRows(..., ..., 0)` to change any `ColumnConst`s to `ColumnVector`s of size 0? Since we have no notion of an "empty" `ColumnConst`, I think we would have to change the type.

rjb32:
> Ok guys @rjb32 @cyrusknopf @suhassn-turing-bio I realize now that the bug cannot be fixed with the current design for 1 particular case. What we are supposed to do (and can kind of do but not quite):
> 
> * `rowsToWrite = std::min(avail, dfRowCount)`
> * `writeRows(inputDf, outputDf, rowCount)`
> * `writeData()`
> * `if (remaining == 0) || (input.closed()) { _input.close(); }`
> 
> This works in 99% of cases, except for one case, if we run this query:
> 
> ```cypher
> RETURN 5 limit 0
> ```
> 
> `rowsToWrite = 0`, we run `writeRows(inputDf, outputDf, 0)`. That actually does nothing since the dataframe contains a single column: a `ColumnConst<uint64_t>`. Then in the output lambda, we have to retrieve the logical row count from the data frame. This is done by looping over the columns in the dataframe, finding the first non column vector column, and retrieving its size. If the dataframe only has a single `ColumnConst<uint64_t>`, the dataframe is assumed to be of length 1 (there is no way to have a dataframe with a single columnconst with length != 1). So the query returns a column with 1 value `5` instead of zero values
> 
> Right now it works because there is a silent bug in the output query, the output is an empty data array `[]`, (the query doesn't produce a single chunk), which is incorrect, it should produce a single empty chunk

Ok so you mean writeRows(inputDf, outputDf, 0) still writes a const if the df contains a const?

LucLabarriere:
> > Ok guys @rjb32 @cyrusknopf @suhassn-turing-bio I realize now that the bug cannot be fixed with the current design for 1 particular case. What we are supposed to do (and can kind of do but not quite):
> > 
> > * `rowsToWrite = std::min(avail, dfRowCount)`
> > * `writeRows(inputDf, outputDf, rowCount)`
> > * `writeData()`
> > * `if (remaining == 0) || (input.closed()) { _input.close(); }`
> > 
> > This works in 99% of cases, except for one case, if we run this query:
> > ```cypher
> > RETURN 5 limit 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `rowsToWrite = 0`, we run `writeRows(inputDf, outputDf, 0)`. That actually does nothing since the dataframe contains a single column: a `ColumnConst<uint64_t>`. Then in the output lambda, we have to retrieve the logical row count from the data frame. This is done by looping over the columns in the dataframe, finding the first non column vector column, and retrieving its size. If the dataframe only has a single `ColumnConst<uint64_t>`, the dataframe is assumed to be of length 1 (there is no way to have a dataframe with a single columnconst with length != 1). So the query returns a column with 1 value `5` instead of zero values
> > Right now it works because there is a silent bug in the output query, the output is an empty data array `[]`, (the query doesn't produce a single chunk), which is incorrect, it should produce a single empty chunk
> 
> Ok so you mean writeRows(inputDf, outputDf, 0) still writes a const if the df contains a const?

Yes exactly, the actual call is this:

```c++
outputDf->copyFromLine(inputDf, 0, rowsToWrite);
```

it doesn't change the layout of the dataframe, it only ensures that it has the right rows.  It works perfectly. The issue is that in the output lambda processor, when we try to output the dataframe, we read its line count by iterating over it and retrieve the size of the first vector we find. If the dataframe only contains ColumnConst cols, then it returns 1. There is no way for a dataframe to return a row count != 1 if it has only ColumnConst columns

LucLabarriere:
> > Ok guys @rjb32 @cyrusknopf @suhassn-turing-bio I realize now that the bug cannot be fixed with the current design for 1 particular case. What we are supposed to do (and can kind of do but not quite):
> > 
> > * `rowsToWrite = std::min(avail, dfRowCount)`
> > * `writeRows(inputDf, outputDf, rowCount)`
> > * `writeData()`
> > * `if (remaining == 0) || (input.closed()) { _input.close(); }`
> > 
> > This works in 99% of cases, except for one case, if we run this query:
> > ```cypher
> > RETURN 5 limit 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `rowsToWrite = 0`, we run `writeRows(inputDf, outputDf, 0)`. That actually does nothing since the dataframe contains a single column: a `ColumnConst<uint64_t>`. Then in the output lambda, we have to retrieve the logical row count from the data frame. This is done by looping over the columns in the dataframe, finding the first non column vector column, and retrieving its size. If the dataframe only has a single `ColumnConst<uint64_t>`, the dataframe is assumed to be of length 1 (there is no way to have a dataframe with a single columnconst with length != 1). So the query returns a column with 1 value `5` instead of zero values
> > Right now it works because there is a silent bug in the output query, the output is an empty data array `[]`, (the query doesn't produce a single chunk), which is incorrect, it should produce a single empty chunk
> 
> I see, what if we specialise `writeRows(..., ..., 0)` to change any `ColumnConst`s to `ColumnVector`s of size 0? Since we have no notion of an "empty" `ColumnConst`, I think we would have to change the type.

I don't think it would work and it would be quite a hack. I think the issue is the same as before, the row count should probably be encoded in the dataframe directly and getRowCount() should just return that value

cyrusknopf:
> > > Ok guys @rjb32 @cyrusknopf @suhassn-turing-bio I realize now that the bug cannot be fixed with the current design for 1 particular case. What we are supposed to do (and can kind of do but not quite):
> > > 
> > > * `rowsToWrite = std::min(avail, dfRowCount)`
> > > * `writeRows(inputDf, outputDf, rowCount)`
> > > * `writeData()`
> > > * `if (remaining == 0) || (input.closed()) { _input.close(); }`
> > > 
> > > This works in 99% of cases, except for one case, if we run this query:
> > > ```cypher
> > > RETURN 5 limit 0
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > `rowsToWrite = 0`, we run `writeRows(inputDf, outputDf, 0)`. That actually does nothing since the dataframe contains a single column: a `ColumnConst<uint64_t>`. Then in the output lambda, we have to retrieve the logical row count from the data frame. This is done by looping over the columns in the dataframe, finding the first non column vector column, and retrieving its size. If the dataframe only has a single `ColumnConst<uint64_t>`, the dataframe is assumed to be of length 1 (there is no way to have a dataframe with a single columnconst with length != 1). So the query returns a column with 1 value `5` instead of zero values
> > > Right now it works because there is a silent bug in the output query, the output is an empty data array `[]`, (the query doesn't produce a single chunk), which is incorrect, it should produce a single empty chunk
> > 
> > 
> > I see, what if we specialise `writeRows(..., ..., 0)` to change any `ColumnConst`s to `ColumnVector`s of size 0? Since we have no notion of an "empty" `ColumnConst`, I think we would have to change the type.
> 
> I don't think it would work and it would be quite a hack. I think the issue is the same as before, the row count should probably be encoded in the dataframe directly and getRowCount() should just return that value

The problem is that there is an arbitrary number of ways that a Column* - which belongs to a dataframe - can be resized without ever "notifying" (whatever that means) the dataframe for it to update that encoded row count.

LucLabarriere:
@rjb32 @cyrusknopf @suhassn-turing-bio

Updated the fix so that now, only `LIMIT 0` is an issue and only if the dataframe contains only ColumnConst columns. Created an issue for it #485.

Among possible options:
- Have a row count in the dataframe, but as @cyrusknopf said, it's really hard to know when the columns in the dataframe changed and to notify the dataframe to update its row count. I think it would be possible to update the value at strategic points in the pipeline (like right before projection), but I'm not event sure that's realistic
- Replace all ColumnConsts in the dataframe by ColumnVector when we do LIMIT (@cyrusknopf idea), could work but I'm a bit worried of having to allocate new columns + we would have to make sure the tags are the same, and it invalidates column pointers used here and there which is something else we have to be careful of (if we store Column* in the pipeline processors for example). This is especially tricky since the pipeline might re-run after the LIMIT execution (may still be possible though)
- We could have a limitRowCount in the dataframe (quite a hack too) that is only set when we call LIMIT on it, and return that value when we do getLogicalRowCount() if it is set (not sure this is practical either)

In the meantime, I disabled the RETURN 5 LIMIT 0 unit test. Do you think that this is acceptable? The behaviour now is:
- Query that is supposed to return no data because there is no Person --> Vehicle in poledb: `MATCH (n:Person)-->(m:Vehicle) RETUN n, m LIMIT 5` will return data: [[[], []]] (one chunk with two empty columns, instead of data: [] (no chunks at all)
- RETURN 5 LIMIT 0: bug, returns data: [[[0]]] instead of [[[]]]

cyrusknopf:
> @rjb32 @cyrusknopf @suhassn-turing-bio
> 
> Updated the fix so that now, only `LIMIT 0` is an issue and only if the dataframe contains only ColumnConst columns. Created an issue for it #485.
> 
> Among possible options:
> 
> * Have a row count in the dataframe, but as @cyrusknopf said, it's really hard to know when the columns in the dataframe changed and to notify the dataframe to update its row count. I think it would be possible to update the value at strategic points in the pipeline (like right before projection), but I'm not event sure that's realistic
> * Replace all ColumnConsts in the dataframe by ColumnVector when we do LIMIT (@cyrusknopf idea), could work but I'm a bit worried of having to allocate new columns + we would have to make sure the tags are the same, and it invalidates column pointers used here and there which is something else we have to be careful of (if we store Column* in the pipeline processors for example). This is especially tricky since the pipeline might re-run after the LIMIT execution (may still be possible though)
> * We could have a limitRowCount in the dataframe (quite a hack too) that is only set when we call LIMIT on it, and return that value when we do getLogicalRowCount() if it is set (not sure this is practical either)
> 
> In the meantime, I disabled the RETURN 5 LIMIT 0 unit test. Do you think that this is acceptable? The behaviour now is:
> 
> * Query that is supposed to return no data because there is no Person --> Vehicle in poledb: `MATCH (n:Person)-->(m:Vehicle) RETUN n, m LIMIT 5` will return data: [[[], []]] (one chunk with two empty columns, instead of data: [] (no chunks at all)
> * RETURN 5 LIMIT 0: bug, returns data: [[[0]]] instead of [[[]]]

Personally I think that it is fine to leave this `RETURN 5 LIMIT 0` bug because it only occurs on silly queries. But I agree there doesn't seem to be an ideal solution right now

rjb32:
Sounds good to me you can merge @LucLabarriere

PR #476, opened 2026-03-04
Integrate visualizer into turingdb

rjb32:
Launch TuringDB Visualizer just with:
```
turingdb -ui
``` 

## How does it work?

* Just a small http proxy thread is spawned inside turingdb itself listening on http://localhost:8080 by default
* The user opens their browser to http://localhost:8080 to see the UI
* All the frontend code uses /api/* or /some_file.html relative URLs
* The small proxy intercepts all the /api/* routes and redirects them to the local turingdb endpoints, and any other route to https://vis.turingdb.org which is centrally serving the frontend on the public internet

**Benefit**: we don't have to package 30MB+ of frontend code and have npm/bun present on the user machine. The UI can also be updated centrally without requiring users to git pull

## Settings

Change UI port: 
```
turingdb -ui -ui-port 7777
``` 

Change the address at which the frontend is served:
```
TURINGDB_VIS_URL=http://custom.com:8888 turingdb -ui
```


server/VisualizerProxy.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+size_t headerCallback(char* buffer, size_t size, size_t nitems, void* userdata) {
+    const size_t total = size * nitems;
+    auto* response = static_cast<CurlResponse*>(userdata);
+
+    std::string_view line(buffer, total);
+
+    static constexpr std::string_view contentTypeKey = "content-type:";
+    if (line.size() > contentTypeKey.size()) {
+        bool match = std::equal(contentTypeKey.begin(), contentTypeKey.end(),
```
rjb32: match can be const

server/VisualizerProxy.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+size_t headerCallback(char* buffer, size_t size, size_t nitems, void* userdata) {
+    const size_t total = size * nitems;
+    auto* response = static_cast<CurlResponse*>(userdata);
+
+    std::string_view line(buffer, total);
+
+    static constexpr std::string_view contentTypeKey = "content-type:";
+    if (line.size() > contentTypeKey.size()) {
```
rjb32, line 39: Please add comment to explain what this is doing

server/VisualizerProxy.h
```diff
@@ -0,0 +1,30 @@
...
+    VisualizerProxy(const std::string& address, uint32_t proxyPort, uint32_t dbPort);
+    ~VisualizerProxy();
+
+    void start();
+    void stop();
+    void wait();
+
+private:
+    std::string _address;
+    uint32_t _proxyPort;
```
rjb32: Give a default value to every POD member

tools/turingdb/TuringDBTool.cpp
```diff
@@ -139,9 +123,17 @@ int main(int argc, const char** argv) {
...
+        std::unique_ptr<VisualizerProxy> proxy;
+        if (launchUI) {
+            proxy = std::make_unique<VisualizerProxy>(address, uiPort, port);
+            proxy->start();
+        }
+
         // CLI Shell
         if (demonize) {
             server.wait();
+            if (proxy) proxy->stop();
```
rjb32: Always use braces for control structures

rjb32:
Basically it uses our HttpServer class to make a VisualizerProxy class inside the turingdb process in its own thread.

@suhassn-turing-bio I think the only controversial thing here may be that the proxy uses curl to fetch content from the frontend host and then writing the content using NetWriter. But this way we don't have to use express or ship nginx. This feels very fast in the browser but this is not very sophisticated.

LucLabarriere:
That seems really good to me, and the HTTP server side can easily be swapped out for something else if we need someday. I'm actually wondering, do you think we could spawn it by default and make the argument `-no-ui` to not spawn UI server instead? I'm thinking it could make people use it more

LucLabarriere:
I'm only wondering how it works on the visualizer side of things though (it does not seem to work on my machine?). When I made the app, I used URL rewriting to change /api calls to point to localhost:6666 but i'm pretty sure that's a vite/nginx thing. Here we would need the SPA itself to rewrite to point to localhost right? And in this case we would need to make it dynamic so that it access any provided port

rjb32:
> I'm only wondering how it works on the visualizer side of things though (it does not seem to work on my machine?). When I made the app, I used URL rewriting to change /api calls to point to localhost:6666 but i'm pretty sure that's a vite/nginx thing. Here we would need the SPA itself to rewrite to point to localhost right? And in this case we would need to make it dynamic so that it access any provided port

The rewriting is dynamic and uses the port of the DB server from TuringConfig when the forwarding URL for /api/* routes is constructed by VisualizerProxy.

Where does it break on your machine?
* Can you access vis.turingdb.org without anything running first? The frontend should show but no reaction and no data.
* turingdb -ui and you go to the URL shown in Visualizer ... line what do you see?
* Are you running on the same machine as your browser is?

LucLabarriere:
> > I'm only wondering how it works on the visualizer side of things though (it does not seem to work on my machine?). When I made the app, I used URL rewriting to change /api calls to point to localhost:6666 but i'm pretty sure that's a vite/nginx thing. Here we would need the SPA itself to rewrite to point to localhost right? And in this case we would need to make it dynamic so that it access any provided port
> 
> The rewriting is dynamic and uses the port of the DB server from TuringConfig when the forwarding URL for /api/* routes is constructed by VisualizerProxy.
> 
> Where does it break on your machine?
> 
> * Can you access vis.turingdb.org without anything running first? The frontend should show but no reaction and no data.
> * turingdb -ui and you go to the URL shown in Visualizer ... line what do you see?
> * Are you running on the same machine as your browser is?

Oh actually you're right it does work, I was incorrectly going to https://vis.turingdb.org/ directly

PR #478, opened 2026-03-04
Implementing value hash join

rjb32:
To replace CARTESIAN PRODUCT + FILTER by JOIN on property values.

Supports joins between all the types defined in PropertyType.h:
* NodeID
* int64_t
* uint64_t
* double
* std::string_view
* CustomBool
and std::optional<> forms of these types.

Example: What are the persons with the same age in the company?
```
MATCH (a:Person), (b:Person) WHERE a.age = b.age RETURN a.name, b.name, a.age
```
image

# Performance comparison

Benchmark on reactome graph (6 queries, 5 rounds each, LIMIT 1000).

| Query | Without VHJ (median) | With VHJ (median) | Speedup |
|-------|---------------------:|-------------------:|--------:|
| pathways_sharing_reaction | 13,175 ms | 403 ms | **33x** |
| shared_complex_components | 223 ms | 265 ms | 0.84x |
| same_species_pathway_join | 15.1 ms | 33.4 ms | 0.45x |
| cross_species_gene_products | 11,587 ms | 239 ms | **49x** |
| same_compartment_reactions | 102 ms | 121 ms | 0.84x |

**Massive speedups** on high-cardinality join keys (Q1: 33x, Q4: 49x) where the cartesian product O(N*M) is expensive.

**Slower** on low-cardinality join keys (Q3: 95 species, Q5: 156 compartments)

## Improvement

After cardinality estimation heuristic (small-product check + LIMIT-aware check) and per-tuple overhead reduction (transparent string lookup, single column read per tuple):

| Query | CartesianProduct | VHJ | Speedup | Decision | Reason |
|-------|-----------------|-----|---------|----------|--------|
| Q1: pathways_sharing_reaction | 12,847 ms | 90 ms | **142x** | VHJ | 94k < LIMIT x 200 |
| Q2: shared_complex_components | 93 ms | 90 ms | **1.0x** | CARTESIAN | 406k > LIMIT x 200 |
| Q3: same_species_pathway_join | 7.4 ms | 7.4 ms | **1.0x** | CARTESIAN | 95 x 95 < 100k (small product) |
| Q4: cross_species_gene_products | 11,496 ms | 114 ms | **101x** | VHJ | 105k < LIMIT x 200 |
| Q5: same_compartment_reactions | 72 ms | 68 ms | **1.0x** | CARTESIAN | 156 x 156 < 100k (small product) |
| Q6: edge_id_join_hasEvent | 904 ms | 900 ms | **1.0x** | N/A | Edge ID join (not property value) |

All regressions eliminated. Q1/Q4 high-cardinality speedups improved (142x, 101x vs 33x, 49x previously).

## Queries

**pathways_sharing_reaction**
```cypher
MATCH (a:Pathway)-[:hasEvent]->(r1:ReactionLikeEvent), (b:Pathway)-[:hasEvent]->(r2:ReactionLikeEvent)
WHERE r1.stId = r2.stId AND a.dbId <> b.dbId
RETURN a.displayName, b.displayName, r1.stId LIMIT 1000
```

**shared_complex_components**
```cypher
MATCH (c1:Complex)-[:hasComponent]->(e1:PhysicalEntity), (c2:Complex)-[:hasComponent]->(e2:PhysicalEntity)
WHERE e1.displayName = e2.displayName AND c1.dbId <> c2.dbId
RETURN c1.displayName, c2.displayName, e1.displayName LIMIT 1000
```

**same_species_pathway_join**
```cypher
MATCH (p1:Pathway)-[:species]->(s1:Species), (p2:Pathway)-[:species]->(s2:Species)
WHERE s1.dbId = s2.dbId
RETURN p1.displayName, p2.displayName, s1.displayName LIMIT 1000
```

**cross_species_gene_products**
```cypher
MATCH (g1:ReferenceGeneProduct)-[:species]->(s1:Species), (g2:ReferenceGeneProduct)-[:species]->(s2:Species)
WHERE g1.geneName = g2.geneName AND s1.displayName <> s2.displayName
RETURN g1.identifier, g2.identifier, g1.geneName, s1.displayName, s2.displayName LIMIT 1000
```

**same_compartment_reactions**
```cypher
MATCH (r:ReactionLikeEvent)-[:input]->(p:PhysicalEntity)-[:compartment]->(c1:Compartment),
      (r2:ReactionLikeEvent)-[:output]->(p2:PhysicalEntity)-[:compartment]->(c2:Compartment)
WHERE c1.displayName = c2.displayName
RETURN r.displayName, r2.displayName, c1.displayName LIMIT 1000
```

**edge_id_join_hasEvent**
```cypher
MATCH (a:Pathway)-[e:hasEvent]->(b), (c)-[e]->(d)
RETURN a.displayName, b.displayName LIMIT 1000
```


query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -26,35 +26,56 @@ void fillOutputColumn(T* outputColumn,
...
 }
 
 template <typename Key>
 std::vector<RowOffset>& getMap(std::unordered_map<Key, std::vector<RowOffset>>& map,
                                Column* col,
-                               size_t index) {
+                               size_t index,
+                               bool isOptional) {
+    if (isOptional) {
+        return map[*(*static_cast<ColumnVector<std::optional<Key>>*>(col))[index]];
```
rjb32: Put the casted column pointer into a temporary

query/pipeline/PipelineBuilder.cpp
```diff
@@ -63,6 +64,12 @@ using namespace db;
 
 namespace {
 
+template <typename T>
+struct RemoveOptional { using type = T; };
```
rjb32: Isn't there a utility file where this should be better placed

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -26,35 +26,80 @@ void fillOutputColumn(T* outputColumn,
...
+        return (*typedCol)[index].has_value();
+    }
 }
 
-template <typename Key>
-std::vector<RowOffset>& getMap(std::unordered_map<Key, std::vector<RowOffset>>& map,
-                               Column* col,
-                               size_t index) {
-    return map[(*static_cast<const ColumnVector<Key>*>(col))[index]];
+// Key is the column's value type (e.g. string_view). The map's key type
```
ghost: Is this true? I don't think so.

ghost: ^I was mistaken

rjb32:
### Code review

Found 2 issues:

1. `processRightStream` loop guard uses `rightDf->getLogicalRowCount()` (line 503) but the post-loop increment guard uses `rightCol->size()` (line 585). These values diverge for `ColumnConst` columns (logical size > physical size), which would cause `_rightInputIdx` to be incorrectly incremented or not incremented, leading to skipped rows or out-of-bounds access. The same inconsistency exists in `processLeftStream` (loop at line 381 uses `leftCol->size()`, but the reset checks at lines 289/313 use `getLogicalRowCount()`). A previous commit `7545569ad` fixed this exact class of bug in `CartesianProductProcessor`.

https://github.com/turing-db/turingdb/blob/99d7c5d8312adc27d3ee43eb31535ac245ad3677/query/pipeline/processors/HashJoinProcessor.cpp#L502-L587

2. Garbled comment at line 442: `// inserting the row into  as _leftRowLen as we we want to copy the left input of` — contains a double space, a double "we we", and the sentence is not grammatically coherent, making the comment misleading rather than helpful.

https://github.com/turing-db/turingdb/blob/99d7c5d8312adc27d3ee43eb31535ac245ad3677/query/pipeline/processors/HashJoinProcessor.cpp#L440-L445

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

query/common/CardinalityEstimation.cpp
```diff
@@ -0,0 +1,41 @@
...
+bool CardinalityEstimation::shouldPreferCartesian(const LabelSet& left, const LabelSet& right, size_t queryLimit) const {
+    const size_t leftCount = estimateNodeCount(left);
+    const size_t rightCount = estimateNodeCount(right);
+
+    // Small cartesian product is cheap enough without a hash join
+    if (leftCount * rightCount < SMALL_CARTESIAN_THRESHOLD) {
+        return true;
+    }
+
+    // With a small LIMIT, cartesian can terminate early while VHJ
```
ghost, line 32: This isn't true - Hash Joins also stream output so we could terminate whenever the limit has been reached.

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -92,14 +159,62 @@ void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
...
+        if (leftDf->getColumn(namedCol->getTag()) != nullptr ||
+            namedCol->getTag() == _rightJoinKey) {
+            continue;
+        }
+        dispatchColumnVector(namedCol->getColumn(), [&](auto* col) {
+            using ColType = std::remove_pointer_t<decltype(col)>;
+            using ValType = typename ColType::ValueType;
+            if constexpr (std::is_trivially_copyable_v<ValType>) {
+                _rightRowByteSize += sizeof(ValType);
+            }
```
cyrusknopf, line 198: What if the `if constexpr` fails? Shouldn't we error?

query/plan/PipelineGenerator.cpp
```diff
@@ -875,44 +876,105 @@ PipelineOutputInterface* PipelineGenerator::translateJoinNode(JoinNode* node) {
...
+                leftJoinTag = leftIt->second;
+            } else {
+                const auto entityVisitor = Overloaded {
+                    [](const EntityOutputStream::NodeStream& s) -> ColumnTag {
+                        return s._nodeIDsTag;
+                    },
+                    [](const EntityOutputStream::EdgeStream& s) -> ColumnTag {
+                        return s._edgeIDsTag;
+                    },
+                };
```
cyrusknopf: Since these lambdas don't need to capture any local state, can we define the entire `entryVisitor` elsewhere and use it here? I think it will read cleaner

query/common/CardinalityEstimation.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "reader/GraphReader.h"
+
+using namespace db;
+
+CardinalityEstimation::CardinalityEstimation(GraphView graphView)
+    : _graphView(graphView)
+{
+}
+
+size_t CardinalityEstimation::estimateNodeCount(const LabelSet& labelset) const {
```
ghost, line 14: This is interesting and good to start with but should definitely be refined over time.

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -26,35 +26,101 @@ void fillOutputColumn(T* outputColumn,
...
+// Looks up or inserts the key in the map, returning the offset vector.
+// For string keys, uses transparent find to avoid allocation when the key exists.
+std::vector<RowOffset>& getMap(auto& map, const auto& key) {
+    using KeyType = std::decay_t<decltype(key)>;
+    if constexpr (std::is_same_v<KeyType, std::string_view>) {
+        const auto it = map.find(key);
+        if (it != map.end()) {
+            return it->second;
+        }
+        return map.emplace(std::string(key), std::vector<RowOffset>()).first->second;
```
ghost: Are we ok allocating a string when we know the column is of type string_view and thus will have the underlying string data backed through the whole pipeline?

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -26,35 +26,101 @@ void fillOutputColumn(T* outputColumn,
...
 }
 
-HashJoinProcessor* HashJoinProcessor::create(PipelineV2* pipeline,
-                                             const ColumnTag leftJoinKey,
-                                             const ColumnTag rightJoinKey) {
+template <typename LeftKey, typename RightKey>
+auto HashJoinProcessor<LeftKey, RightKey>::create(PipelineV2* pipeline,
+                                                  const ColumnTag leftJoinKey,
+                                                  const ColumnTag rightJoinKey)
+    -> HashJoinProcessor* {
```
ghost: Trailing return types seems out of place here?

query/pipeline/processors/LimitProcessor.cpp
```diff
@@ -35,6 +36,7 @@ LimitProcessor* LimitProcessor::create(PipelineV2* pipeline, size_t limit) {
 }
 
 void LimitProcessor::prepare(ExecutionContext* ctxt) {
+    ctxt->setChunkSize(std::min(ctxt->getChunkSize(), _limit));
```
ghost: Should we have this in a different PR?

query/plan/ReadStmtGenerator.cpp
```diff
@@ -575,6 +598,133 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+            for (PlanGraphNode* output : outputs) {
+                input->connectOut(output);
+            }
+        }
+
+        _tree->removeNode(filterNode);
+    }
+}
+
+bool ReadStmtGenerator::tryPlaceValueHashJoin(Predicate* pred,
```
ghost: This whole function has to be moved into `placePredicateJoins()` and `insertDataFlowNode()`.  This is a bit of a redundant function.

query/plan/ReadStmtGenerator.cpp
```diff
@@ -575,6 +598,133 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+                                               VarNode* var) {
+    // Must have exactly 2 variable dependencies and no function dependencies
+    auto& varDeps = deps.getVarDeps();
+    if (varDeps.size() != 2 || !deps.getFuncDeps().empty()) {
+        return false;
+    }
+
+    // Must be a BinaryExpr with Equal operator
+    const Expr* expr = pred->getExpr();
+    if (expr->getKind() != Expr::Kind::BINARY) {
```
ghost: These checks are nice and can be moved into `placePredicateJoins()`

query/plan/ReadStmtGenerator.cpp
```diff
@@ -575,6 +598,133 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+    const auto* binExpr = static_cast<const BinaryExpr*>(expr);
+    if (binExpr->getOperator() != BinaryOperator::Equal) {
+        return false;
+    }
+
+    // Both dep expressions must resolve to a single VarDecl (simple property/symbol)
+    auto& dep0 = varDeps[0];
+    auto& dep1 = varDeps[1];
+
+    if (!dep0._expr->getExprVarDecl() || !dep1._expr->getExprVarDecl()) {
```
ghost: these checks as well could be moved to `placePredicateJoins()`

query/plan/ReadStmtGenerator.cpp
```diff
@@ -575,6 +598,133 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+    }
+
+    // Check which dep is on a different island from var (NoPath = remote)
+    const auto [path0, ancestor0] = _topology->getShortestPath(var, dep0._producerNode);
+    const auto [path1, ancestor1] = _topology->getShortestPath(var, dep1._producerNode);
+
+    ExprDependencies::VarDependency* localDep = nullptr;
+    ExprDependencies::VarDependency* remoteDep = nullptr;
+
+    if (path0 == PlanGraphTopology::PathToDependency::NoPath &&
```
ghost: This checking of which islands the dependencies are on is a bit confusing. If you look at what happens when we find the commonSucessor between the dependencies in the `placePredicateJoin()` function (in the case where we are placing a predicate join) the VarNode returned from that function (which is passed to this function as the `VarNode* var` parameter) is either a common successor of these two dependencies or if there is no common successor it returns the lhs var node of the binary expression. so this if block would only ever fall to the else if condition (in the case where we would want to do a value hash join) or the else which would return out of the function. (Overall I think this implementation looks and works correctly on the surface but is very unclear) we can discuss a bit more about it in person.

query/plan/ReadStmtGenerator.cpp
```diff
@@ -575,6 +598,133 @@ void ReadStmtGenerator::placePredicateJoins() {
         filterNode->addPredicate(pred.get());
         pred->setFilterNode(filterNode);
     }
+
+    // Bypass filters left empty by value hash joins.
+    // We do this here rather than in a general optimizer pass to avoid
+    // masking planner bugs that accidentally produce empty filters.
+    for (FilterNode* filterNode : vhjFilters) {
```
ghost: instead of this we could have a `_tree->replace(PlanGraphNode* node)` function that we use to replace to insert the ValueHashJoin.

rjb32:
First fix the table of results

query/common/CardinalityEstimation.cpp
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    // With a small LIMIT, cartesian can terminate early while VHJ
+    // must process full chunks from both sides
+    if (queryLimit > 0 && queryLimit <= MAX_LIMIT_FOR_VHJ_HEURISTIC) {
+        const size_t largerSide = std::max(leftCount, rightCount);
+        if (largerSide > queryLimit * VHJ_LIMIT_RATIO) {
+            return true;
+        }
+    }
```
cyrusknopf, line 61: Do we still want this?

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -26,35 +26,101 @@ void fillOutputColumn(T* outputColumn,
...
+    } else {
+        if constexpr (IsOptional) {
+            auto* typedCol = static_cast<ColumnVector<std::optional<Key>>*>(col);
+            return *(*typedCol)[index];
+        } else {
+            auto* typedCol = static_cast<const ColumnVector<Key>*>(col);
+            return (*typedCol)[index];
+        }
+    }
+}
```
cyrusknopf: Can all `typedCol` be const? Also I don't like the static cast here, its a bit brittle using `IsOptional` as some kind of proxy to determine the type. I would prefer using the ColumnDispatcher, but I am not sure you could achieve the same variant return types as you do here. At the very least I think we should use dynamic cast and check

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -26,35 +26,101 @@ void fillOutputColumn(T* outputColumn,
...
+std::vector<RowOffset>& getMap(auto& map, const auto& key) {
+    using KeyType = std::decay_t<decltype(key)>;
+    if constexpr (std::is_same_v<KeyType, std::string_view>) {
+        const auto it = map.find(key);
+        if (it != map.end()) {
+            return it->second;
+        }
+        return map.emplace(std::string(key), std::vector<RowOffset>()).first->second;
+    } else {
+        return map[key];
```
cyrusknopf, line 78: The return here is very hard to follow: `emplace(...).first->second`. I think it needs to split up. Also can you replace `std::vector<RowOffset>()` with just `{}`?

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -26,35 +26,101 @@ void fillOutputColumn(T* outputColumn,
...
+void fillJoinColumn(Column* outputColumn,
+                    Column* inputColumn,
+                    size_t rowIndex,
+                    size_t offset,
+                    size_t rowsToCopy) {
+    dispatchColumnVector(outputColumn, [&](auto* outCol) {
+        auto* typedOut = static_cast<decltype(outCol)>(outputColumn);
+        using OutVal = typename std::decay_t<decltype(*outCol)>::ValueType;
+
+        dispatchColumnVector(inputColumn, [&](auto* inCol) {
```
cyrusknopf, line 93: The line
`auto* typedOut = static_cast<decltype(outCol)>(outputColumn);`
is redundant. `outCol` is already the type you need. You can replace uses of `typedOut` with just `outCol`.
Also debate as to whether we should use the new Column dispatcher or not

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -26,35 +26,101 @@ void fillOutputColumn(T* outputColumn,
...
+            auto* typedIn = static_cast<decltype(inCol)>(inputColumn);
+            using InVal = typename std::decay_t<decltype(*inCol)>::ValueType;
+
+            if constexpr (std::is_constructible_v<OutVal, const InVal&>) {
+                bioassert(typedOut->size() >= offset + rowsToCopy,
+                          "rows to copy do not fit in output column");
+                std::fill_n(typedOut->begin() + offset, rowsToCopy,
+                            OutVal((*typedIn)[rowIndex]));
+            }
+        });
```
cyrusknopf, line 104: I think there needs to be some check here if the `if constexpr` is false

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -92,14 +159,62 @@ void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
     //the size of the output dataframe
     _rightRowLen = outDf->size() - _leftRowLen - 1;
 
+    // Compute actual byte sizes of rows stored in the RowStore
+    _leftRowByteSize = 0;
+    for (const auto* namedCol : leftDf->cols()) {
+        if (namedCol->getTag() == _leftJoinKey) {
+            continue;
+        }
```
cyrusknopf, line 171: Would be worth adding a comment as to why we skip the join key

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -92,14 +159,62 @@ void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
...
+        dispatchColumnVector(namedCol->getColumn(), [&](auto* col) {
+            using ColType = std::remove_pointer_t<decltype(col)>;
+            using ValType = typename ColType::ValueType;
+            if constexpr (std::is_trivially_copyable_v<ValType>) {
+                _leftRowByteSize += sizeof(ValType);
+            }
+        });
```
cyrusknopf, line 180: Need a check for the `if constexpr` failing?

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -92,14 +159,62 @@ void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
...
+    }
+
+    const auto* rightDf = _rightInput.getDataframe();
+
+    _rightRowByteSize = 0;
+    for (const auto* namedCol : rightDf->cols()) {
+        if (leftDf->getColumn(namedCol->getTag()) != nullptr ||
+            namedCol->getTag() == _rightJoinKey) {
+            continue;
+        }
```
cyrusknopf, line 190: This condition is quite complex and I think needs explaining

query/pipeline/processors/HashJoinProcessor.cpp
cyrusknopf: Still relevant

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -92,14 +159,62 @@ void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
...
+            using ValType = typename ColType::ValueType;
+            if constexpr (std::is_trivially_copyable_v<ValType>) {
+                _rightRowByteSize += sizeof(ValType);
+            }
+        });
+    }
+
+    // Detect if the join key columns are optional (nullable)
+    {
+        auto* leftKeyCol = leftDf->getColumn(_leftJoinKey)->getColumn();
```
cyrusknopf: Can be const. Also I think we should avoid using auto here

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -92,14 +159,62 @@ void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
...
+        const auto leftKind = leftKeyCol->getKind();
+        _isLeftOptionalKey = (leftKind == ColumnVector<std::optional<LeftKey>>::staticKind());
+    }
+
+    {
+        auto* rightKeyCol = rightDf->getColumn(_rightJoinKey)->getColumn();
```
cyrusknopf: Can be const, avoid auto again IMO

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -153,34 +268,12 @@ void HashJoinProcessor::execute() {
...
-                                                  totalRowsInserted + totalSizeIncrease);
-            for (const auto* namedCol : outDf->cols()) {
-                dispatchColumnVector(namedCol->getColumn(),
-                                     [&](auto* col) {
-                                         auto* colVector = static_cast<decltype(col)>(col);
-                                         colVector->resize(newOutputSize);
-                                     });
-            }
+        if (_isLeftOptionalKey) {
+            processLeftStream<true>(rowsRemaining, totalRowsInserted);
```
cyrusknopf: Would be clearer if there is a `constexpr bool` with a meaningful name instead of `true` and `false`.

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -205,30 +298,12 @@ void HashJoinProcessor::execute() {
...
-                dispatchColumnVector(namedCol->getColumn(),
-                                     [&](auto* col) {
-                                         auto* colVector = static_cast<decltype(col)>(col);
-                                         colVector->resize(newOutputSize);
-                                     });
-            }
+        if (_isRightOptionalKey) {
+            processRightStream<true>(rowsRemaining, totalRowsInserted);
+        } else {
+            processRightStream<false>(rowsRemaining, totalRowsInserted);
```
cyrusknopf: Same comment on `true` and `false` being exchanged for `constexpr bool` with meaningful name

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -240,6 +315,14 @@ void HashJoinProcessor::execute() {
 
     // If we have written to the output but have not completed a whole chunk
     if (totalRowsInserted > 0) {
+        // Trim output columns to the actual number of rows inserted
+        for (const auto* namedCol : outDf->cols()) {
+            dispatchColumnVector(namedCol->getColumn(),
+                                 [&](auto* col) {
+                                     auto* colVector = static_cast<decltype(col)>(col);
+                                     colVector->resize(totalRowsInserted);
+                                 });
```
cyrusknopf, line 310: Static cast is redundant, `col` is already typed

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -248,6 +331,14 @@ void HashJoinProcessor::execute() {
...
     if (_leftInput.getPort()->isClosed() && _rightInput.getPort()->isClosed()) {
         if (!_hasWritten) {
+            // Trim output columns to 0 since no rows were produced
+            for (const auto* namedCol : outDf->cols()) {
+                dispatchColumnVector(namedCol->getColumn(),
+                                     [&](auto* col) {
+                                         auto* colVector = static_cast<decltype(col)>(col);
+                                         colVector->resize(0);
+                                     });
+            }
```
cyrusknopf, line 326: Static cast is redundant, `col` is already typed

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -261,15 +352,38 @@ void HashJoinProcessor::execute() {
...
+        const size_t chunkSize = _ctxt->getChunkSize();
+        for (const auto* namedCol : outDf->cols()) {
+            dispatchColumnVector(namedCol->getColumn(),
+                                 [&](auto* col) {
+                                     auto* colVector = static_cast<decltype(col)>(col);
+                                     if (colVector->size() < chunkSize) {
+                                         colVector->resize(chunkSize);
+                                     }
+                                 });
+        }
```
cyrusknopf, line 359: Static cast is redundant, col is already typed.

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -261,15 +352,38 @@ void HashJoinProcessor::execute() {
...
+    }
+
```
cyrusknopf, line 361: Why do we resize to a chunk and then shrink if we didn't use all the space? I think that is wasteful to default initialise the entire column by eagerly resizing

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -287,13 +401,14 @@ void HashJoinProcessor::processLeftStream(size_t& rowsRemaining,
...
                 if (cols[j]->getTag() == _leftJoinKey) {
                     continue;
                 }
 
                 auto* inputColumn = cols[j]->getColumn();
-                auto* outputColumn = outCols[j]->getColumn();
+                auto* outputColumn = outCols[outColIdx]->getColumn();
 
                 dispatchColumnVector(outputColumn, [&](auto* col) {
                     auto* typedOutCol = static_cast<decltype(col)>(outputColumn);
```
cyrusknopf: First static cast is redundant, second should be a dynamic cast IMO

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -354,21 +466,44 @@ void HashJoinProcessor::processLeftStream(size_t& rowsRemaining,
     // from next row on the next cycle. We don't do this if we
     // still have to finish processing a retreived vector of row offsets,
     // or if we have fully processed the input
-    if (!_rowOffsetState.hasRowOffsets() && _leftInputIdx != leftCol->size()) {
+    if (!_rowOffsetState.hasRowOffsets() && _leftInputIdx != leftRowCount) {
```
cyrusknopf: I think this would be easier to follow if you extracted the two conditions into Booleans with meaningful names, something like `const bool finishedProcessingInput = _leftInputIdx != leftRowCount` (assuming I understood  the semantics from the comment correctly.)

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -354,21 +466,44 @@ void HashJoinProcessor::processLeftStream(size_t& rowsRemaining,
...
 
-    for (; _rightInputIdx < rightDf->getLogicalRowCount(); ++_rightInputIdx) {
-        const auto it = findInMap(_leftMap, rightCol, _rightInputIdx);
+    // Resize output columns to chunkSize upfront; we trim after the loop.
+    {
+        const size_t chunkSize = _ctxt->getChunkSize();
+        for (const auto* namedCol : outDf->cols()) {
+            dispatchColumnVector(namedCol->getColumn(),
+                                 [&](auto* col) {
+                                     auto* colVector = static_cast<decltype(col)>(col);
```
cyrusknopf: Redundant cast

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -354,21 +466,44 @@ void HashJoinProcessor::processLeftStream(size_t& rowsRemaining,
...
+                                     if (colVector->size() < chunkSize) {
+                                         colVector->resize(chunkSize);
```
cyrusknopf: Same concern with eager overresizing

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -451,13 +577,14 @@ void HashJoinProcessor::processRightStream(size_t& rowsRemaining,
 
     // if we don't need to finish of copying a row offset vector
     // we can start from the next index
-    if (!_rowOffsetState.hasRowOffsets() && _rightInputIdx != rightCol->size()) {
+    if (!_rowOffsetState.hasRowOffsets() && _rightInputIdx != rightRowCount) {
         _rightInputIdx += 1;
     }
```
cyrusknopf: Same comment about extracting conditions into meaningfully named Boolean variables

query/pipeline/processors/HashJoinProcessor.h
```diff
@@ -1,18 +1,47 @@
...
-namespace db {
-using HashJoinMap = std::unordered_map<NodeID, std::vector<RowOffset>>;
+// Transparent hash for std::string keys, enabling find() by std::string_view
+// without allocating a temporary std::string.
+struct StringHash {
+    using is_transparent = void;
+    size_t operator()(std::string_view sv) const noexcept {
+        return std::hash<std::string_view>{}(sv);
+    }
+};
```
cyrusknopf, line 24: What is `is_transparent` used for? Why is this marked `noexcept`?

query/pipeline/processors/HashJoinProcessor.h
```diff
@@ -1,18 +1,47 @@
...
+// hash/equal so find() accepts both string and string_view without allocation.
+template <typename Key>
+struct HashJoinMapType {
+    using Type = std::unordered_map<Key, std::vector<RowOffset>>;
+};
+
+template <>
+struct HashJoinMapType<std::string_view> {
+    using Type = std::unordered_map<std::string, std::vector<RowOffset>,
+                                    StringHash, std::equal_to<>>;
```
cyrusknopf, line 37: Can you omit `std::equal_to<>`? I thought it was the default argument

query/pipeline/processors/LimitProcessor.cpp
```diff
@@ -35,6 +36,7 @@ LimitProcessor* LimitProcessor::create(PipelineV2* pipeline, size_t limit) {
...
     markAsPrepared();
```
cyrusknopf, line 39: What is this for?

query/pipeline/PipelineBuilder.cpp
```diff
@@ -479,31 +481,67 @@ PipelineBlockOutputInterface& PipelineBuilder::addOrderBy(std::span<const OrderB
...
+                    throw PipelineException("incompatible right key type for string hash join");
+                }
+            });
+        } else {
+            throw PipelineException("unsupported column kind for hash join");
+        }
+    });
 
-    return outInterface;
+    return *result;
```
cyrusknopf, line 607: I think we should `bioassert(result)` before dereferencing it here.

query/pipeline/PipelineBuilder.cpp
```diff
@@ -479,31 +481,67 @@ PipelineBlockOutputInterface& PipelineBuilder::addOrderBy(std::span<const OrderB
...
+
+        // For std::string/std::string_view and their symetries,
+        // we have a different type for the left and the right key
+        // but they are compatible in the sense that they are "string-like"
+        if constexpr (std::is_same_v<LeftKey, NodeID> ||
+                     std::is_same_v<LeftKey, EdgeID> ||
+                     std::is_same_v<LeftKey, int64_t> ||
+                     std::is_same_v<LeftKey, uint64_t> ||
+                     std::is_same_v<LeftKey, double> ||
+                     std::is_same_v<LeftKey, CustomBool>) {
```
cyrusknopf: This should be defined elsewhere I think its way too big to just be checked used inline

query/common/CardinalityEstimation.cpp
ghost: @rjb32 ?

query/pipeline/PipelineBuilder.cpp
```diff
@@ -479,31 +481,67 @@ PipelineBlockOutputInterface& PipelineBuilder::addOrderBy(std::span<const OrderB
...
+            dispatchColumnVector(rightKeyCol, [&](auto* rightTypedCol) {
+                using RightValueType = typename std::decay_t<decltype(*rightTypedCol)>::ValueType;
+                using RightKey = TypeUtils::unwrap_optional_t<RightValueType>;
+
+                if constexpr (StringLike<RightKey>) {
+                    auto* join = HashJoinProcessor<LeftKey, RightKey>::create(_pipeline, leftJoinKey, rightJoinKey);
+                    result = &setupJoin(join);
+                } else {
+                    throw PipelineException("incompatible right key type for string hash join");
+                }
```
cyrusknopf: I'm assuming that the `if` case is when both sides are equal, so we only provide one type template argument which is `LeftKey` where `LeftKey = RightKey`. And in the `else` case it is some permutation of {string, string_view}, and both types are passed. But it is not really obvious, the comment above and the condition `std::is_same_v<LeftKey, X>` doesn't really communicate that in that case both types are equal. Some comments on each branch would be useful, and maybe `using JoinType = LeftKey` to signal that the type is uniform.

query/pipeline/EntityOutputStream.h
```diff
@@ -70,31 +76,40 @@ class EntityOutputStream {
...
-    auto visit(const auto& visitor) const {
-        return std::visit(StreamVisitor<decltype(visitor)> {visitor}, _stream);
+    template <typename Visitor, typename... Args>
+    auto visit(Visitor&& visitor, Args&&... args) const {
+        return std::visit(
+            StreamVisitor<std::decay_t<Visitor>, std::decay_t<Args>...> {
+                std::forward<Visitor>(visitor),
+                std::tuple<std::decay_t<Args>...> {std::forward<Args>(args)...}},
+            _stream);
     }
```
cyrusknopf: I think this needs commenting, its not clear how this should be used, or what `args` is/could be. Also `T` might not be the best typename for this, I think it is a functor/callable? maybe a more verbose typename

query/plan/nodes/JoinNode.h
```diff
@@ -15,36 +15,36 @@ enum class JoinType : uint8_t {
...
+        _joinKeyVar2(joinKeyVar2),
         _dependencyVarDecl(dependencyTag),
         _type(type)
     {
     }
 
-    const VarDecl* getLeftVarDecl() const { return _leftJoinKeyVar; }
-    const VarDecl* getRightVarDecl() const { return _rightJoinKeyVar; }
+    const VarDecl* getFirstJoinKey() const { return _joinKeyVar1; }
+    const VarDecl* getSecondJoinKey() const { return _joinKeyVar2; }
```
cyrusknopf: Do we want the naming unified to either rename members to align with method names: `_fstJoinKey, _sndJoinKey`; or rename method names to align with member names: `getJoinKey1, getJoinKey2`.

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "Overloaded.h"
+
+using namespace db;
+
+ColumnTag TranslateJoinHelpers::getStreamTag(const PipelineOutputInterface* input) {
+    const auto visitor = Overloaded {
+        [](const EntityOutputStream::NodeStream& s) { return s._nodeIDsTag; },
+        [](const EntityOutputStream::EdgeStream& s) { return s._edgeIDsTag; }};
+    return input->getStream().visit(visitor);
+}
```
cyrusknopf, line 18: Is there never a case where there is an Edgestream, but we want to continue with the target node tag instead of edge ID tag?

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+                                                     ColumnTag joinTag) {
+    const auto visitor = Overloaded {
+        [](const EntityOutputStream::NodeStream&, ColumnTag tag) {
+            return EntityOutputStream::createNodeStream(tag);
+        },
+        [](const EntityOutputStream::EdgeStream& s, ColumnTag tag) {
+            return EntityOutputStream::createEdgeStream(tag, s._otherIDsTag, s._edgeTypesTag);
+        }};
+    return stream.visit(visitor, joinTag);
+}
```
cyrusknopf, line 33: I think this needs some comments, it looks like you're replacing the edge id tag but keeping the edge src and target tag?

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+    ColumnTag leftJoinTag;
+    ColumnTag rightJoinTag;
+
+    switch (node->getJoinType()) {
+        // In the common ancestor case the join var decls will exist in
+        // the decl to column map so we can just find them
+        case JoinType::COMMON_ANCESTOR: {
+            // Both the first and second join keys are the same in this case
+            leftJoinTag = declMap.find(node->getFirstJoinKey())->second;
+            rightJoinTag = declMap.find(node->getFirstJoinKey())->second;
```
cyrusknopf: Should check if `find` returns `end()`. Also should we assert that both are equal? as the comment suggests?

query/pipeline/processors/HashJoinProcessor.cpp
ghost: I think this code is pretty safe if you follow how all the templates are set. Given that this falls in the hot path of the join it is preferable not to have any runtime type checks.
```c++
    {
        auto* leftKeyCol = leftDf->getColumn(_leftJoinKey)->getColumn();
        const auto leftKind = leftKeyCol->getKind();
        _isLeftOptionalKey = (leftKind == ColumnVector<std::optional<LeftKey>>::staticKind());
    }

    {
        auto* rightKeyCol = rightDf->getColumn(_rightJoinKey)->getColumn();
        const auto rightKind = rightKeyCol->getKind();
        _isRightOptionalKey = (rightKind == ColumnVector<std::optional<RightKey>>::staticKind());
    }
```
```c++
        if (_isRightOptionalKey) {
            processRightStream<true>(rowsRemaining, totalRowsInserted);
        } else {
            processRightStream<false>(rowsRemaining, totalRowsInserted);
        }
```
```c++
        const auto rightKey = extractKeyView<RightKey, IsOptional>(rightCol, _rightInputIdx);
```

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+                    rightJoinTag = firstKeyIt->second;
+                }
+            }
+            break;
+        }
+        case JoinType::DIAMOND: {
+            throw PlannerException("Common Successor Joins With Common Ancestor Unsupported");
+        }
+    }
+    return std::pair<ColumnTag, ColumnTag>(leftJoinTag, rightJoinTag);
```
cyrusknopf: Can we do this?
```suggestion
    return {firstJoinTag, secondJoinTag}; 
```

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+        }};
+    return stream.visit(visitor, joinTag);
+}
+
+std::pair<ColumnTag, ColumnTag> TranslateJoinHelpers::getJoinKeyTags(const JoinNode* node,
+                                                                     const PipelineOutputInterface* lhs,
+                                                                     const PipelineOutputInterface* rhs,
+                                                                     const DeclToTagMap& declMap) {
+    ColumnTag leftJoinTag;
+    ColumnTag rightJoinTag;
```
cyrusknopf, line 40: "left" and "right" naming is outdated compared to "first" and "second" join tags that are used by the methods

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+            rightJoinTag = getStreamTag(rhs);
+            break;
+        }
+        case JoinType::PREDICATE: {
+            // Check which input the dependency VarDecl is in this is should
+            // be the fully expanded branch ( we will refer to this as the dependency branch)
+            bool rhsIsDependencyBranch = true;
+
+            const auto firstKeyIt = declMap.find(node->getFirstJoinKey());
+            const auto secondKeyIt = declMap.find(node->getSecondJoinKey());
```
cyrusknopf, line 72: Should check `find` does not return `end()`

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+//           - When we have a predicate join, there are two cases of this as well:
+//                       - Property Predicate Join - in this case we just need to propogate
+//                         the undexpanded branch.
+//                       - Entity ID Predicate Join - in this case the unexpanded branch's
+//                         stream would contain the temporary VarDecl (of the Entity dep)
+//                         whose values would be written into the new join column. So the
+//                         stream we propagate has to be merged with join tag.
+//
+//  In all other cases (regular Common Ancestor case) we don't care about the stream
+//  as the join happens as the last processor in the pipeline
```
cyrusknopf, line 143: Nice to have: example CYPHER queries which illustrate each example so its easier to visualise

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+    if (node->getDependencyVarDecl()) {
+        const auto it = declMap.find(node->getDependencyVarDecl());
+        if (it == declMap.end()) {
+            throw PlannerException("Join Dependency VarDecl Not Found");
+        }
+        if (!lhs->getDataframe()->hasColumn(it->second)) {
+            // the decl column wouldn't contain the first join key only
+            // in the EntityID predicate join case
+            if (!declMap.contains(node->getFirstJoinKey())) {
+                stream = mergeStream(lhs->getStream(), joinTag);
```
cyrusknopf, line 161: Yo already have `lhs->getStream()` saved as the `stream` variable

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+            } else {
+                stream = lhs->getStream();
```
cyrusknopf, line 163: Isn't this redundant because of the first line in this function?

query/plan/PipelineGenerator.cpp
```diff
@@ -856,62 +859,20 @@ PipelineOutputInterface* PipelineGenerator::translateJoinNode(JoinNode* node) {
...
-        }
-
-        // The input that doesn't have to dependencyVarDecl is the one we wish to stream
-        if (!lhs->getDataframe()->hasColumn(it->second)) {
-            streamedTag = lhs->getStream().visit(visitor);
-        } else {
-            streamedTag = rhs->getStream().visit(visitor);
-        }
-    }
+    ColumnTag joinTag = outputIf.getDataframe()->cols().back()->getTag();
```
cyrusknopf: Should split this up IMO. What if the dataframe has no columns?

query/plan/PipelineGenerator.cpp
```diff
@@ -856,62 +859,20 @@ PipelineOutputInterface* PipelineGenerator::translateJoinNode(JoinNode* node) {
...
+    outputIf.setStream(TranslateJoinHelpers::resolveStream(node, lhs, rhs, joinTag, _declToColumn));
```
cyrusknopf: I think it's clerer if you save the result of `resolveStream` to a local var so its easier to reason about what it returns

query/plan/PipelineGenerator.cpp
```diff
@@ -856,62 +859,20 @@ PipelineOutputInterface* PipelineGenerator::translateJoinNode(JoinNode* node) {
...
```
cyrusknopf, line 884:
```suggestion
```

query/plan/PlanGraph.cpp
```diff
@@ -47,3 +47,9 @@ void PlanGraph::removeIsolatedNodes() {
 
     _nodes = std::move(newNodes);
 }
+
+void PlanGraph::removeNode(PlanGraphNode* node) {
+    std::erase_if(_nodes, [node](const std::unique_ptr<PlanGraphNode>& n) {
+        return n.get() == node;
+    });
+}
```
cyrusknopf: Is this intending to remove a single node, or all occurences of `node`? From the name it seems like removing a single node, but `erase_if` could remove multiple instances of `node`

query/plan/PlanGraph.h
```diff
@@ -66,6 +66,7 @@ class PlanGraph {
     }
 
     void removeIsolatedNodes();
+    void removeNode(PlanGraphNode* node);
```
cyrusknopf: PlanGraphNode can be const?

query/plan/PlanGraphGenerator.cpp
```diff
@@ -198,7 +202,21 @@ void PlanGraphGenerator::generateSinglePartQuery(const SinglePartQuery* query) {
...
+            if (proj && proj->hasLimit()) {
+                const Expr* limitExpr = proj->getLimit()->getExpr();
+                if (limitExpr->getKind() == Expr::Kind::LITERAL) {
+                    const auto* lit = static_cast<const LiteralExpr*>(limitExpr)->getLiteral();
+                    if (lit->getKind() == Literal::Kind::INTEGER) {
+                        readGenerator.setQueryLimit(static_cast<const IntegerLiteral*>(lit)->getValue());
+                    }
+                }
+            }
+        }
```
cyrusknopf, line 219: What is this for?

query/plan/ReadStmtGenerator.cpp
```diff
@@ -229,13 +235,19 @@ void ReadStmtGenerator::generatePatternElement(const PatternElement* element) {
 
     VarNode* currentNode = generatePatternElementOrigin(origin);
 
+    _edgesInPattern.clear();
```
cyrusknopf, line 240: Why

query/plan/ReadStmtGenerator.cpp
```diff
@@ -229,13 +235,19 @@ void ReadStmtGenerator::generatePatternElement(const PatternElement* element) {
...
     const auto& chain = element->getElementChain();
     for (const auto& [edge, node] : chain) {
         const EdgePattern* e = dynamic_cast<const EdgePattern*>(edge);
         if (!edge) {
             throwError("Pattern element edge must be an edge pattern", element);
         }
 
+        if (e->getDecl() && !_edgesInPattern.insert(e->getDecl()).second) {
+            throwError("Re-using the same edge variable in a single pattern is not supported", edge);
+        }
```
cyrusknopf: I think  we should break this up into  some local variables, seems like way too many operations in the condition of an if

query/plan/ReadStmtGenerator.cpp
```diff
@@ -564,16 +574,58 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+        bool beenPlaced = std::find(filterPreds.begin(), filterPreds.end(), pred.get()) != filterPreds.end();
+        if (!vhjPlaced && !beenPlaced) {
+            filterNode->addPredicate(pred.get());
+            pred->setFilterNode(filterNode);
+        }
+    }
+}
+
+bool ReadStmtGenerator::shouldPlaceValueHashJoin(VarNode* localVar, PlanGraphNode* remoteNode) {
+    auto* rightVar = dynamic_cast<VarNode*>(remoteNode);
```
cyrusknopf: Can check that the dynamic cast succeeded

query/plan/ReadStmtGenerator.cpp
```diff
@@ -564,16 +574,58 @@ void ReadStmtGenerator::placePredicateJoins() {
...
 
-        // Step 3: Place the constraint
+        // Step 3: For single-var and common sucessor predicates
+        // all deps return early (SameVar/BackwardPath) and never add the predicate,
+        // so add it directly here if not already present.
         FilterNode* filterNode = _variables->getNodeFilter(var);
-        filterNode->addPredicate(pred.get());
-        pred->setFilterNode(filterNode);
+        const auto& filterPreds = filterNode->getPredicates();
+        bool beenPlaced = std::find(filterPreds.begin(), filterPreds.end(), pred.get()) != filterPreds.end();
```
cyrusknopf: Can be const. Also can we use `std::ranges::find`?

query/plan/ReadStmtGenerator.cpp
```diff
@@ -564,16 +574,58 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+            leftLabels = nf->getLabelConstraints();
+        }
+    }
+
+    CardinalityEstimation estimation(_graphView);
+
+    if (rightVar) {
+        FilterNode* remoteFilter = _variables->getNodeFilter(rightVar);
+        if (remoteFilter) {
+            if (auto* nf = remoteFilter->asNodeFilter()) {
```
cyrusknopf: `asNodeFilter` is static cast, so this if check does nothing. Should probably explicitly dynamic cast

query/plan/ReadStmtGenerator.cpp
```diff
@@ -767,17 +821,40 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
-            // If nodes are on two different islands
-            // Cartesian product. This can be optimized in the future into a ValueHashJoin
-            CartesianProductNode* join = _tree->insertBefore<CartesianProductNode>(filter);
-            PlanGraphNode* depBranchTip = _topology->getBranchTip(dependency);
-            depBranchTip->connectOut(join);
-            return;
+            // If nodes are on two different islands depending on cardinality we can choose
+            // to do a Value Hash Join or a Cartesian Product + Filter
+
+            const auto* predExpr = pred->getExpr();
```
cyrusknopf: Probably shouldnt be auto

query/plan/ReadStmtGenerator.cpp
```diff
@@ -767,17 +821,40 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
+            const bool isEqualityPred = predExpr->getKind() == Expr::Kind::BINARY && static_cast<const BinaryExpr*>(predExpr)->getOperator() == BinaryOperator::Equal;
```
cyrusknopf: Column line limit. Also I think these things should be broken up quite a lot. a static cast + dereference is a bit much

query/plan/ReadStmtGenerator.cpp
```diff
@@ -767,17 +821,40 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
+
+            const ExprDependencies& deps = pred->getDependencies();
+            const auto& varDeps = deps.getVarDeps();
+
+            bool canPlaceValueHashJoin = isEqualityPred && deps.getFuncDeps().empty() && varDeps.size() == 2;
```
cyrusknopf: Can be const. Also I think it would be easier to read if each clause of the && is its own meaninfully named condition

query/plan/ReadStmtGenerator.cpp
```diff
@@ -767,17 +821,40 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
+                                                               binExpr->getRHS()->getExprVarDecl(),
+                                                               JoinType::PREDICATE);
+                PlanGraphNode* depBranchTip = _topology->getBranchTip(dependency);
+                depBranchTip->connectOut(join);
+                return true;
+            } else {
+                CartesianProductNode* join = _tree->insertBefore<CartesianProductNode>(filter);
+                PlanGraphNode* depBranchTip = _topology->getBranchTip(dependency);
+                depBranchTip->connectOut(join);
+                return false;
```
cyrusknopf: Don't need `else` after early return

query/plan/ReadStmtGenerator.cpp
```diff
@@ -767,17 +821,40 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
+
+            bool canPlaceValueHashJoin = isEqualityPred && deps.getFuncDeps().empty() && varDeps.size() == 2;
+
+            if (canPlaceValueHashJoin && shouldPlaceValueHashJoin(node, dependency)) {
+                const auto* binExpr = static_cast<const BinaryExpr*>(predExpr);
+                JoinNode* join = _tree->insertBefore<JoinNode>(filter,
+                                                               binExpr->getLHS()->getExprVarDecl(),
+                                                               binExpr->getRHS()->getExprVarDecl(),
+                                                               binExpr->getRHS()->getExprVarDecl(),
+                                                               JoinType::PREDICATE);
```
cyrusknopf: I think the arguments should be dereferenced into local vars then passed

query/plan/ReadStmtGenerator.cpp
```diff
@@ -767,17 +821,40 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
+            const auto* predExpr = pred->getExpr();
+            const bool isEqualityPred = predExpr->getKind() == Expr::Kind::BINARY && static_cast<const BinaryExpr*>(predExpr)->getOperator() == BinaryOperator::Equal;
+
+            const ExprDependencies& deps = pred->getDependencies();
+            const auto& varDeps = deps.getVarDeps();
+
+            bool canPlaceValueHashJoin = isEqualityPred && deps.getFuncDeps().empty() && varDeps.size() == 2;
+
+            if (canPlaceValueHashJoin && shouldPlaceValueHashJoin(node, dependency)) {
+                const auto* binExpr = static_cast<const BinaryExpr*>(predExpr);
```
cyrusknopf: I think dynamic cast + check here is safer since its not a hotpath

query/plan/ReadStmtGenerator.h
```diff
@@ -73,18 +78,24 @@ class ReadStmtGenerator {
...
     PlanGraph* _tree {nullptr};
     PlanGraphVariables* _variables {nullptr};
     std::unique_ptr<PlanGraphTopology> _topology;
     bool _isStandaloneCall {false};
+    size_t _queryLimit {0};
+    std::unordered_set<const VarDecl*> _edgesInPattern;
 
     void generateDependency(PlanGraphNode* producer, Expr* rawExpr);
 
+    bool shouldPlaceValueHashJoin(VarNode* localVar, PlanGraphNode* remoteNode);
```
cyrusknopf, line 98: Can this method be const?

scripts/bench_vhj.sh
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env bash
+#
```
cyrusknopf, line 2: Regarding this entire directory: do we want this committed? Perhaps the `turing-db/profiling` repo is a better place

db/TuringDB.h
```diff
@@ -33,40 +34,38 @@ class TuringDB {
...
+    SystemManager& getSystemManager() { return *_systemManager; }
+    JobSystem& getJobSystem() { return *_jobSystem; }
+    const ProcedureManager* getProcedures() const { return _procedures.get(); }
     ExtensionManager* getExtensions() { return _extensions.get(); }
+
     void stop();
 
 private:
     const TuringConfig* _config {nullptr};
+    QueryConfig _defaultQueryConfig;
```
cyrusknopf, line 68: Why is this member called "default"? Surely it should just be "queryConfig", which is then default initialised.

db/TuringDB.h
```diff
@@ -33,40 +34,38 @@ class TuringDB {
...
-    }
-
-    JobSystem& getJobSystem() {
-        return *_jobSystem;
-    }
-
-    const ProcedureManager* getProcedures() const {
-        return _procedures.get();
-    }
+    const QueryConfig& getDefaultQueryConfig() const { return _defaultQueryConfig; }
```
cyrusknopf, line 57: Same comments on the name of this method as on the name of the member

query/pipeline/processors/HashJoinProcessor.cpp
cyrusknopf: It is already not safe because
`_isXOptionalKey == false` for `XKeyCol == ColumnConst<X>`
which would result in attempting to static cast a ColumnConst into ColumnVector.

And that is just currently. If the implementation surrounding this method changes, the static cast is even more susceptible to failing. Also to what extent is this on the hot path? Is it executed per row? or per column? If it is per column then the check is negligible

query/plan/TranslateJoinHelpers.cpp
ghost: There is a comment in PipelineGenerator explaining that left and right join tag correspond to wether the join will be searched for in the lhs or rhs input.

query/pipeline/processors/HashJoinProcessor.cpp
ghost: In the previous case we did one resize but this would require iterating over the inputs and calculating the number of new rows.

query/pipeline/processors/HashJoinProcessor.h
ghost: https://en.cppreference.com/w/cpp/named_req/UnorderedAssociativeContainer.html

`is_transparent` is part of how we indicate that the associative container can hash values other than it's key type.

query/pipeline/processors/HashJoinProcessor.h
ghost: default argument is std::equal_to<std::string>. std::equal_to<> is a transparent comparison function that let's us compare different types than the key type.

query/plan/nodes/JoinNode.h
```diff
@@ -15,36 +15,36 @@ enum class JoinType : uint8_t {
...
+    const VarDecl* getFirstJoinKey() const { return _fstJoinKeyVar; }
+    const VarDecl* getSecondJoinKey() const { return _sndJoinKeyVar2; }
     const VarDecl* getDependencyVarDecl() const { return _dependencyVarDecl; }
     JoinType getJoinType() const { return _type; };
 
 private:
-    const VarDecl* _leftJoinKeyVar {nullptr};
-    const VarDecl* _rightJoinKeyVar {nullptr};
+    const VarDecl* _fstJoinKeyVar {nullptr};
+    const VarDecl* _sndJoinKeyVar2 {nullptr};
```
cyrusknopf: Should be `_sndJoinKeyVar` I'm assuming?

db/TuringDB.cpp
```diff
@@ -134,29 +134,33 @@ void TuringDB::init() {
     SystemEventHandler::setOnStop([this] {
         _config->getOnStopRequest()();
     });
+
+    // Configure default query config from environment
+    const char* vhjEnv = getenv("TURING_VALUE_HASH_JOIN");
+    if (vhjEnv) {
+        const bool enabled = std::string(vhjEnv) != "0";
```
rjb32: Instead of recreating a string it's more idiomatic to use strcmp

query/plan/ReadStmtGenerator.cpp
```diff
@@ -971,17 +883,64 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
         }
 
         case PlanGraphTopology::PathToDependency::NoPath: {
-            // If nodes are on two different islands
-            // Cartesian product. This can be optimized in the future into a ValueHashJoin
+            // If nodes are on two different islands depending on cardinality we can choose
+            // to do a Value Hash Join or a Cartesian Product + Filter
+
+            const BinaryExpr* binExpr = dynamic_cast<const BinaryExpr*>(pred->getExpr());
+            if(binExpr) {
```
cyrusknopf: `if (binExpr)`

query/plan/ReadStmtGenerator.h
```diff
@@ -97,7 +95,7 @@ class ReadStmtGenerator {
 
     void generateDependency(PlanGraphNode* producer, Expr* rawExpr);
 
-    bool tryPlaceValueHashJoin(Predicate* pred, ExprDependencies& deps, VarNode* var);
+    const bool shouldPlaceValueHashJoin(VarNode* localVar, PlanGraphNode* remoteNode);
```
cyrusknopf: No need for const?

db/TuringDB.h
```diff
@@ -33,40 +34,38 @@ class TuringDB {
...
+    SystemManager& getSystemManager() { return *_systemManager; }
+    JobSystem& getJobSystem() { return *_jobSystem; }
+    const ProcedureManager* getProcedures() const { return _procedures.get(); }
     ExtensionManager* getExtensions() { return _extensions.get(); }
+
     void stop();
 
 private:
     const TuringConfig* _config {nullptr};
+    QueryConfig _queryConfig;
```
rjb32: This member does not seem to be used anymore sine queryConfig is passed as a pointer everytime

fuzz/fuzz_query_engine.cpp
```diff
@@ -92,7 +94,7 @@ static int fuzzOne(const char* data, size_t size) {
     }
 
     // Plan
-    db::PlanGraphGenerator planGen(ast, view);
+    db::PlanGraphGenerator planGen(ast, view, &(queryConfig.getPlanGenConfig()));
```
rjb32: This is ugly to wrap this in parenthesis.

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,188 @@
...
+//
+//  In all other cases (regular Common Ancestor case) we don't care about the stream
+//  as the join happens as the last processor in the pipeline
+EntityOutputStream TranslateJoinHelpers::resolveStream(const JoinNode* node,
+                                                       const PipelineOutputInterface* lhs,
+                                                       const PipelineOutputInterface* rhs,
+                                                       ColumnTag joinTag,
+                                                       DeclToTagMap& declMap) {
+    // Default value is to push random stream
+    EntityOutputStream stream = lhs->getStream();
```
cyrusknopf, line 150: Did we talk about initialising this as invalid?

query/common/CardinalityEstimation.cpp
rjb32: Let's keep it for now

query/pipeline/processors/HashJoinProcessor.cpp
rjb32: Yes we could just use a normal return type notation here

query/plan/ReadStmtGenerator.cpp
```diff
@@ -620,16 +632,62 @@ void ReadStmtGenerator::placePredicateJoins() {
...
+    FilterNode* localFilter = _variables->getNodeFilter(localVar);
+    if (localFilter) {
+        if (auto* nf = localFilter->asNodeFilter()) {
+            leftLabels = nf->getLabelConstraints();
+        }
+    }
+
+    CardinalityEstimation estimation(_graphView);
+
+    auto* rightVar = dynamic_cast<VarNode*>(remoteNode);
```
rjb32: rightVar can be const

query/plan/ReadStmtGenerator.cpp
```diff
@@ -823,17 +883,64 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
 
         case PlanGraphTopology::PathToDependency::NoPath: {
-            // If nodes are on two different islands
-            // Cartesian product. This can be optimized in the future into a ValueHashJoin
+            // If nodes are on two different islands depending on cardinality we can choose
+            // to do a Value Hash Join or a Cartesian Product + Filter
+
+            const BinaryExpr* binExpr = dynamic_cast<const BinaryExpr*>(pred->getExpr());
+            if(binExpr) {
+
```
rjb32: Blank line

query/plan/ReadStmtGenerator.cpp
```diff
@@ -823,17 +883,64 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
+                const ExprDependencies& deps = pred->getDependencies();
+                const auto& funcDeps = deps.getFuncDeps();
+                const bool noFunctionDependencies = funcDeps.empty();
+
+                const bool canPlaceValueHashJoin = isEqualityPred
+                    && isValidLhs
+                    && isValidRhs
+                    && noFunctionDependencies;
+
```
rjb32: Blank line

query/plan/ReadStmtGenerator.cpp
```diff
@@ -823,17 +883,64 @@ void ReadStmtGenerator::insertDataFlowNode(VarNode* node, PlanGraphNode* depende
...
             PlanGraphNode* depBranchTip = _topology->getBranchTip(dependency);
             depBranchTip->connectOut(join);
-            return;
+            return false;
         }
 
         case PlanGraphTopology::PathToDependency::NoPath: {
-            // If nodes are on two different islands
-            // Cartesian product. This can be optimized in the future into a ValueHashJoin
+            // If nodes are on two different islands depending on cardinality we can choose
```
rjb32: This is very long this should be inside a function

query/plan/TranslateJoinHelpers.h
```diff
@@ -0,0 +1,31 @@
...
+#include "dataframe/ColumnTag.h"
+
+
+namespace db {
+class JoinNode;
+class VarDecl;
+class PipelineOutputInterface;
+
+class TranslateJoinHelpers {
+public:
```
rjb32, line 16: Put constructor and destructor as delete to show this can not be constructed

query/common/CardinalityEstimation.cpp
```diff
@@ -0,0 +1,62 @@
...
+    return reader.getNodeCountMatchingLabelset(labelset.handle());
+}
+
+bool CardinalityEstimation::shouldPreferCartesian(const LabelSet& left, const LabelSet& right, size_t queryLimit) const {
+    const size_t leftCount = estimateNodeCount(left);
+    const size_t rightCount = estimateNodeCount(right);
+
+    // Small cartesian product is cheap enough without a hash join
+    if (leftCount * rightCount < SMALL_CARTESIAN_THRESHOLD) {
```
rjb32: Potential arithmetic overflow here if large size_t values

query/common/CardinalityEstimation.cpp
```diff
@@ -0,0 +1,62 @@
...
+        const size_t largerSide = std::max(leftCount, rightCount);
+        if (largerSide > queryLimit * VHJ_LIMIT_RATIO) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+bool CardinalityEstimation::shouldPreferCartesian(const LabelSet& left, const int rightCardinality, size_t queryLimit) const {
```
rjb32: rightCardinality should use size_t

query/plan/TranslateJoinHelpers.h
```diff
@@ -0,0 +1,31 @@
+#include <unordered_map>
```
rjb32, line 3: Missing #pragma once

query/plan/TranslateJoinHelpers.cpp
```diff
@@ -0,0 +1,188 @@
...
+                                                                     const DeclToTagMap& declMap) {
+    ColumnTag leftJoinTag;
+    ColumnTag rightJoinTag;
+
+    switch (node->getJoinType()) {
+        // In the common ancestor case the join var decls will exist in
+        // the decl to column map so we can just find them
+        case JoinType::COMMON_ANCESTOR: {
+            // Both the first and second join keys are the same in this case
+            auto leftJoinIt = declMap.find(node->getFirstJoinKey());
```
rjb32: Save node->getFirstJoinKey() it is called twice

query/pipeline/processors/HashJoinProcessor.h
rjb32: hasRowOffsets and numRemainingOffsets can be const

query/plan/TranslateJoinHelpers.cpp
ghost: That was breaking some assumptions we had, so I've left it like this.

db/TuringDB.h
ghost: It is used - the pointer passed around in `TuringDB::query` comes from here. DBServerProcessor, TuringShell etc use it.

PR #482, opened 2026-03-05
Fix flushRightStream column offset and reset _hasWritten

rjb32:
`flushRightStream` used `_leftRowLen + j` as the output column index, but j iterates over all right columns including skipped ones (columns shared with the left dataframe). 

This is inconsistent with processRightStream which correctly uses a separate columnOffset variable that only increments for non-skipped columns.

On main today the bug is latent: 
The only overlap between left and right dataframes is the join key column itself (COMMON_SUCCESSOR joins share the entity ID tag), and that column is already excluded by the join-key skip condition, so j and a correct columnOffset never diverge. 
However, value hash joins (PREDICATE type) use different tags for left and right join keys, making it possible for other columns to overlap between the two sides. This would cause flushRightStream to write into wrong output column positions.

Also reset _hasWritten in reset() so the empty-input write guard at the end of execute() works correctly if the processor is reused across pipeline cycles.

### DIFF
--- a/query/pipeline/processors/HashJoinProcessor.cpp
+++ b/query/pipeline/processors/HashJoinProcessor.cpp
@@ -95,6 +95,7 @@ void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
     markAsPrepared();
 }
 void HashJoinProcessor::reset() {
+    _hasWritten = false;
     markAsReset();
 }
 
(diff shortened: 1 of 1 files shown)


rjb32:
@suhassn-turing-bio A small potential bug fix that Claude suggested

query/pipeline/processors/HashJoinProcessor.cpp
```diff
@@ -95,6 +95,7 @@ void HashJoinProcessor::prepare(ExecutionContext* ctxt) {
     markAsPrepared();
 }
 void HashJoinProcessor::reset() {
+    _hasWritten = false;
```
ghost, line 98: Will we ever see the hash join being reset as it needs to maintain its state through all the processor cycles. Adding `_hasWritten = false` but maybe we can add a comment explaining it so that we don't get confused in the future.

ghost:
Other than my comment the bug is valid! Weirdly I thought I remembered fixing this not too long ago haha.

query/pipeline/processors/HashJoinProcessor.cpp
rjb32: Yes true indeed

PR #487, opened 2026-03-05
Support `FuncEvalNode`, with `labels()` function

cyrusknopf:
Adds function evaluation to `ExprProgram`.

- Non-aggregate functions such as `labels()`, `toInteger()`, etc. are now evaluated by an `ExprProgram`, most often in an `ExprEvalNode`, but also possibly part of a `FilterNode`.
- This allows for operations such as `MATCH (n) WHERE n.age = toInteger("32") RETURN n.age + toFloat("10.2"), labels(n)`

### DIFF
--- a/test/query/queries/QueriesTest.cpp
+++ b/test/query/queries/QueriesTest.cpp
@@ -3,12 +3,15 @@
 #include <functional>
 #include <string_view>
 
+#include <range/v3/view/drop.hpp>
+
 #include "TuringDB.h"
 #include "Graph.h"
 #include "SimpleGraph.h"
 #include "SystemManager.h"
 #include "columns/ColumnIDs.h"
 #include "columns/ColumnOptVector.h"
+#include "metadata/LabelSetHandle.h"
 #include "metadata/PropertyType.h"
 #include "ID.h"
 #include "versioning/CommitBuilder.h"
@@ -23,6 +26,9 @@
 #include "versioning/VersionController.h"
 
 using namespace turing::test;
+
+namespace rg = ranges;
+namespace rv = rg::views;
 
 class QueriesTest : public TuringTest {
 public:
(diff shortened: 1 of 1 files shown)


query/pipeline/processors/EvalFunction.cpp
```diff
@@ -0,0 +1,87 @@
...
+    using Types = TypeRestrictions<Op>;
+    Eval<Op> fn {res, {}};
+    using Dispatcher = ColumnSingleDispatcher<typename Types::Allowed,
+                                              Eval<Op>,
+                                              typename Types::Excluded>;
+    Dispatcher::dispatch(arg, fn);
+}
+
+template void EvalFunction::eval<OP_FUNC_LABELS>(Column* res, const Column* arg, GraphView view);
+
```
rjb32, line 84: Remove extra blank space

test/query/queries/QueriesTest.cpp
```diff
@@ -1409,36 +1409,6 @@ TEST_F(QueriesTest, db_history) {
     actualRows.clear();
 }
```
rjb32, line 1416: We should add a unit test for the labels function

query/pipeline/processors/EvalFunction.cpp
cyrusknopf: I added the line to separate the functions which need a `GraphView` as parameter versus those which do not.

rjb32: Ah ok fair enough

test/query/queries/QueriesTest.cpp
cyrusknopf: Added a `labelsFunction` unit test in `QueriesTest.cpp`

PR #491, opened 2026-03-06
Feature: variable length paths (n)-->+(m)

LucLabarriere:
image

- 8 unit tests
  - `MATCH (n:Person)-[e]->+(m:Person) RETURN n.name, e, m.name`
  - `MATCH (n:Person)<-[e]-+(m:Person) RETURN n.name, e, m.name`
  - `MATCH (n:Person)-[e]->*(m:Person) RETURN n.name, e, m.name`
  - `MATCH (n:Person)<-[e]-*(m:Person) RETURN n.name, e, m.name`
  - `MATCH (n:Person)-[e]-*(m) RETURN n.name, e, m.name SKIP 50 LIMIT 5`
  - `MATCH (n:Person)-[e]->{2,4}(m:Interest) RETURN n.name, e, m.name`
  - `MATCH (n:Person)-[e]->{2,4}(m:Interest)-->(p:Person) RETURN n.name, e, m.name, p.name`
  - `MATCH (i:Supernatural)-->(n:Person)-[e]->*(m:Person) RETURN i.name, n.name, e, m.name`


query/AST/QuantifiedPath.h
```diff
@@ -0,0 +1,34 @@
...
+    void setLhs(int64_t lhs) { _lhs = lhs; }
+    void setRhs(int64_t rhs) { _rhs = rhs; }
+
+    int64_t getLhs() const { return _lhs; }
+    int64_t getRhs() const { return _rhs; }
+
+    bool isRhsUnbounded() const { return _rhs == UNBOUNDED; }
+
+private:
+    int64_t _lhs {0};
```
rjb32, line 27: Why using a signed type instead of size_t?

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+    _bfsIndices = _mem->alloc<ColumnIndices>();
+
+    switch (_dir) {
+        case PathExplorationDir::Forward: {
+            auto* bfsWriter = new GetOutEdgesChunkWriter(view, _bfsSources);
+            bfsWriter->setIndices(_bfsIndices);
+            bfsWriter->setEdgeIDs(_bfsEdges);
+            bfsWriter->setTgtIDs(_bfsIntermediates);
+            _bfsWriter = std::unique_ptr<Iterator>(bfsWriter);
+        } break;
```
rjb32: Style: put the break on the next line

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+            } break;
+        }
+
+        _depthNeedsSetup = false;
+    }
+
+    // Step 3. Fill one chunk of edges from the current depth set of sources.
+    switch (_dir) {
+        case PathExplorationDir::Forward: {
+            static_cast<GetOutEdgesChunkWriter*>(_bfsWriter.get())->fill(remaining);
```
rjb32: Please put the casted chunk writers in their own local variable

query/AST/QuantifiedPath.h
LucLabarriere: Good point, left over I thought we could use negative numbers but no

LucLabarriere: Actually, I let it be int64_t so I can catch negative number errors in the analyzer (digits in the parser are int64_t by default), but use uint64_t in the processor

query/outputs/JsonEncoder.h
```diff
@@ -111,6 +111,37 @@ class ChunkJsonEncoder {
...
+        _writer.write("[{\"type\":\"");
+        _writer.write(firstValue._type == EntityType::Node ? "node" : "edge");
+        _writer.write("\",\"id\":");
+        _writer.write(std::to_string(firstValue._id.getValue()));
+        _writer.write('}');
+
+        auto it = entries.begin();
+        it++;
+        for (; it != entries.end(); ++it) {
+            const auto& [type, id] = *it;
```
cyrusknopf: would somehting like

`for (const auto& [type, id] : entries | rv::drop(1)) {}`

be more idiomatic/expressive?

query/outputs/OutputValues.h
```diff
@@ -69,7 +69,7 @@ struct ColumnTypeGenerator {
         } else if constexpr (IsNull<T>) {
             _name = fmt::format("NULL");
         } else {
-            COMPILE_ERROR("Unexpected column type");
+            _name = fmt::format("Object");
```
cyrusknopf: Why do we allow arbitrary types instead of just specialising all those that we want to output?

query/pipeline/processors/GetPropertiesProcessor.cpp
```diff
@@ -78,13 +78,15 @@ void GetPropertiesProcessor<Entity, T>::reset() {
...
-    _input.getPort()->consume();
-    _output.getPort()->writeData();
+    if (!_propWriter->isValid()) {
+        _input.getPort()->consume();
+        finish();
+    }
 
-    finish();
+    _output.getPort()->writeData();
 }
```
cyrusknopf, line 90: Why the change of logic now? Is it possible that it does not finish in one step? Also do we actually want to be marked as reset from the `reset()` call? or just reset the `_propWriter`. I was under the assumption that `reset()` was called by the `PipelineExecutor` as an external control of the processor, but maybe that understanding is wrong

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+    bioassert(_inputSources, "Input nodes column is null");
+
+    // Output
+    bioassert(_outputTargets, "Output target nodes column is null");
+    bioassert(_outputPaths, "Output paths column is null");
+    bioassert(_outputIndices, "Output indices column is null");
+    bioassert(dynamic_cast<ColumnVector<EntityList>*>(_outputPaths->getColumn()),
+              "Output paths column is not a path column");
+    bioassert(dynamic_cast<ColumnNodeIDs*>(_outputTargets->getColumn()),
+              "Output targets column is not an entity ID column");
```
cyrusknopf: Error message and cast are inconsistent

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+    _depthEnd = 0;
+    markAsReset();
+}
+
+void PathExplorerProcessor::reconstructPath(size_t entryIdx, EntityList& path) const {
+    size_t idx = entryIdx;
+
+    path.resize(_depth);
+
+    size_t i = _depth - 1;
```
cyrusknopf: Can `_depth` ever `= 0` at this point?

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        if (e.parentIdx != SIZE_MAX) {
+            auto& entry = path[i--];
+            entry._type = EntityType::Edge;
+            entry._id = e.edge.getValue();
+        }
+
+        idx = e.parentIdx;
+    }
+}
```
cyrusknopf: This loop is quite hard to follow, maybe some comments as to what is happening. Also are the `!= SIZE_MAX` conditions the most descriptive condition we can have here?

query/pipeline/processors/PathExplorerProcessor.cpp
cyrusknopf: Also is `i` the depth counter? I think `i` should have a more descriptive name

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        idx = e.parentIdx;
+    }
+
+    return false;
+}
+
+void PathExplorerProcessor::execute() {
+    auto* outputTargets = _outputTargets->getColumn()->cast<ColumnNodeIDs>();
+    auto* outputPaths = _outputPaths->getColumn()->cast<ColumnVector<EntityList>>();
```
cyrusknopf: can you use `NamedColumn::as<>` instead of first getting the Column*?

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+        const size_t inputSize = inputNodes.size();
+
+        // Seed _allEntries with root entries (one per input source node).
+        // Root entries have parentIdx = -1 and no edge.
+        _allEntries.resize(inputSize);
+        for (size_t i = 0; i < inputSize; i++) {
+            _allEntries[i] = FrontierEntry {
+                .node = inputNodes[i],
+                .edge = EdgeID {},
+                .parentIdx = SIZE_MAX,
```
cyrusknopf: I see SIZE_MAX used a lot and it seems to have some semantic meaning for the path/parent tree. Can we use a meaningfully named constexpr variable of std::numeric_limits<size_t> instead?

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+            };
+        }
+
+        if (_minHops == 0) {
+            // Write all paths corresponding to no hops
+            for (size_t i = 0; i < inputSize; i++) {
+                _outputIndices->push_back(i);
+                outputTargets->push_back(inputNodes[i]);
+                outputPaths->emplace_back();
+            }
```
cyrusknopf, line 193: So a path of 0 hops is just returning the seed/source nodes?

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+    if (_depthNeedsSetup) {
+        // Step 2. This is the first time execute() is called on the current depth.
+        //         It sets up the current depth window and feeds it to the BFS writer.
+        //         It prepares a chunk of size `windowSize` to be fed to the ChunkWriter.
+        //
+        //         > [!WARNING]
+        //         > `windowSize` can exceed the maximum ChunkSize
+        //         >   -> the memory usage is unbounded.
+
+        if (_depthStart == _depthEnd || _depth > _maxHops) {
```
cyrusknopf: I understand the `_depth > _maxHops`, but why does `_depthStart == _depthEnd` mean that we instantly finish?

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+        //           - Check if the edge is already used in the current path.
+        //           - If not, add it to the entries (to be used by deeper levels).
+        //           - Reconstruct and write the path in the output if minHops is met.
+
+        const NodeID intermediate = bfsIntermediates[i];
+        const EdgeID edge = bfsEdges[i];
+
+        // `parentIdx` references the parent of the current edge in the
+        // frontier entries array
+        const size_t parentIdx = _depthStart + bfsIndices[i];
```
cyrusknopf, line 258: Can any `i` of `bfsIndices` ever contain `SIZE_MAX`?

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,315 @@
...
+        _depth++;
+        _depthNeedsSetup = true;
+
+        if (_depthStart == _depthEnd || _depth > _maxHops) {
+            _input.getPort()->consume();
+            finish();
+        }
+    }
+
+    _output.getPort()->writeData();
```
cyrusknopf, line 299: What if all the edges in the `for (size_t i = 0; i < bfsEdges.size(); i++) {` loop on 262 are already in the path? Would we write an empty chunk?

query/pipeline/processors/PathExplorerProcessor.h
```diff
@@ -0,0 +1,104 @@
...
+        size_t sourceIdx {SIZE_MAX}; // original input source index
+    };
+
+    PathExplorerProcessor(LocalMemory* mem,
+                          PathExplorationDir dir,
+                          int64_t minHops,
+                          int64_t maxHops);
+    ~PathExplorerProcessor() override;
+
+    LocalMemory* _mem {nullptr};
```
cyrusknopf: Instead of passing a LocalMemory and allocating in the processor, would we rather allocate in the PipelineBuilder and pass pointers to the processor?

query/pipeline/PipelineBuilder.cpp
```diff
@@ -318,6 +323,42 @@ PipelineEdgeOutputInterface& PipelineBuilder::addGetEdges() {
...
+    _pendingOutput.connectTo(input);
+    input.propagateColumns(output);
+
+    Dataframe* outDf = output.getDataframe();
+
+    NamedColumn* targetNodes = allocColumn<ColumnNodeIDs>(outDf);
+    NamedColumn* pathCol = allocColumn<ColumnVector<EntityList>>(outDf);
+    const NamedColumn* indices = allocColumn<ColumnIndices>(outDf);
+
+    proc->setOutputIndicesColumn(static_cast<ColumnIndices*>(indices->getColumn()));
```
cyrusknopf: Can you use `indices->as<ColumnIndices>`?

query/pipeline/PipelineBuilder.h
```diff
@@ -224,11 +228,14 @@ class PipelineBuilder {
...
     void setOutputDataframe(const Dataframe* df);
 
+    Processor* getLastProc() const { return _lastProc; }
+
 private:
     LocalMemory* _mem {nullptr};
     PipelineV2* _pipeline {nullptr};
     DataframeManager* _dfMan {nullptr};
     PendingOutputView _pendingOutput;
+    Processor* _lastProc {nullptr};
```
cyrusknopf, line 238: Why is this now necessary?

query/plan/PipelineGenerator.cpp
```diff
@@ -1505,6 +1510,27 @@ PipelineOutputInterface* PipelineGenerator::translateShowVectorIndexesNode(ShowV
 }
 
 
+PipelineOutputInterface* PipelineGenerator::translatePathExplorerNode(PathExplorerNode* node)
+{
```
cyrusknopf:
```suggestion
PipelineOutputInterface* PipelineGenerator::translatePathExplorerNode(PathExplorerNode* node) {
```

query/plan/PipelineGenerator.cpp
```diff
@@ -1505,6 +1510,27 @@ PipelineOutputInterface* PipelineGenerator::translateShowVectorIndexesNode(ShowV
...
+    const VarDecl* edgeDecl = node->getEdgeDecl();
+    const VarDecl* targetDecl = node->getTargetDecl();
+
+    NamedColumn* pathsCol = proc->getOutputPathsColumn();
+    pathsCol->rename(edgeDecl->getName());
+    _declToColumn[edgeDecl] = pathsCol->getTag();
+
+    NamedColumn* sourceCol = proc->getOutputTargetsColumn();
+    sourceCol->rename(targetDecl->getName());
+    _declToColumn[targetDecl] = sourceCol->getTag();
```
cyrusknopf: Why do you need to rename the columns?

cyrusknopf (commented):
Maybe you considered it, but could the processor not be templated on the direction?

query/outputs/OutputValues.h
LucLabarriere: Do you mean that we could also have "Path", "EntityList", etc? I think we could indeed, it would require to change the python package every time we add a type though, maybe we could wait a bit until the API is more stable? The idea is that, right now from, if we just see "Object", it becomes a regular python dict

query/pipeline/processors/GetPropertiesProcessor.cpp
LucLabarriere: Oh yeah the `reset()` is a leftover thanks!

The reason why I changed the logic is that the original implementation assumed it would receive at most 1 chunk of data (which should almost always be the case), but it's actually a silent bug because if it receives more than a chunk, it still outputs a single chunk and stops there (I had the issue with one of PathExplorerProcessor's versions). Generally speaking you're right that if it receives an input > chunkSize, it's probably a bug, but this change protects us in the case it does not.

It would be a pretty tricky bug because it would only occur if the previous output exceeds a chunk:
`ProcessorA -[1 million outputs (bug)]-> GetProperties ->[65536 outputs (bug too)]`

query/pipeline/processors/GetPropertiesWithNullProcessor.cpp
```diff
@@ -75,11 +75,12 @@ void GetPropertiesWithNullProcessor<Entity, T>::execute() {
     reset();
```
LucLabarriere, line 75: @cyrusknopf Do you know why `GetPropertiesWithNullProcessor` has that reset() ?

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: It's called in a part of the algorithm where depth is always > 0 yes, but you're right I added a bioassert to make sure (not in this function that is a hot path though but higher up the algorithm)

LucLabarriere: And yes `i` is the cursor that walks back the path, I'll find a better name

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: I'll create a constexpr to give SIZE_MAX a better meaning, and add comments thanks!

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: Done! thanks

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: yes it's a bit weird but `MATCH (n)-[e]->{0,1}(m) RETURN n, e, m`
returns:

```
n  |  e  |  m
-------------
0  | [ ] |  0
1  | [ ] |  1
2  | [ ] |  2
... Then all paths of length 1
0  | [0] |  1
0  | [1] |  2
...
```

LucLabarriere: You can also do (unbounded path exploration, including 0 edges)
```
MATCH (n)--*(m) RETURN n, m
```

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: `_depthStart` and `_depthEnd` are cursors into the `_allEntries` that define the current window we need to expand, if `_depthStart == _depthEnd` it means that are no more edges to expand (reached the end of the graph kind of). I'll add a comment

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: No never, `bfsIndices` is the indices column outputed by the `GetXEdgesChunkWriter`, it always maps the output edge to the input node it received. `SIZE_MAX` (now renamed `ROOT`) is only when we initialise the exploration (it allows to tag the entries in `_allEntries` as being the roots of the exploration)

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: Yes exactly, we would, same thing if the `GetXEdgesChunkWriter` outputted no edges (no more edges to expand, reach the limits of the graph)

query/pipeline/processors/PathExplorerProcessor.h
LucLabarriere: Oh yes thanks!

query/pipeline/PipelineBuilder.cpp
LucLabarriere: Yes thanks!

query/pipeline/PipelineBuilder.h
LucLabarriere: Because I need to rename the columns in the pipeline builder... Not sure if I can do it differently, any idea?

```c++
const auto* proc = dynamic_cast<PathExplorerProcessor*>(_builder.getLastProc());
bioassert(proc, "Failed to cast last proc to PathExplorerProcessor");

const VarDecl* edgeDecl = node->getEdgeDecl();
const VarDecl* targetDecl = node->getTargetDecl();

NamedColumn* pathsCol = proc->getOutputPathsColumn();
pathsCol->rename(edgeDecl->getName());
_declToColumn[edgeDecl] = pathsCol->getTag();

NamedColumn* sourceCol = proc->getOutputTargetsColumn();
sourceCol->rename(targetDecl->getName());
_declToColumn[targetDecl] = sourceCol->getTag();
```

query/plan/PipelineGenerator.cpp
LucLabarriere: So there are three columns for which I need to set the name in this query:

```
MATCH (n)-[e]->*(m) RETURN n, m, e
```

- `n` -> automatically handled when we translate the var node
- `e` -> usually for regular edges (getOutEdges etc ), the var node takes care of it, but here it can't because `e` is not represented by a var node in the plan graph
- `m` -> Also taken care of by the var node that is created when we add a `GetEdgeTargetNode` but here it was also too complicated to do it this way (because GetEdgesTarget expects to occur from an edge plan graph node to properly set the entity stream etc...)

LucLabarriere:
> Maybe you considered it, but could the processor not be templated on the direction?

I did and I discarded that for some reason, I'll see if I can change it back again, see if it simplifies stuff

query/pipeline/processors/PathExplorerProcessor.cpp
```diff
@@ -0,0 +1,308 @@
...
+            break;
+        }
+
+        auto& out = path[pathCursor--];
+        out._type = EntityType::Edge;
+        out._id = e.edge.getValue();
+
+        idx = e.parentIdx;
+    }
+}
```
LucLabarriere, line 136: @cyrusknopf Is it better?

LucLabarriere:
@cyrusknopf ready for another round of review if you want to have a look!

- Fixed most remarks
- PathExplorerProcessor is now templated

Remaining questions:
- GetPropertiesWithNullProcessor: do you know why we do a reset() call in execute()?
- 'Object' as type name in the returned JSON schema: I think it's fine for now and we could make specialized types 
later on, but I think it should only be done once we handle it properly in python.
- The fact that I added `_lastProc` to the PipelineBuilder -> is there a better way? (necessary to rename columns for now)
- Cleaned up reconstructPath() method in PathExplorerProcessor.cpp

query/outputs/OutputValues.h
cyrusknopf: Ah okay, so if we had `else if constexpr (IsPath<T>)`, etc. then this would requires some changes on the Python SDK?

query/pipeline/processors/PathExplorerProcessor.cpp
cyrusknopf: Yeah I think this is clearer thanks!

query/pipeline/processors/PathExplorerProcessor.cpp
cyrusknopf: Makes sense thanks

query/pipeline/processors/PathExplorerProcessor.cpp
cyrusknopf: Cool thanks for clarification

query/pipeline/processors/PathExplorerProcessor.cpp
cyrusknopf: Is it possible that this empty chunk gets emitted, and then we output more chunks in a subsequent cycle?

query/pipeline/PipelineBuilder.h
cyrusknopf: I think I have some gaps in my understanding on how the renaming works, I though that renaming only mattered for the RETURN projection? and during the other stages in the pipeline columns were just identified by their ColumnTag. So do you need to do this renaming so that it works in RETURNS? or a different reason?

cyrusknopf:
> @cyrusknopf ready for another round of review if you want to have a look!
> 
> * Fixed most remarks
> * PathExplorerProcessor is now templated
> 
> Remaining questions:
> 
> * GetPropertiesWithNullProcessor: do you know why we do a reset() call in execute()?
> * 'Object' as type name in the returned JSON schema: I think it's fine for now and we could make specialized types
>   later on, but I think it should only be done once we handle it properly in python.
> * The fact that I added `_lastProc` to the PipelineBuilder -> is there a better way? (necessary to rename columns for now)
> * Cleaned up reconstructPath() method in PathExplorerProcessor.cpp

On the added `reset()`: yes I believe I added it due to a bug which occurred when:
1. We retrieved a property prior to the `RETURN` projection (e.g. for a filter, `WHERE n.age = 32` requires a `GetPropertiesWithNull` to be inserted
2. We retrieve a similarly* typed property in the `RETURN` projection (e.g. `RETURN n.duration`)

So for example in a full query such as
```
MATCH (n) WHERE n.age = 32 RETURN n.duration
```
> \*"similarly" meaning `EntityType` and `T` were the same, I believe.

In such a case there was a bug where the `GetPropertiesWithNull` processor which was added for the `RETURN n.duration` would start at the `end()` because of the previous scan for `n.age`. Thinking about this now, I am not sure this should happen, so I will add this to be investiaged.

That's my mistake for not documenting that when I added the `reset()` though, so I will add some inline docs for that.

Thanks for the changes! Do you like the templated version more? or prefer how it was previously?

query/outputs/OutputValues.h
LucLabarriere: Not necessarily, it's mostly that it would be weird to have special types defined in the header, but parse them as regular dicts... I think we can always do that in the near future, I'm just not sure it's necessary here

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: Yes definitely, we output one chunk everytime we run `GetXEdgesChunkWriter::fill()` so if the input to the chunk writer produces no edges, we produce an empty chunk, but then subsequent `fill()` calls may produce non empty ones

query/pipeline/PipelineBuilder.h
LucLabarriere: Yes I need it for the RETURN exactly, I have to do it somewhere since it's not automatically handled by a VarNode in this case

query/outputs/OutputValues.h
rjb32: We can keep the compile error here also for the sake of not have unintended silent cases

LucLabarriere:
> > @cyrusknopf ready for another round of review if you want to have a look!
> > 
> > * Fixed most remarks
> > * PathExplorerProcessor is now templated
> > 
> > Remaining questions:
> > 
> > * GetPropertiesWithNullProcessor: do you know why we do a reset() call in execute()?
> > * 'Object' as type name in the returned JSON schema: I think it's fine for now and we could make specialized types
> >   later on, but I think it should only be done once we handle it properly in python.
> > * The fact that I added `_lastProc` to the PipelineBuilder -> is there a better way? (necessary to rename columns for now)
> > * Cleaned up reconstructPath() method in PathExplorerProcessor.cpp
> 
> On the added `reset()`: yes I believe I added it due to a bug which occurred when:
> 
> 1. We retrieved a property prior to the `RETURN` projection (e.g. for a filter, `WHERE n.age = 32` requires a `GetPropertiesWithNull` to be inserted
> 2. We retrieve a similarly* typed property in the `RETURN` projection (e.g. `RETURN n.duration`)
> 
> So for example in a full query such as
> 
> ```
> MATCH (n) WHERE n.age = 32 RETURN n.duration
> ```
> 
> > *"similarly" meaning `EntityType` and `T` were the same, I believe.
> 
> In such a case there was a bug where the `GetPropertiesWithNull` processor which was added for the `RETURN n.duration` would start at the `end()` because of the previous scan for `n.age`. Thinking about this now, I am not sure this should happen, so I will add this to be investiaged.
> 
> That's my mistake for not documenting that when I added the `reset()` though, so I will add some inline docs for that.
> 
> Thanks for the changes! Do you like the templated version more? or prefer how it was previously?

Interesting the reset() thing, I tried to remove it and I got unit tests failing so the behaviour is still the same as of today. Yes the templated version seems better indeed, I remember that I discarded in the first draft because it required a templated lambda in PipelineGenerator, but I think it's fine:

```c++
    const auto makeProcessor = [&]<PathExplorationDir Dir>() {
        _builder.addPathExplorer<Dir>(minHops, maxHops);

        const auto* proc = dynamic_cast<PathExplorerProcessor<Dir>*>(_builder.getLastProc());
        bioassert(proc, "Failed to cast last proc to PathExplorerProcessor");

        const VarDecl* edgeDecl = node->getEdgeDecl();
        const VarDecl* targetDecl = node->getTargetDecl();

        NamedColumn* pathsCol = proc->getOutputPathsColumn();
        pathsCol->rename(edgeDecl->getName());
        _declToColumn[edgeDecl] = pathsCol->getTag();

        NamedColumn* sourceCol = proc->getOutputTargetsColumn();
        sourceCol->rename(targetDecl->getName());
        _declToColumn[targetDecl] = sourceCol->getTag();
    };

    switch (node->getDir()) {
        case PathExplorationDir::BOTH:
            makeProcessor.template operator()<PathExplorationDir::BOTH>();
        break;
        case PathExplorationDir::FORWARD:
            makeProcessor.template operator()<PathExplorationDir::FORWARD>();
        break;
        case PathExplorationDir::BACKWARD:
            makeProcessor.template operator()<PathExplorationDir::BACKWARD>();
        break;
    }

```

rjb32:
@cyrusknopf Please comment

query/pipeline/processors/PathExplorerProcessor.cpp
cyrusknopf: @rjb32 correct me if I am wrong, but I think we want to avoid this. Emitting empty chunks when it is not preferred because it creates "bubbles" in the pipeline which means that all subsequent processors execute a cycle even if their input is an empty chunk.

The intended behaviour is then to emit and empty chunk only in the case that both
1. the processor's input is closed, meaning it will never have `execute` called again;

and

2. the processor would otherwise emit and empty chunk

this means that empty chunks are stopped from propagating down the pipeline at the earliest point possible, and removes the bubbles, but also prevents the pipeline stopping due to subsequent processors never being able to execute due to never having input in the case of 1.

I fixed this in my processors by guarding the `writeData` call by a check with `input->isClosed()`

query/pipeline/PipelineBuilder.h
cyrusknopf: Ah okay, makes sense thanks

cyrusknopf:
> @cyrusknopf Please comment

I think I've replied to everything remaining (let me know if I missed something @LucLabarriere), sorry for delay

query/pipeline/processors/PathExplorerProcessor.cpp
LucLabarriere: Generally speaking this is a Pipeline execution strategy question though. It also happens in GetOutEdgesProcessor, if the processor receives a chunk of input nodes that don't have outgoing edges, it will produce an empty chunk.

This is complex because some pipeline processors may have side effects while producing empty chunks so we cannot even guaranty: "Processor outputed empty chunk? -> then it's done".

It may be the first time we encounter the issue in a processor but I suspect this will happen again because, generally speaking, we don't want a pipeline processor to keep a hold onto its executor thread. So at some point we want to pause the execution of a processor and give back control over to the orchestrator thread (for example if we want to support cancelling queries, we can't have the processor keep working until it finds at least one node because it may take hours depending on the work load).

So the issue is "we want to stop the pipeline when we encounter an empty chunk", I think this should be done when entering a pipeline processor rather than when exiting one. So my take on the strategy would be:
- Allow pipeline processor to emit empty chunks
- If a pipeline processor receives an empty chunk, it may close its output so that we stop the propagation. (That's processor dependent, some processor may want to propagate the empty chunk anyway, although I can't think of one that would require that)

LucLabarriere: A little update on what I think the correct strategy is:
- It's not that a processor should stop the propagation if its input is empty
- It's rather that a processor should stop the propagation once it treated the entirety of its input, hence be allowed to produce any number of empty chunks until it starved its input stream

cyrusknopf: I agree with the general sentiment that there may be a processor which still needs to execute even if its input is empty, but I think this is handled by the `needsData()` flag of an input port. Then we don't need empty chunks to ever propagate to processors which don't need them, and execution is partly detached from flow of chunks. Let me know if I misunderstood what you said and that this doesn't address your point.

rjb32: Yes we should avoid to emit empty chunks whenever we can because this is a waste for the downstream processors. But it should not be catastrophic if we do so, in the sense of being incorrect.

rjb32:
What is the situation regarding empty chunks now?

LucLabarriere:
> What is the situation regarding empty chunks now?

I still don't know what to do. The processor emits empty chunks right now. 

What I can do is check if we generated at least one row before calling `writeData()`, but if I understand correctly, the OrderByProcessor is supposed to receive at least one empty chunk? In which case I don't know how to ensure that exactly one empty chunk was sent. Moreover, I don't see how the PathExplorerProcessor's current behaviour is different to any other processor that would process its input and emit empty chunks if the transform operation generates no output.

For example, ScanNodes + GetOutEdges could:
- ScanNodes generates 5 chunks
- GetOutEdges generates one full chunk for the first input chunk
- Then generates no output (two empty chunks) for the two next inputs
- Then generates 5 chunks for the fourth input
- Then generates 2 chunks for the last input + one empty chunk for that same output (the remaining edges it didn't process in the two previous `execute()` cycles

rjb32:
> > What is the situation regarding empty chunks now?
> 
> I still don't know what to do. The processor emits empty chunks right now.
> 
> What I can do is check if we generated at least one row before calling `writeData()`, but if I understand correctly, the OrderByProcessor is supposed to receive at least one empty chunk? In which case I don't know how to ensure that exactly one empty chunk was sent. Moreover, I don't see how the PathExplorerProcessor's current behaviour is different to any other processor that would process its input and emit empty chunks if the transform operation generates no output.
> 
> For example, ScanNodes + GetOutEdges could:
> 
> * ScanNodes generates 5 chunks
> * GetOutEdges generates one full chunk for the first input chunk
> * Then generates no output (two empty chunks) for the two next inputs
> * Then generates 5 chunks for the fourth input
> * Then generates 2 chunks for the last input + one empty chunk for that same output (the remaining edges it didn't process in the two previous `execute()` cycles

Ok, I think that since this is not enforced for any other processor indeed we can proceed with it.

rjb32 (approved):

PR #494, opened 2026-03-06
Fixed concurrency and distance metric issues in the vectordb

LucLabarriere:
There were three main issues:
- Return a shard that was invalidated by an `evictOne()` call
- Data race when doing these steps: retrieve shard, lock shard (someone could have offloaded the shard to disk between the two calls)
- Distance metric ordering depends on the metric (one metric = ascending is better, the other is ascending is worse), so I added a boolean to keep track of that

Tested on reactome

> [!WARNING]
> Main issue right now is the OS' file opened count limit. Right now I fixed dynamically in python in my tests when I spawn turingdb, but we should probably come up with a better solution

> [!NOTE]
> Implemented a generic cache system based on a custom CLOCK algorithm (using an `age` variable instead of just a boolean)


vector/VectorDatabase.h
```diff
@@ -58,6 +58,12 @@ class VectorDatabase {
     VectorDatabase();
 
     [[nodiscard]] VectorResult<void> load();
+
+    // Lock-free overload
+    [[nodiscard]] VecLibAccessor getLibraryLocked(const VecLibID& libID);
```
rjb32: Where is it used?

vector/ShardCache.cpp
```diff
@@ -113,25 +119,40 @@ void ShardCache::evictLibraryShards(VecLibID libID) {
...
+            continue;
         }
+
+        std::unique_lock<std::mutex> lock(it->second._shard->_mutex, std::try_to_lock);
+        if (!lock.owns_lock()) {
+            continue;
+        }
+
+        // Better candidate found — take ownership of its lock, releasing the previous winner
+        victimLock = std::move(lock);
```
rjb32: This pattern is a bit weird. Usually caches where pages/shards are pinned have a lock that resides inside each frame/slot of the cache

vector/VecLibShardAccessor.h
```diff
@@ -0,0 +1,32 @@
...
+
+    explicit VecLibShardAccessor(VecLibShard* shard);
+    ~VecLibShardAccessor();
+
+    VecLibShardAccessor(const VecLibShardAccessor&) = delete;
+    VecLibShardAccessor(VecLibShardAccessor&&) = delete;
+    VecLibShardAccessor& operator=(const VecLibShardAccessor&) = delete;
+    VecLibShardAccessor& operator=(VecLibShardAccessor&&) = delete;
+
+    VecLibShard* operator->() { return _shard; }
```
rjb32: It's better to standardise the access in terms of style to something like get() or getShard() instead of relying on the -> or * operators in the STL style. I find it more readable from the outside

vector/VectorDatabase.h
LucLabarriere: Indeed not used anymore -> I'll remove that

vector/ShardCache.cpp
LucLabarriere: Indeed, I'll see what other patterns we can use

LucLabarriere:
@cyrusknopf @suhassn-turing-bio @rjb32 Ready for review!

io/cache/AgingRingCache.h
```diff
@@ -0,0 +1,557 @@
...
+    void setCalculateMemoryUsage(const CalculateMemoryUsageCallback& callback) {
+        _calculateMemoryUsage = callback;
+    }
+
+    void setMaxMemUsage(size_t maxMemUsage) {
+        _maxMemUsage = maxMemUsage;
+    }
+
+    template <typename... Args>
+    AgingRingCacheResult<Handle> acquire(const Key& key, void* data = nullptr) {
```
cyrusknopf, line 63: I'm not sure if we have rules on this in the guidelines, but I would prefer to keep only the declarations inside the `class AginRingCache {};`, and to have the defitions of the functions defined after. It makes it hard to judge the shape of the class when the entire definition is in the class definition.

io/cache/AgingRingCache.h
```diff
@@ -0,0 +1,557 @@
...
+                    });
+
+                Entry* entryPtr = &newIt->second;
+
+                // Step 2: Load the payload but keep the cache unlocked during io
+                _inFlightIoCount++;
+
+                lock.unlock();
+                const auto loadRes = _onLoad(key, *payloadPtr, data);
+                lock.lock();
```
cyrusknopf, line 94: Why do we need to unlock and then immediately lock?

io/cache/AgingRingCache.h
```diff
@@ -0,0 +1,557 @@
+#pragma once
+
+#include <condition_variable>
+#include <functional>
+#include <unordered_map>
+#include <list>
+
+#include "BioAssert.h"
```
rjb32: Put it down in the order of includes

io/cache/AgingRingCache.h
```diff
@@ -0,0 +1,557 @@
...
+    template <typename... Args>
+    AgingRingCacheResult<Handle> acquire(const Key& key, void* data = nullptr) {
+        for (;;) {
+            std::unique_lock lock(_mutex);
+
+            if (_terminating) {
+                return AgingRingCacheError::result(AgingRingCacheErrorCode::TERMINATING);
+            }
+
+            auto it = _map.find(key);
```
rjb32: it can be const?

io/cache/AgingRingCache.h
```diff
@@ -0,0 +1,557 @@
...
+                const auto it = _map.find(*victimKey);
+                bioassert(it != _map.end(), "Could not find key in map");
+
+                Entry& entry = it->second;
+                victimPayload = entry._payload.get();
+            }
+
+            _onEvict(*victimKey, **victimPayload);
+
+            std::unique_lock lock(_mutex);
```
rjb32: It would be best to have the critical section of that lock in between brackets

rjb32:
@LucLabarriere The principle look sounds as we discussed during the meeting. I just think that allocating a new payload on the heap on cache miss can be questionable if we want to use it as a buffer cache. We may want to have fixed cache frames that are preallocated.

LucLabarriere:
> @LucLabarriere The principle look sounds as we discussed during the meeting. I just think that allocating a new payload on the heap on cache miss can be questionable if we want to use it as a buffer cache. We may want to have fixed cache frames that are preallocated.

Indeed, I can have a look but it complexifies the implementation because we would have to main the stability of ptrs/references. Definitely doable but I wanted to keep it simple for this PR

io/cache/AgingRingCache.h
LucLabarriere: Indeed it's a bit ugly right now, I quickly tried to extract the implementation out of the class, but it's harder to follow in this case because the class and the methods are templated + we use a lot of using directives to simplify return types, which won't be usable outside the class and I'll have to use the full typename: `AgingRingCache<Key, Payload, Hash, Equal>::Handle`

io/cache/AgingRingCache.h
LucLabarriere: The IO are performed while unlocked to avoid contention

rjb32:
@LucLabarriere @cyrusknopf To me it's a fairly clear implementation of a cache based on the clock algorithm.
There is a dynamic allocation of a Payload object when we insert, that's fine for a general cache for now I think but it would be best to allocate from a pre-existing pool of cache frames.

rjb32 (approved):

PR #498, opened 2026-03-09
Add vectors as properties

rjb32:
Be able to store embedding vectors or feature vectors as properties of nodes and edges.

Create nodes with a vector property:
```
CREATE (n:Vec {emb: [1.0, 2.0, 3.0, 4.0]})
``` 

Fetch vector properties:
```
MATCH (n:Vec) RETURN n.emb
``` 

Filter by vector properties:
```
MATCH (n:Vec) WHERE n.emb = [1.0, 2.0, 3.0] RETURN n.emb
``` 

Filter using cosine similarity function:
```
MATCH (n:Vec) WHERE cosineSimilarity(n.emb, [1.0, 0.0, 0.0]) > 0.5 RETURN n.name
```

### DIFF
--- a/query/pipeline/processors/WriteProcessor.cpp
+++ b/query/pipeline/processors/WriteProcessor.cpp
@@ -231,6 +231,20 @@ LabelSet WriteProcessor::getLabelSet(std::span<const std::string_view> labels) {
     return labelset;
 }
 
+PropertyTypeID WriteProcessor::resolvePropertyTypeID(std::string_view name,
+                                                     ValueType type,
+                                                     const Column* valueCol) {
+    if (type == ValueType::Embedding) {
+        const auto* embCol = dynamic_cast<const ColumnEmbeddingConst*>(valueCol);
+        bioassert(embCol, "Embedding property value is not a ColumnEmbeddingConst");
+        const uint32_t dimension = embCol->dimension();
+        EmbeddingPropertyConfig config(dimension, EmbeddingPrecision::Float32);
+        return _metadataBuilder->getOrCreateEmbeddingPropertyType(name, config)._id;
+    } else {
+        return _metadataBuilder->getOrCreatePropertyType(name, type)._id;
+    }
+}
+
 void WriteProcessor::createNodes(size_t numIters) {
     NodeID nextNodeID = _writeBuffer->numPendingNodes();
     const Dataframe* outDf = _output.getDataframe();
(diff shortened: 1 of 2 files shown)


query/pipeline/processors/GetEmbeddingPropertiesProcessor.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    outCol->clear();
+    indices->clear();
+    outCol->reserve(ids->size());
+    indices->reserve(ids->size());
+
+    const GraphView& view = _ctxt->getGraphView();
+    PartIterator partIt(view);
+
+    for (; partIt.isNotEnd(); partIt.next()) {
```
rjb32: WTF it should use getpropertiesiterator

storage/dump/EmbeddingPropertyContainerLoader.h
```diff
@@ -0,0 +1,139 @@
...
+class EmbeddingPropertyContainerLoader {
+public:
+    using Constants = EmbeddingPropertyContainerDumpConstants;
+
+    explicit EmbeddingPropertyContainerLoader(fs::FilePageReader& reader)
+        : _reader(reader)
+    {
+    }
+
+    [[nodiscard]] DumpResult<std::unique_ptr<PropertyContainer>> load() {
```
rjb32: Put in cpp file

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -248,8 +258,16 @@ void WriteProcessor::createNodes(size_t numIters) {
         std::vector<CommitWriteBuffer::UntypedProperty> constProps;
         constProps.reserve(node._properties.size());
         for (const auto& [name, type, valueCol] : node._properties) {
-            const PropertyTypeID propID =
-                _metadataBuilder->getOrCreatePropertyType(name, type)._id;
+            PropertyTypeID propID;
```
rjb32: Why something special is necessary here for embedding properties compared to other properties?

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -352,8 +370,16 @@ void WriteProcessor::createEdges(size_t numIters) {
         std::vector<CommitWriteBuffer::UntypedProperty> constProps;
         constProps.reserve(edge._properties.size());
         for (const auto& [name, type, valueCol] : edge._properties) {
-            const PropertyTypeID propID =
-                _metadataBuilder->getOrCreatePropertyType(name, type)._id;
+            PropertyTypeID propID;
```
rjb32: Same here?

query/pipeline/processors/WriteProcessor.cpp
rjb32: Embedding properties need special handling because they require registering an `EmbeddingPropertyConfig` (dimension + precision) alongside the property type, which regular properties don't need. But you're right that having it inlined in both places was unnecessary — extracted into a `resolvePropertyTypeID` helper so the logic lives in one place.

query/pipeline/processors/WriteProcessor.cpp
rjb32: Addressed — both call sites now use the shared `resolvePropertyTypeID` helper.

PR #502, opened 2026-03-11
Add test suite test for join on common interest

rjb32:


ghost:
Only comment is that the name of the test-suite test is a bit weird?

PR #512, opened 2026-03-13
Use `MaskApplicator` and dispatcher, enable disabled tests

cyrusknopf:
Migrates legacy usage of `switch` cases to apply masks for the favoured `ColumnDispatcher`.


storage/columns/MaskOperators.h
```diff
@@ -115,10 +114,14 @@ struct MaskApplicator {
            }
         }
     }
+
+    static void apply(ColumnMask*, const ColumnMask*, const ColumnMask*) {
+        bioassert(false, "Invalid mask operatoin");
+    }
```
cyrusknopf, line 120:
```suggestion
    /// NOTE: Required for dispatcher type-checking, but never used.
    static void apply(ColumnMask*, const ColumnMask*, const ColumnMask*) {
        bioassert(false, "Invalid mask operation.");
    }
```

query/pipeline/processors/ApplyMask.h
```diff
@@ -0,0 +1,44 @@
...
+#include "columns/AllowedKinds.h"
+#include "columns/ColumnCombinations.h"
+#include "columns/ColumnOperatorDispatcher.h"
+#include "columns/MaskOperators.h"
+
+namespace db {
+
+namespace {
+
+struct Eval {
```
rjb32: If Eval is strictly private to the ApplyMask class then it is worth making it an internal struct to that class

rjb32 (approved):

PR #513, opened 2026-03-13
First version of the specification of a new binary file format

rjb32:
https://github.com/turing-db/turingdb/blob/file-format-spec/docs/FILE_FORMAT.md

### DIFF
--- a/docs/FILE_FORMAT.md
+++ b/docs/FILE_FORMAT.md
@@ -1,0 +1,180 @@
+# New Storage File Format
+
+We describe in this specification a new format for turing graphs.
+
+The objective is to store a TuringDB graph in one file if possible,
+so that turingdb graphs just like SQLite database files can be copied other easily,
+and for the esthetics of being self-contained and for decreasing the amount of system calls.
+
+Another object pertaining to this file format in itself is to have more compact
+disk storage for graphs than the first disk encoding format we have.
+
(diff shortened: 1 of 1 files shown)


docs/FILE_FORMAT.md
```diff
@@ -0,0 +1,182 @@
...
+* We want writers from different branches to be able to write in parallel to the file
+  
+### Parallelism and contention: 
+* Contention 1: new page allocation in the file. But we could have a pool of free pages reserved per branch.
+* Contention 2: creation of a new branch. No contention in the file for writing pages that are intra-branch.
+
+Writing commits and data parts inside a branch should be able to proceed in parallel over multiple writers provided that we have free pages and the branch is already created in the file.
+
+Why there are no contention on data parts and why we don't need a central datapart directory in the file:
+* We commits are created by inheriting data parts from their parents and creating new data parts
```
ghost, line 29: does this mean dataparts will be copied over for every commit - we could potentially use file offset pointers?

rjb32: no the dataparts are not copied of course, if you look into the detailed format structure below commits just have sections that just reference the offset of the datapart pages

ghost: Ah I see in the next few points

PR #515, opened 2026-03-16
Implement indexes framework

cyrusknopf:
Allows for the creation of indexes on unconstrained nodes and edges (i.e. no labelset-
specific indexes; no edgetype-specific indexes).

Indexes are integrated into the commit system: they require a write transaction to create.

Indexes are propagated to subsequent commits (included rebases) as long as the property which they index is not modified. A property is said to be modified if an entity which owns that property has that value updated, if an entity is created with that property, or if an entity which has that property is deleted.

Indexes are not dumped or loaded.

Indexes can be created using the following queries:
~~~cypher
CREATE INDEX my_node_index FOR (n) ON n.age

CREATE INDEX my_edge_index FOR [e] ON e.duration
~~~

for indexes on the property `age` for nodes, and `duration` for edges, respectively.


query/analyzer/CypherAnalyzer.cpp
```diff
@@ -119,9 +141,6 @@ void CypherAnalyzer::analyze() {
             case QueryCommand::Kind::SHOW_EXTENSIONS_QUERY:
             break;
 
-            default:
```
rjb32, line 122: Why? I thought this was necessary even for enum class with clang?

query/optimizer/PlanOptimizer.cpp
```diff
@@ -482,3 +500,256 @@ void PlanOptimizer::rewriteConstWriteSources() {
...
+            if (!binExpr) {
+                // Can occur with Boolean props as predicates e.g. NOT isFrench, but we
+                // cannot use index in such a case
+                return;
+            }
+
+            // Walk down the left side of any OR chain to find the first leaf equality,
+            // from which we extract the property expression used to "anchor" the full
+            // chain i.e. assert that all other expressions are on the same property.
+            // Since all we need to assertis that all property expressions are on the same
```
rjb32: "assertis" greek typo

query/optimizer/PlanOptimizer.cpp
```diff
@@ -482,3 +500,256 @@ void PlanOptimizer::rewriteConstWriteSources() {
...
+        if (!propExpr) {
+            continue;
+        }
+        // @ref propExpr now contains the property constraint (e.g n.age) which all other
+        // expressions need also have
+
+        // Traverse the expression tree, asserting that all property expressions are the
+        // same as @ref propExpr, and that all constraint values are literals.
+        // Store the literal values into @ref lookupValueLiterals.
+        std::vector<const LiteralExpr*> lookupValueLiterals;
```
rjb32: This vector can be hoistered out of the loop and cleared

query/optimizer/PlanOptimizer.cpp
```diff
@@ -482,3 +500,256 @@ void PlanOptimizer::rewriteConstWriteSources() {
...
+    const GraphMetadata& md = _view.metadata();
+    const PropertyTypeMap& props = md.propTypes();
+    const std::optional<PropertyType> maybeProp = props.get(propName);
+    bioassert(maybeProp.has_value(), "Invalid property to index.");
+    const PropertyType prop = *maybeProp;
+    const PropertyTypeID propID = prop._id;
+
+    const Index* matchingIndex = nullptr;
+    // TODO: Check if there is more than one index
+    for (const WeakArc<Index>& index : _view.indexes()) {
```
rjb32, line 655: Isn't it a bit funny that we have to loop over indexes to get the one matching the propID? Could we have a map for that?

storage/reader/GraphReader.cpp
```diff
@@ -1,13 +1,15 @@
...
 #include "DataPart.h"
+#include "ID.h"
 #include "NodeContainer.h"
 #include "EdgeContainer.h"
 #include "indexers/EdgeIndexer.h"
 #include "views/EdgeView.h"
 #include "properties/PropertyManager.h"
 
 #include "BioAssert.h"
+#include <algorithm>
```
rjb32: STL headers should be up the header file

query/analyzer/CypherAnalyzer.cpp
cyrusknopf: I will double check by running a full matrix, but I think it should compile. By not having default it removes the risk of forgetting to implement new nodes

query/optimizer/PlanOptimizer.cpp
cyrusknopf: haha fixed thanks

query/optimizer/PlanOptimizer.cpp
cyrusknopf: Fixed thanks

query/optimizer/PlanOptimizer.cpp
cyrusknopf: I left it as a linear search for now because I envisioned we would have more complicated "tiering" of indexes, like the decision tree we discussed, e.g. indexes which index a property for nodes which have a certain labelset:

`CREATE INDEX my-index FOR (n:Person) ON n.name`

with this, we cannot have a single-tiered map of PropID -> Index, because we also need to consider labelsets.

For this reason, I didn't want to make big decisions about adding a map in CommitData that would probably need to be removed. Also, since I imagine the number of indexes at any one time is << 100, I thought a linear search was fine.

storage/reader/GraphReader.cpp
cyrusknopf: Fixed thanks

rjb32 (approved):

PR #519, opened 2026-03-17
RETURN propertyTypes(n) feature + fast retrieval of property types of an entity from property manager

LucLabarriere:
- Loading times from the binary before/after: 2.9/3.2 seconds (reactome)
- quiet mode `MATCH (n) RETURN propertyTypes(n)` on reactome: 400 ms
- quiet mode `MATCH (n)-[e]->(m) RETURN propertyTypes(e)`: 508 ms


storage/columns/Functions.h
```diff
@@ -181,6 +187,90 @@ class EuclideanDistanceFunction {
...
+            bioassert(name, "Could not get name of PropertyTypeID {}.", ptId.getValue());
+
+            if (!first) {
+                out += ", ";
+            }
+            out += *name;
+            first = false;
+        }
+    }
+}
```
cyrusknopf: Can we template this on TypedInternalID instead? then have an `if constexpr` to retrieve the PropertyManager* (nodeProperties or edgeProperties)? since the functions are otherwise identical

storage/properties/PropertyManager.cpp
```diff
@@ -34,31 +46,55 @@ void PropertyManager::fillEntityPropertyView(EntityID entityID,
     };
 
     for (const auto& [ptID, rawContainer] : _uint64s) {
+        if (!has(ptID, entityID)) {
+            continue;
+        }
+
```
cyrusknopf: Is this just an early-exit optimisation?

storage/DataPart.cpp
```diff
@@ -162,6 +162,8 @@ CommitResult<void> DataPart::load(const GraphView& view, JobSystem& jobSystem, D
 
     jobs.wait();
 
+    _nodeProperties->buildTypeMapping();
+
```
cyrusknopf: I presume that both the node property and edge property type mapping needs to be rebased in DataPartRebaser.

LucLabarriere: Yes indeed, at added the rebasing of entity IDs too

storage/properties/PropertyManager.cpp
LucLabarriere: Was a left over, I removed it thanks

storage/properties/PropertyTypeTrie.h
```diff
@@ -0,0 +1,277 @@
...
+class PropertyTypeTrie {
+    static constexpr uint16_t kNoNode = std::numeric_limits<uint16_t>::max();
+
+public:
+    void build(const std::unordered_map<PropertyTypeID, std::span<const EntityID>>& typesToEntities,
+               EntityID firstCoreEntityID,
+               size_t coreEntityCount) {
+        const Profile profile("PropertyTypeTrie::build");
+
+        _firstCoreEntityID = firstCoreEntityID;
```
cyrusknopf: This is not templated, shouldn't all the implementation be in the cpp file

storage/properties/PropertyTypeTrie.h
```diff
@@ -0,0 +1,277 @@
...
+#include "ID.h"
+#include "PropertyTypeSet.h"
+
+#include "BioAssert.h"
+#include "Profiler.h"
+
+namespace db {
+
+class PropertyTypeTrie {
+    static constexpr uint16_t kNoNode = std::numeric_limits<uint16_t>::max();
```
cyrusknopf: What is kNoNode

storage/properties/PropertyTypeTrie.h
```diff
@@ -0,0 +1,277 @@
...
+        _patchEntityToNode = std::move(newMap);
+        _firstCoreEntityID = remapper(_firstCoreEntityID);
+    }
+
+private:
+    struct TrieNode {
+        PropertyTypeSet propertyTypeSet;
+        uint16_t lacksPt {kNoNode};
+        uint16_t hasPt {kNoNode};
+    };
```
cyrusknopf, line 93:
```suggestion
    struct TrieNode {
        PropertyTypeSet _propertyTypeSet;
        uint16_t _lacksPt {kNoNode};
        uint16_t _hasPt {kNoNode};
    };
```

storage/properties/PropertyTypeTrie.h
```diff
@@ -0,0 +1,277 @@
...
+     * @return True if the entity has the property type, false otherwise.
+     * */
+    bool hasPropertyType(EntityID entityId, PropertyTypeID ptId) const {
+        if (entityId < _firstCoreEntityID) {
+            return _nodes[_patchEntityToNode.at(entityId)].propertyTypeSet.contains(ptId);
+        }
+
+        const size_t idx = (entityId - _firstCoreEntityID).getValue();
+        if (idx >= _coreEntityCount || _coreEntityToNode[idx] == kNoNode) {
+            throw std::out_of_range("Entity not found in PropertyTypeTrie");
```
cyrusknopf: Why not one of our exception classes?

cyrusknopf:
Also any thoughts on how this would work w.r.t SET?

storage/properties/PropertyTypeTrie.h
LucLabarriere: Sorry, claude code

LucLabarriere:
> Also any thoughts on how this would work w.r.t SET?

That's a good question, I would say:

- The PropertyTypeSet is the representation of the properties a node has in the current DataPart, it's not related to deletes or updates
- If a node has a new property in a later datapart, it gets a dedicated PropertyTypeSet in the new datapart (say property type id 5):
    - Datapart 0: set = [0, 1, 2]
    - Datapart 1: set = [5]

For deletions, we will have to have some sort of tombstones on property types, our delete the node itself ?

storage/columns/Functions.h
```diff
@@ -181,6 +187,69 @@ class EuclideanDistanceFunction {
...
+
+class PropertyTypesFunction {
+public:
+    using ResultType = std::string;
+    std::unordered_set<PropertyTypeID> _seen;
```
rjb32: _seen should not be public

storage/columns/Functions.h
```diff
@@ -325,4 +394,40 @@ struct BinaryFunc {
...
+    static void apply(ColumnVector<std::string>* res,
+                      const ColumnNodeIDs* arg,
+                      GraphView view) {
+        const size_t size = arg->size();
+        res->resize(size);
+
+        const auto& argd = arg->getRaw();
+        auto& resd = res->getRaw();
+
+
```
rjb32: Too many blank lines

storage/columns/Functions.h
rjb32: Why do we need to even deduplicate with an unordered_set here? Why can't we just give the property types of each NodeID regardless of repetitions?

storage/properties/PropertyTypeSet.h
```diff
@@ -0,0 +1,42 @@
...
+        return std::find(begin(), end(), ptID) != end();
+    }
+
+    std::vector<PropertyTypeID>::const_iterator begin() const { return _ids.begin(); }
+    std::vector<PropertyTypeID>::const_iterator end() const { return _ids.end(); }
+
+    size_t size() const { return _ids.size(); }
+    void clear() { _ids.clear(); }
+
+    const std::vector<PropertyTypeID>& get() const { return _ids; }
```
rjb32: Use type alias for this vector and the type returned by the getter

storage/properties/PropertyTypeTrie.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include "Profiler.h"
+#include "FatalException.h"
+
+using namespace db;
+
+PropertyTypeTrie::PropertyTypeTrie() = default;
+
+PropertyTypeTrie::~PropertyTypeTrie() = default;
+
+void PropertyTypeTrie::build(const std::unordered_map<PropertyTypeID, std::span<const EntityID>>& typesToEntities,
```
rjb32: Arguments must be aligned each below the other

storage/properties/PropertyTypeTrie.h
```diff
@@ -0,0 +1,103 @@
...
+namespace db {
+
+class PropertyTypeTrie {
+    static constexpr uint16_t INVALID_NODE = std::numeric_limits<uint16_t>::max();
+
+public:
+    PropertyTypeTrie();
+    ~PropertyTypeTrie();
+
+    void build(const std::unordered_map<PropertyTypeID, std::span<const EntityID>>& typesToEntities,
```
rjb32: Arguments must be aligned each below the other

storage/columns/Functions.h
LucLabarriere: This is mostly for the SET command and it depends how it will be implemented. If we just append modifications in newer dataparts, a node will have multiple definitions of the property and the property type ID will appear multiple times.

```
SET n.displayName = "hello"
COMMIT
SET n.displayName = "world"
```

displayName is defined twice (in two dataparts), hence the need to deduplicate. That will depend on the actual implementation of SET

cyrusknopf:
Looks nice, I think it works well with the current SET implementation, currently you can't delete properties anyway, so whilst it might need a rethink to support that, so will the SET itself.

It is worth adding a unit test for testing the rebase behaviour. I have tested the below manually and it appears to work:

~~~
change new
change new

checkout change-1
match (n) where n.name = "Cyrus" set n.age = 23, n.new_prop1 = false
change submit

checkout change-0
match (n) where n.name = "Remy" set n.new_prop2 = true
change submit

match (n) return n.name, n.propertyTypes()
~~~

i.e. both changes create a new property, and they should be rebased correctly so that Remy has new_prop2 and I have new_prop1. We need this case and other similar examples in ChangeQueriesTest.

PR #520, opened 2026-03-17
Add types::Embedding property type

rjb32:
Adds the support for the Embedding property type as a std::span<float>.
Uses the same EmbeddingBucket/EmbeddingContainer strategy as for StringContainer.
Uses a vector of views so that spans have stability and memcpy tricks are possible.

Does not contain query language integration.

### DIFF
--- a/test/storage/dump/EmbeddingGraphDumpLoadTest.cpp
+++ b/test/storage/dump/EmbeddingGraphDumpLoadTest.cpp
@@ -7,9 +7,6 @@
 
 #include "Graph.h"
 #include "comparators/GraphComparator.h"
-#include "metadata/PropertyType.h"
-#include "reader/GraphReader.h"
-#include "versioning/Transaction.h"
 #include "writers/GraphWriter.h"
 
 using namespace db;
@@ -26,7 +23,6 @@ TEST_F(EmbeddingGraphDumpLoadTest, NodeEmbeddings) {
     const size_t dimension = 4;
     const size_t nodeCount = 100;
 
-    // Build a graph with nodes carrying both scalar and embedding properties
     auto graph = Graph::create("embgraph", fs::Path {_outDir} / "original");
     {
         GraphWriter writer(graph.get());
@@ -47,61 +43,19 @@ TEST_F(EmbeddingGraphDumpLoadTest, NodeEmbeddings) {
         writer.submit();
     }
 
-    // Dump
-    {
-        GraphDumper dumper;
-        auto res = dumper.dump(*graph, dumpPath);
-        if (!res) {
-            throw TuringException("Failed to dump graph:\n" + res.error().fmtMessage());
-        }
+    GraphDumper dumper;
+    auto res = dumper.dump(*graph, dumpPath);
+    if (!res) {
+        throw TuringException("Failed to dump graph:\n" + res.error().fmtMessage());
     }
 
-    // Load
     auto loadedGraph = Graph::create();
-    {
-        const auto loadRes = GraphLoader::load(loadedGraph.get(), dumpPath);
-        if (!loadRes) {
-            throw TuringException("Failed to load graph:\n" + loadRes.error().fmtMessage());
-        }
+    const auto loadRes = GraphLoader::load(loadedGraph.get(), dumpPath);
+    if (!loadRes) {
+        throw TuringException("Failed to load graph:\n" + loadRes.error().fmtMessage());
     }
 
     ASSERT_TRUE(GraphComparator::same(*graph, *loadedGraph));
-
-    // Verify embedding values survive the round-trip
-    const FrozenCommitTx tx = loadedGraph->openTransaction();
-    const GraphReader reader = tx.readGraph();
-
-    const auto& propTypes = tx.viewGraph().metadata().propTypes();
-    const auto vecType = propTypes.get("vec");
-    const auto idxType = propTypes.get("idx");
-    ASSERT_TRUE(vecType.has_value());
-    ASSERT_TRUE(idxType.has_value());
-    ASSERT_EQ(vecType->_valueType, ValueType::Embedding);
-
-    ColumnNodeIDs nodeIDs;
-    for (size_t i = 0; i < nodeCount; i++) {
-        nodeIDs.push_back(NodeID(i));
-    }
-
-    const auto vecRange = reader.getNodeProperties<types::Embedding>(vecType->_id, &nodeIDs);
-
-    size_t total = 0;
-    for (auto it = vecRange.begin(); it.isValid(); it.next()) {
-        const NodeID nodeID = it.getCurrentEntityID();
-        const auto view = it.get();
-        ASSERT_EQ(view.size(), dimension);
-
-        const auto* origIdx = reader.tryGetNodeProperty<types::Int64>(idxType->_id, nodeID);
-        ASSERT_NE(origIdx, nullptr);
-        const size_t i = static_cast<size_t>(*origIdx);
-
-        for (size_t d = 0; d < dimension; d++) {
-            ASSERT_EQ(view[d], static_cast<float>(i * dimension + d));
-        }
-        total++;
-    }
-
-    ASSERT_EQ(total, nodeCount);
 }
 
 TEST_F(EmbeddingGraphDumpLoadTest, NodeAndEdgeEmbeddings) {
@@ -127,7 +81,6 @@ TEST_F(EmbeddingGraphDumpLoadTest, NodeAndEdgeEmbeddings) {
             nodes.push_back(node);
         }
 
-        // Create edges between consecutive nodes with embedding properties
         for (size_t i = 0; i + 1 < nodeCount; i++) {
             const auto edge = writer.addEdge("LINKS", nodes[i], nodes[i + 1]);
             for (size_t d = 0; d < dimension; d++) {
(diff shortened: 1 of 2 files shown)


rjb32:
@cyrusknopf I think it is ready for review. There is no integrate with the query language voluntarily for now to keep things not to big in this PR

query/outputs/JsonEncoder.h
```diff
@@ -121,6 +121,23 @@ class ChunkJsonEncoder {
...
     }
 
+    void encodeValue(types::Embedding::Primitive value) {
+        _writer.write("[");
+        for (size_t i = 0; i < value.size(); i++) {
+            if (i > 0) _writer.write(",");
+            _writer.write(std::to_string(value[i]));
+        }
+        _writer.write("]");
+    }
```
cyrusknopf, line 134: For writing variable lengthed types, I think we should use the below pattern of writing the head :: tail to avoid a branch to check if we need to write a comma in each iteration, making the loop tighter.

```cpp
template <typename T>
    void operator()(const ColumnConst<T>* col) {
        if (_logicalRowCount == 0) {
            return;
        }

        const T& firstValue = col->operator[](0);
        encodeValue(firstValue);

        for (size_t row = 1; row < _logicalRowCount; row++) {
            _writer.write(",");

            const T& value = col->operator[](row);

            encodeValue(value);
        }
    }
```

query/outputs/OutputValues.h
```diff
@@ -55,12 +55,18 @@ concept IsPath = std::is_same_v<T, Path>;
 template <typename T>
 concept IsNull = std::is_same_v<T, PropertyNull>;
 
+template <typename T>
+concept IsEmbedding = std::is_same_v<T, types::Embedding::Primitive>
+                   || std::is_same_v<T, std::optional<types::Embedding::Primitive>>;
+
```
cyrusknopf, line 61: You can use `TypeUtils::unwrap_optional` to avoid having 2 `is_same_v` clauses for clarity.
```cpp
template <typename T>
    struct unwrap_optional {
        using underlying_type = T;
    };
```

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -1,6 +1,7 @@
 #include "WriteProcessor.h"
 
 #include <algorithm>
+#include <numeric>
```
cyrusknopf: Still needed?

server/DBServerProcessor.cpp
```diff
@@ -677,6 +677,7 @@ void DBServerProcessor::get_node_properties() {
                 treat.template operator()<types::String>();
                 break;
             }
+            case db::ValueType::Embedding:
```
cyrusknopf: Why is embedding not allowed here, but all other property types are?

server/HTTPResponseWriter.h
```diff
@@ -290,6 +290,15 @@ class HTTPResponseWriter {
...
     }
 
+    void writeValue(std::span<const float> v) {
+        write('[');
+        for (size_t i = 0; i < v.size(); i++) {
+            if (i > 0) write(',');
+            write(std::to_string(v[i]));
+        }
+        write(']');
+    }
```
cyrusknopf, line 303: See my above comment about writing head :: tail pattern

server/HTTPResponseWriter.h
```diff
@@ -302,7 +311,13 @@ class HTTPResponseWriter {
...
         size_t count = 0;
         for (const auto& [ptID, value] : props) {
-            std::visit([&](const auto& v) { writeKeyValue(propTypes.getName(ptID).value(), *v); }, value);
+            std::visit([&](const auto& v) {
+                if constexpr (std::is_same_v<std::decay_t<decltype(v)>, std::span<const float>>) {
+                    writeKeyValue(propTypes.getName(ptID).value(), v);
+                } else {
+                    writeKeyValue(propTypes.getName(ptID).value(), *v);
+                }
+            }, value);
```
cyrusknopf: I'm not sure I understand what this is for?

server/PayloadWriter.h
```diff
@@ -280,7 +293,11 @@ class PayloadWriter {
                 std::visit(
                     [&](const auto& v) {
                         this->key(propTypes.getName(ptID).value());
-                        this->value(*v);
+                        if constexpr (std::is_same_v<std::decay_t<decltype(v)>, std::span<const float>>) {
+                            this->value(v);
+                        } else {
+                            this->value(*v);
+                        }
```
cyrusknopf: Again why do we need a specialisation just for optional/non optional embedding properties, but no others?

storage/dump/DataPartLoader.cpp
```diff
@@ -219,6 +219,19 @@ DumpResult<WeakArc<DataPart>> DataPartLoader::load(const fs::Path& dataPartDir,
...
+                auto props = loader.load();
+                if (!props) {
+                    return props.get_unexpected();
+                }
+
+                auto* container = props.value().release();
+                manager._map.emplace(pt->_id, static_cast<PropertyContainer*>(container));
+                manager._embeddings.emplace(pt->_id, static_cast<PropertyContainer*>(container));
+                break;
+            }
```
cyrusknopf, line 234: Do we have tests for loading and dumping of a graph that has embedding properties? I see we have for individual containers, but do we want a graph load/dump too?

storage/mergers/DataPartMerger.cpp
```diff
@@ -174,6 +176,38 @@ std::unique_ptr<DataPartBuilder> DataPartMerger::merge(DataPartSpan dataParts) c
...
+                    for (const auto& [start, size] : keptRanges) {
+                        std::memcpy(newContainer._values._views.data() + _nodePropertiesCount[propertyID],
+                                    oldContainer._values._views.data() + start,
+                                    (size * sizeof(std::span<const float>)));
+                        std::memcpy(newContainer.ids().data() + _nodePropertiesCount[propertyID],
+                                    oldContainer.ids().data() + start,
+                                    (size * sizeof(EntityID)));
+                        _nodePropertiesCount[propertyID] += size;
+                    }
+
```
cyrusknopf, line 210: I think @suhassn-turing-bio has more context on this section than me

storage/EmbeddingContainer.h
```diff
@@ -0,0 +1,56 @@
...
+        _views = std::move(other._views);
+        return *this;
+    }
+
+    void alloc(types::Embedding::Primitive content);
+
+    const types::Embedding::Primitive& getView(size_t index) const { return _views[index]; }
+
+    size_t getDimension() const { return _dimension; }
+    size_t getEmbeddingCount() const { return _views.size(); }
```
cyrusknopf: Should this not be `size_t size() const` to be uniform with `PropertyContainer`?

test/query-test-suite/QueryTestRunner.cpp
```diff
@@ -144,6 +144,16 @@ template <typename T>
...
+[[maybe_unused]] std::string valueToString(const std::span<const float>& value) {
+    std::string result = "[";
+    for (size_t i = 0; i < value.size(); i++) {
+        if (i > 0) result += ",";
+        result += std::to_string(value[i]);
+    }
+    result += "]";
+    return result;
+}
+
```
cyrusknopf, line 159: Same thing with head : tail stringify

test/query/pipeline/PipelineTest.cpp
```diff
@@ -1,12 +1,12 @@
 #include <gtest/gtest.h>
 
 #include <math.h>
+#include <numeric>
```
cyrusknopf: Still needed?

query/outputs/JsonEncoder.h
rjb32: Yes indeed we can extract the first iteration of the loop.

storage/EmbeddingContainer.h
rjb32: Indeed fixed

query/outputs/JsonEncoder.h
rjb32: Fixed, I extracted the first iteration

query/outputs/OutputValues.h
rjb32: Fixed

server/HTTPResponseWriter.h
rjb32: Fixed

server/PayloadWriter.h
rjb32: Fixed

test/query/pipeline/PipelineTest.cpp
rjb32: Fixed

query/pipeline/processors/WriteProcessor.cpp
rjb32: Fixed

server/DBServerProcessor.cpp
rjb32: Now handled like the other types

server/HTTPResponseWriter.h
rjb32: Indeed I extracted the first iteration

test/query-test-suite/QueryTestRunner.cpp
rjb32: Fixed

storage/dump/DataPartLoader.cpp
rjb32: Currently there is only a round trip test for dump & load of the EmbeddingContainer itself, not the full graph with embeddings

rjb32: I just added a full graph dump & load test with embeddings

test/storage/dump/EmbeddingGraphDumpLoadTest.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+        const auto* origIdx = reader.tryGetNodeProperty<types::Int64>(idxType->_id, nodeID);
+        ASSERT_NE(origIdx, nullptr);
+        const size_t i = static_cast<size_t>(*origIdx);
+
+        for (size_t d = 0; d < dimension; d++) {
+            ASSERT_EQ(view[d], static_cast<float>(i * dimension + d));
+        }
+        total++;
+    }
```
cyrusknopf: Since we call `GraphComparator::same` what is the need for the rest of these checks?

rjb32: Indeed. There was an early return true in GraphComparator previously that made the comparison of data parts unreachable. So I removed the code that was generator in the unit test to remediate to that.

Issue #521, opened 2026-03-17
`CALL db.history()` does not work whilst on a change


cyrusknopf:
```
turing:simpledb> change new
+----------+
| changeID |
+----------+
| 0        |
+----------+
Query returned 1 rows.
Query executed in 0.666674 ms.
turing:simpledb> checkout change-0
turing:simpledb@0> call db.history()
[2026-03-17 17:50:18] [error] EXEC_ERROR: Unexpected exception: Internal Error: The assertion 'data._commit' failed at /home/cyrus/turingdb/query/pipeline/procedures/HistoryProcedure.cpp:107
headCommitHash not found

turing:simpledb@0>
```

rjb32:
Fixed in main by #522

PR #522, opened 2026-03-17
Fix history procedures when on a change

rjb32:
Fix for https://github.com/turing-db/turingdb/issues/521

When we are doing history in a change, we are given a pending commit which is not stored in VersionController but in the Change. So if we are on a change, we start for now from the current head in VersionController.
This will be changed when we revamp VersionController and changes.

### DIFF
--- a/query/pipeline/procedures/DescribeCommitProcedure.cpp
+++ b/query/pipeline/procedures/DescribeCommitProcedure.cpp
@@ -10,6 +10,8 @@
 #include "columns/ColumnVector.h"
 #include "columns/ColumnConst.h"
 #include "versioning/Commit.h"
+#include "versioning/CommitBuilder.h"
+#include "versioning/Transaction.h"
 #include "versioning/VersionController.h"
 
 #include "PipelineException.h"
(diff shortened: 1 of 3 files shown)


ghost:
This could work by checking if we had a read or frozen transaction. In the frozen transaction case we do what the code path does. In the read transaction case we retrieve the tip commit of the change and follow back the prevCommit pointers from there?

rjb32:
> This could work by checking if we had a read or frozen transaction. In the frozen transaction case we do what the code path does. In the read transaction case we retrieve the tip commit of the change and follow back the prevCommit pointers from there?

Yes I considered it and I want to do this once the changes are proper branches just like "main". For now there is not an equality with main, it's just changes.

That's why I wondered what db history should be from a user perspective given that we will change this very soon. Should it be only what is "submitted" and available for all in main? To be considered "the db history"

ghost:
hmm I see what you mean, but depending on when we get to the change system refactor, this could leave users and also us developers in a problematic situation if we want to jump between commits on a change.

rjb32:
> hmm I see what you mean, but depending on when we get to the change system refactor, this could leave users and also us developers in a problematic situation if we want to jump between commits on a change.

Ok fixed. We now follow the commits in a given change when we are in a change.

PR #525, opened 2026-03-19
Implement a fuzzing pipeline based on AFL++

rjb32:
Implement a fuzzing pipeline based on AFL++ for 2 levels:
* Cypher query engine level: the input is cypher queries
* HTTP parser level: the input is HTTP queries

The goal is to trigger signal crashes (SIGSEGV, SIGABRT..etc) and abnormal exception crashes (FatalException, TuringException not transformed into a higher level exception). This catches notably bioassert failures.

The README: https://github.com/turing-db/turingdb/blob/fuzzing/fuzz/README.md

How to run:
* Read the README.md in the fuzz/ directory
* Install afl++ with `apt-get install afl++`
* Run fuzz/run_afl.sh (follow the instructions in the README.md)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,7 +124,9 @@ if(NOT APPLE)
 endif()
 
 # C++ flags
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=${CPP_STANDARD} ${CPP_WARNINGS} ${OPTIMIZE_FLAGS}")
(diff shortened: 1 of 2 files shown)


cyrusknopf:
Does this need to be in the main repo? It seems like it should be an external tool

rjb32:
> Does this need to be in the main repo? It seems like it should be an external tool

There are fuzz harnesses that are compiled against TuringDB code and depend on it, to make binaries for the fuzzer. So there is an actual code dependency.
We could put it in an another repository but it's not just strictly external to TuringDB as something that could be applied from outside to turingdb. It could be difficult to keep in sync with turingdb..etc.

PR #526, opened 2026-03-19
Fix overflow in HTTP parser based on fuzzing inputs

rjb32:
Fix overflows in HTTP parser due to 2 causes:
* In UriParser the condition `if (i > HTTP::MAX_PARAM_COUNT) {` needs to be 
`if (i >= HTTP::MAX_PARAM_COUNT) {`

Example: this query sends 9 key-value pairs, the 9th was written out of bound because i=8 which is MAX_PARAM_COUNT
```
GET ?a=1&b=2&c=3&d=4&e=5&f=6&g=7&h=8&i=9 HTTP/1.1\r\nContent-Length: 0\r\n\r\n
``` 

* We can not use strtoull for content length parsing because the buffer is not null-terminated and it behaves badly if the input is arbitrarily-long

Example:
```
GET / HContent-Length:499999999999999999999999999999999999999999999999999&G1.-=Sv GET
```

### DIFF
--- a/net/http_parser/HTTPParser.h
+++ b/net/http_parser/HTTPParser.h
@@ -212,10 +212,21 @@ private:
             if (!isdigit(c)) {
                 break;
             }
+
+            // Uses Horner's rule for decimal number parsing
+            // (if you don't know what it is please read about it)
+            //
+            // Arithmetic overflow: we first check that _payloadSize will not exceed BUFFER_SIZE
+            // if multiplied by 10 and therefore will not overflow as BUFFER_SIZE is way smaller
+            // than the bound
             if (_payloadSize > NetBuffer::BUFFER_SIZE / 10) {
                 return BadResult(HTTP::Error::REQUEST_TOO_BIG);
             }
+
+            // We know that c is a digit between '0' and '9' here
+            // because we checked isdigit earlier
             _payloadSize = _payloadSize * 10 + (c - '0');
+
             _currentPtr++;
         }
 


net/http_parser/HTTPParser.h
```diff
@@ -195,16 +195,30 @@ class HTTPParser : public AbstractHTTPParser {
...
+        _payloadSize = 0;
+        while (_currentPtr != endPtr) {
+            const char c = *_currentPtr;
+            if (!isdigit(c)) {
+                break;
+            }
+            if (_payloadSize > NetBuffer::BUFFER_SIZE / 10) {
+                return BadResult(HTTP::Error::REQUEST_TOO_BIG);
+            }
+            _payloadSize = _payloadSize * 10 + (c - '0');
```
cyrusknopf, line 228: Why `10`

test/net/HTTPParserTest.cpp
```diff
@@ -0,0 +1,378 @@
...
+    writer.writeString(data, size);
+
+    net::HTTPParser<net::URIParser> parser(&buffer);
+
+    try {
+        return parser.analyze();
+    } catch (const FatalException&) {
+        throw;
+    } catch (...) {
+    }
```
cyrusknopf, line 31: Should this be empty?

net/http_parser/HTTPParser.h
rjb32: That's Horner's rule for decimal number parsing https://en.wikipedia.org/wiki/Horner%27s_method
You multiply by ten what you got so far at each iteration and you add the new digit. 

That's also why I need to check first that I don't exceed BUFFER_SIZE divided by ten to also avoid numerical overflow.

test/net/HTTPParserTest.cpp
rjb32: I just fixed it now to rethrow anything in the catch-all block

net/http_parser/HTTPParser.h
cyrusknopf: I see, I think its good to have these things as comments then

PR #529, opened 2026-03-19
Integrate embedding properties into the query language

rjb32:


rjb32:
Ready for review @cyrusknopf 
I added a python regress test as well to test interaction through python sdk

query/AST/Literal.cpp
```diff
@@ -57,6 +57,20 @@ void MapLiteral::set(Symbol* key, Expr* value) {
...
+
+EmbeddingLiteral::~EmbeddingLiteral() {
+}
+
+EmbeddingLiteral* EmbeddingLiteral::create(CypherAST* ast, std::vector<float>&& data) {
+    EmbeddingLiteral* literal = new EmbeddingLiteral(std::move(data));
+    ast->addLiteral(literal);
+    return literal;
+}
+
```
cyrusknopf, line 73: Do we want to use move semantics in this way?

query/parser/ParserUtils.cpp
```diff
@@ -0,0 +1,46 @@
...
+    out.reserve(list->size());
+    for (Expr* elem : *list) {
+        const auto* litExpr = static_cast<const LiteralExpr*>(elem);
+        const Literal* lit = litExpr->getLiteral();
+        if (lit->getKind() == Literal::Kind::DOUBLE) {
+            const auto* doubleLit = static_cast<const DoubleLiteral*>(lit);
+            out.push_back(static_cast<float>(doubleLit->getValue()));
+        } else if (lit->getKind() == Literal::Kind::INTEGER) {
+            const auto* intLit = static_cast<const IntegerLiteral*>(lit);
+            out.push_back(static_cast<float>(intLit->getValue()));
```
cyrusknopf: Arbitrary loss of precision + potential overflow with the cast from int. Not sure there's a pretty way to handle this but should be noted

regress/embedding_properties/main.py
```diff
@@ -0,0 +1,105 @@
...
+
+    assert len(names) == 3, f"Expected 3 nodes, got {len(names)}"
+
+    for i, name in enumerate(names):
+        assert name in expected_vecs, f"Unexpected node name '{name}'"
+        expected = expected_vecs[name]
+        actual = vecs[i]
+        assert len(actual) == len(expected), f"Dimension mismatch for {name}"
+        for j in range(len(expected)):
+            assert abs(actual[j] - expected[j]) < 1e-5, \
```
cyrusknopf, line 60: Why is 1e-5 used as the sufficiently similar threshold here but 1e-4 is used below

storage/columns/BinaryPredicates.h
```diff
@@ -277,9 +279,75 @@ struct BinaryPredicate {
...
+
+        const bool equal = (a->size() == b->size())
+                           && std::equal(a->begin(), a->end(), b->begin());
+        if constexpr (std::is_same_v<F, std::equal_to<>>) {
+            return CustomBool{equal};
+        } else {
+            return CustomBool{!equal};
+        }
+    }
+
```
cyrusknopf: We want to avoid having such boiler plate for each optional combination for each type. I recommend using `optionalPredicate` in `BinaryPredicates.h`, so all you need to do is define the `Pred` functor. I imagine it would look something like:
```
struct TuringEqual {
operator()(Embedding a, Embedding b) {
    // ... logic for equality of embeddings
}

operator()(T a, U b) {
    return std::equal<>(a,b);
}

}
```

storage/columns/ColumnCombinations.h
```diff
@@ -202,6 +202,24 @@ class FunctionColumnResult {
...
+    using InternalRes = typename Op::ResultType;
+    using DecayT = TypeUtils::decay_col_t<PColT>;
+    using DecayU = TypeUtils::decay_col_t<PColU>;
+    static constexpr bool isVector = IsBinaryFuncVector<DecayT>::value
+                                  || IsBinaryFuncVector<DecayU>::value;
+public:
+    using ResultColumnType = std::conditional_t<isVector,
+        ColumnVector<InternalRes>, ColumnConst<InternalRes>>;
+};
+
```
cyrusknopf: Can we follow the existing pattern instead?:
```
template <typename AnyT, typename InternalT>
class FunctionResultImpl<ColumnVector<AnyT>, InternalT> {
public:
    using ResultColumnType = ColumnVector<InternalT>;
};

template <typename AnyT, typename InternalT>
class FunctionResultImpl<ColumnConst<AnyT>, InternalT> {
public:
    using ResultColumnType = ColumnConst<InternalT>;
};
```

So we'd define `BinaryFunctionResultImpl<ColumnX<T>>` for any `X` we need (const, vector, mask, set, etc.) This is more extensible as the current logic does not scale/work for anything other than Const and Vector

storage/columns/Functions.h
```diff
@@ -252,4 +218,63 @@ struct FunctionExecutor<EdgeTypesFunction, Res, Arg> {
...
+        const size_t size = lhs->size();
+
+        res->resize(size);
+        auto op = Op {};
+
+        const auto& lhsRaw = lhs->getRaw();
+        const auto& rhsRaw = rhs->getRaw();
+        auto& resRaw = res->getRaw();
+        for (size_t i = 0; i < size; i++) {
+            resRaw[i] = op(TypeUtils::unwrap(lhsRaw[i]), TypeUtils::unwrap(rhsRaw[i]));
```
cyrusknopf: Why is this calling `unwrap`?

storage/versioning/CommitWriteBuffer.cpp
```diff
@@ -105,8 +105,9 @@ void CommitWriteBuffer::buildPendingNode(DataPartBuilder& builder,
                         nodeID, id, std::forward<decltype(val)>(val));
                 } else if constexpr (std::is_same_v<T, types::Bool::Primitive>) {
                     builder.addNodeProperty<types::Bool>(nodeID, id, val);
-                } else if constexpr (std::is_same_v<T, types::Embedding::Primitive>) {
-                    builder.addNodeProperty<types::Embedding>(nodeID, id, val);
+                } else if constexpr (std::is_same_v<T, std::vector<float>>) {
```
cyrusknopf: Can we define something like `types::Emebdding::OwningPrimitive = std::vector<float>`? I'll need it a lot in the write buffer/set command as well

query/parser/ParserUtils.cpp
rjb32: You're right it should rather be a reinterpret_cast but this isn't much prettier

query/AST/Literal.cpp
rjb32: I carefully considered it but the other option would be to copy the vector over?
To me it looked like a good case for an exception to the rule

query/parser/ParserUtils.cpp
cyrusknopf: I'm talking about having an integer literal that exceeds the float max. Reinterpret cast wouldn't help there

query/AST/Literal.cpp
cyrusknopf: Yep I agree

query/parser/ParserUtils.cpp
rjb32: Fixed by comparing explicitly against the integer part of float max

regress/embedding_properties/main.py
rjb32: Fixed

storage/versioning/CommitWriteBuffer.cpp
rjb32: Yes done

rjb32:
Just found a more robust say to check if an int64_t value fits into a float. It is not just a question of comparison the integer part to float max as there are non-float representable values of int64_t in that range.

PR #533, opened 2026-03-23
Fix critical rebase bugs

cyrusknopf:
- {LabelSet, Label, PropertyType}Maps were std::moved unconditionally in MetadataRebaser::rebase, but ptrs were only updated conditionally on the contents of the maps changing
- In edges were not rebased
- In edges conflict checking was double iterated

### DIFF
--- a/storage/versioning/ChangeConflictChecker.cpp
+++ b/storage/versioning/ChangeConflictChecker.cpp
@@ -127,8 +127,6 @@ void ChangeConflictChecker::checkNewEdgesIncidentToDeleted(const CommitData& lat
                     edgeID, record._nodeID);
                 throw VersionControlException(std::move(errorMsg));
             }
-
-            inEdgesIt.next();
         }
     }
 }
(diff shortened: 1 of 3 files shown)


storage/versioning/ChangeConflictChecker.cpp
```diff
@@ -127,8 +127,6 @@ void ChangeConflictChecker::checkNewEdgesIncidentToDeleted(const CommitData& lat
                     edgeID, record._nodeID);
                 throw VersionControlException(std::move(errorMsg));
             }
-
-            inEdgesIt.next();
```
cyrusknopf, line 131: This was causing double increments of the iterator

storage/versioning/DataPartRebaser.cpp
```diff
@@ -75,19 +72,16 @@ bool DataPartRebaser::rebase(const MetadataRebaser& metadata,
...
         }
-    } else if (_nodeOffset != 0) {
-        for (auto& e : edges->_outEdges) {
+
+        for (auto& e : edges->_inEdges) {
             e._nodeID = _idRebaser->rebaseNodeID(e._nodeID);
             e._otherID = _idRebaser->rebaseNodeID(e._otherID);
-        }
-    } else if (_edgeOffset != 0) {
-        for (auto& e : edges->_outEdges) {
```
cyrusknopf, line 90: Complicated `_XOffset` logic was not consistent with new `idRebaser` logic

Issue #485, opened 2026-03-05
LIMIT 0 doesn't work if the dataframe contains only ColumnConst columns


LucLabarriere:
## Test Details
- Name: `return-literal-limit`
- URL: http://localhost:5555/?test=return-literal-limit
- Enabled: false
- Write required: false
- Tags: projection, limit, literal, return

## Query
```cypher
RETURN 5 LIMIT 0
```

## Last Run
- Plan matched: true
- Result matched: true
- Result JSON matched: true
- Time: 72 μs

## Problem

LIMIT 0 doesn't work if the dataframe contains only ColumnConst columns (no way to specify the dataframe row count).

rjb32:
Solved in main in #537

PR #537, opened 2026-03-23
Fix RETURN 5 LIMIT 0

rjb32:
Fixes LIMIT 0 issue in #485

### DIFF
--- a/storage/columns/ColumnConst.h
+++ b/storage/columns/ColumnConst.h
@@ -6,6 +6,7 @@
 #include "DebugDump.h"
 #include "BioAssert.h"
 #include "NameOf.h"
+#include "FatalException.h"
 
 namespace db {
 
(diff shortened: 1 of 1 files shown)


query/pipeline/processors/LimitProcessor.cpp
```diff
@@ -67,6 +67,9 @@ void LimitProcessor::execute() {
         }
 
         _currentRowCount += rowsToWrite;
-        _output.getPort()->writeData();
+
+        if (_currentRowCount > 0) {
+            _output.getPort()->writeData();
+        }
```
cyrusknopf: It feels like this processor is not following the "emit at most one empty chunk, only to allow the next processor to proceed" guideline that we said.

cyrusknopf:
Actually I'm not sure how this fix works? Surely with `LIMIT 0` this branch always triggers:

~~~cpp
if (_reachedLimit) {
    // We have reached the limit, do nothing
    return;
} else {
    ...
}
~~~

rjb32:
I changed the approach, given that we need to emit one chunk at the end for the time being, with the way the pipeline works currently.

* ColumnConst was returning a size of one every time, preventing an empty chunk with at least a ColumnConst in the dataframe to ever be produced
* We can't just emit nothing at the end due to the way the pipeline works
* I didn't want to change LimitProcessor in a complex and subtle way
* Actually ColumnConst was still shown as size one even if we do assign with rowCount=0

-> So it hinted at being a ColumnConst problem that needs to have an _empty flag when it has never been assigned or when it is assigned with "zero rows" and return size zero to have an "empty" state.

PR #538, opened 2026-03-23
Optimize NodeIDs used as seed into a ConstScan

rjb32:
In many applications we may use NodeIDs as seeds for multihop queries with filtering using a potentially very big OR predicate:
```cypher
MATCH (n)-->(m)-->(p) WHERE n = 10 OR n = 11 OR n = 12 OR ... RETURN n, m, p
```  

It has been observed to be very costly on Reactome when there are ~850 seed NodeIDs for doing graph sampling with a 2 hops expansion from these seeds, as the equality predicate will be evaluated for all the nodes of reactome.

So we implement in this PR an PlanGraph optimisation to write ScanNode with a predicate consisting exclusively of equalities on NodeIDs into a ConstScan, which takes directly a column of NodeIDs known at plan time. This avoids the need of doing a scan and of filtering afterwards.

For 2-hop expansion of 838 seeds in scripts/bench_const_scan.py

| | Server time | Wall time |
|---|---|---|
| **Without** ConstScan | 725.78 ms | 782.16 ms |
| **With** ConstScan | 23.08 ms | 66.98 ms |
| **Speedup** | **31.4x** | **11.7x** |

<table><tr><td>Before</td><td>After</td></tr></table>

### DIFF
--- a/test/query/queries/ConstScanOptTest.cpp
+++ b/test/query/queries/ConstScanOptTest.cpp
@@ -47,12 +47,21 @@ public:
     }
 
     void execute(std::string_view cypher, auto callback) {
+        executeImpl(cypher, db::ChangeID::head(), callback);
+    }
+
+    void execute(std::string_view cypher, db::ChangeID change, auto callback) {
+        executeImpl(cypher, change, callback);
+    }
+
+private:
+    void executeImpl(std::string_view cypher, db::ChangeID change, auto callback) {
         _procedures = db::ProcedureManager::create();
         _procedures->init();
 
         auto txRes = _sysMan->openTransaction(_graphName,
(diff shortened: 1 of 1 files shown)


rjb32:
I miss the validation of NodeIDs here because for now it's just YOLO

query/pipeline/processors/ConstScanProcessor.cpp
cyrusknopf, line 80: I think it would be interesting to sort the input column, because sorted and contiguous node IDs work better as inputs to GetOutEdges, GetProperties, etc.

rjb32: Yes done

PR #539, opened 2026-03-24
Fix segfault with `RETURN COUNT(*)`

cyrusknopf:
`COUNT` always expected an input, but `*` did not trigger an `EvalExprProcessor` to be inserted before, like it would with other literals, e.g. `COUNT(5)`.

Add a check that `COUNT` has an input, and error otherwise.


query/interpreter/QueryInterpreterV2.cpp
```diff
@@ -182,6 +183,10 @@ void QueryInterpreterV2::executeImpl(const InterpreterContext& ctxt,
         status.setStatus(QueryStatus::Status::PLAN_ERROR);
         status.setMessage(e.what());
         return;
+    } catch (const TuringException& e) {
```
ghost: Since CompilerException inherits from TuringException can't we just have one code path if we don't handle them differently?

cyrusknopf: Do you mean use a CompilerException instead of a TuringException? Or remove this new `catch (const TuringException&` clause?

rjb32: That new catch block will catch also FatalException which will not be "unexpected" anymore

query/pipeline/PipelineBuilder.cpp
```diff
@@ -439,6 +443,10 @@ PipelineBlockOutputInterface& PipelineBuilder::addLimit(size_t count) {
 }
 
 PipelineValueOutputInterface& PipelineBuilder::addCount(ColumnTag colTag) {
+    if (!_pendingOutput.getInterface()) {
+        throw TuringException("COUNT had no input.");
```
rjb32: This is legal cypher actually. It should return zero.

cyrusknopf: It returns 1 as per neo4j, but this requires a lot more wiring to get it to be consistent with neo4j: I would need to intercept at some point - either PlanGraph or PipelineGenerator - detecting that COUNT(*) is done with no variables defined, and then insert a lambda processor to create a dataframe with a single column const (or similar to achieve the COUNT = 1). I decided that this is a useless feature and that it is not worth it right now since it requires a lot of wiring. I can create an issue to be in line with neo4j but this PR was just designed to not segfault

rjb32: Ok to be supported later as this is just for the sake of compliance

PR #544, opened 2026-03-25
Add local copy of CommitWriteBuffer with rollback in WriteProcessor

cyrusknopf:
_This is more a proof of concept of a fix that needs to be added._

It may be possible that `WriteProcessor::execute` modifies the `CommitWriteBuffer` of the current change, and errors in the same cycle. This violates the atomicity of a single query. This PR adds a naïve fix of keeping a copy of the `CommitWriteBuffer` on `prepare`, and rolling back to the local copy before throwing any exceptions to be caught by the `QueryInterpreter`.

Implemented in this PR by using `Result` types in the `WriteProcessor` instead of exceptions. This means we have minimal `try ... catch` blocks, and that errors can be propagated up to a single point, where the rollback occurs if it needs to, and then finally an exception can be thrown to be caught by `QueryInterpreter`.

### DIFF
--- a/query/pipeline/processors/WriteProcessor.cpp
+++ b/query/pipeline/processors/WriteProcessor.cpp
@@ -1,6 +1,7 @@
 #include "WriteProcessor.h"
 
 #include <algorithm>
+#include <exception>
 #include <optional>
 #include <string_view>
 #include <utility>
@@ -48,9 +49,9 @@ bool isColumnConst(const Column* col) {
 }
 
 template <TypedInternalID IDT>
-void validateDeletions(const GraphReader reader, const ColumnVector<IDT>* col) {
+WriteResult validateDeletions(const GraphReader reader, const ColumnVector<IDT>* col) {
     if (!col) {
-        throw FatalException("Attempted to delete null column.");
+        return WriteError {"Attempted to delete null column."};
     }
 
     constexpr bool isNode = std::is_same_v<IDT, NodeID>;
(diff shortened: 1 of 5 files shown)


query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -179,7 +182,16 @@ void getUntypedProperties(const Column* valueColumn,
...
     ColumnVectorToUntypedProperties toVec(props, pid);
-    Dispatcher::dispatch(valueColumn, toVec);
+
+    try {
+        Dispatcher::dispatch(valueColumn, toVec);
+    } catch (const std::exception& e) {
+        return WriteError {e.what()};
+    } catch (...) {
+        return WriteError {"Unknown error."};
+    }
```
cyrusknopf, line 192: Unfortunately I think this `try ... catch` might be unavoidable, without adding a result type to the functor passed to the dispatcher

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -440,33 +467,33 @@ void WriteProcessor::createEdges(size_t numIters) {
...
 
         ColumnNodeIDs* tgtCol = nullptr;
         const ColumnTag tgtTag = edge._tgtTag;
         const bool tgtIsPending = !inDf || inDf->getColumn(tgtTag) == nullptr;
 
         if (!outDf->hasColumn(tgtTag)) {
-            throw FatalException(fmt::format("Attempted to create edge {} with "
-                                             "target node with no such column: {}",
-                                             edge._name, edge._tgtTag.getValue()));
+            return WriteError {fmt::format("Attempted to create edge {} with "
```
rjb32, line 487: Please use the construction style with parenthesis

query/pipeline/processors/WriteProcessor.cpp
```diff
@@ -598,7 +627,11 @@ void WriteProcessor::updateNodes() {
                 _writeBuffer->addNodeUpdate(n, propBuffer);
             }
         } else { // Column(Opt)Vector -> unique property for each node
-            getUntypedProperties(valueColumn, propsBuffer, pid);
+            WriteResult res = getUntypedProperties(valueColumn, propsBuffer, pid);
```
rjb32, line 630: Can res be const?

PR #568, opened 2026-04-02
Add checks for null columns in EdgeTypesProcedure, LabelsProcedure, PropertyTypesProcedure

### LINKED ISSUE #551: `db.edgeTypes()` has inconsistent column checking
~~~cpp
void EdgeTypesProcedure::execute(ProcedureState* proc) {
    ...
    Column* rawIdsCol = data.getReturnColumn(0);
    Column* rawNamesCol = data.getReturnColumn(1);

    auto* idsCol = static_cast<ColumnVector<EdgeTypeID>*>(rawIdsCol);
    auto* namesCol = static_cast<ColumnVector<std::string_view>*>(rawNamesCol);

    switch (proc->getStep()) {
        case ProcedureState::Step::PREPARE: {
            data._it = std::make_unique<ScanEdgeTypesChunkWriter>(
                view.metadata().edgeTypes());
            data._it->setIDs(idsCol);
            data._it->setNames(namesCol);
        }
    ...
}
~~~
does not appear to check that its column is non-null, but other procedures do. They potentially may be null if the column is never `YIELD`ed.

rjb32:
Add checks for columns being potentially nullptr in procedures.
Fixes #551

### DIFF
--- a/query/pipeline/procedures/EdgeTypesProcedure.cpp
+++ b/query/pipeline/procedures/EdgeTypesProcedure.cpp
@@ -51,10 +51,15 @@ void EdgeTypesProcedure::execute(ProcedureState* proc) {
 
     switch (proc->getStep()) {
         case ProcedureState::Step::PREPARE: {
-            data._it = std::make_unique<ScanEdgeTypesChunkWriter>(
-                view.metadata().edgeTypes());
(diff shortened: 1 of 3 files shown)


cyrusknopf (approved):
Thanks

PR #571, opened 2026-04-08
Add the `DROP INDEX` query

cyrusknopf:
Allows for dropping indexes from subsequent commits; specifying the index to drop by its name.

Usage:
~~~
DROP INDEX <index name>
~~~

Indexes to drop are added to the `CommitWriteBuffer`. When a `CommitWriteBuffer` is flushed, the vector of valid indexes is filtered to remove those which are to be dropped.

### DIFF
--- a/query/AST/DropIndexQuery.h
+++ b/query/AST/DropIndexQuery.h
@@ -17,6 +17,8 @@ public:
 
 private:
     DropIndexQuery(DeclContext* declCtxt, std::string_view indexName);
+    ~DropIndexQuery() final;
+
     std::string_view _indexName;
 };
 
--- a/query/AST/DropIndexQuery.h
+++ b/query/AST/DropIndexQuery.h
@@ -1,0 +1,23 @@
+#pragma once
+
+#include <string_view>
+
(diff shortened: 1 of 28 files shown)


query/AST/DropIndexQuery.h
```diff
@@ -0,0 +1,23 @@
...
+class DropIndexQuery final : public QueryCommand {
+public:
+    static DropIndexQuery* create(CypherAST* ast, std::string_view indexName);
+
+    std::string_view indexName() const { return _indexName; }
+
+    Kind getKind() const final { return QueryCommand::Kind::DROP_INDEX_QUERY; }
+
+private:
+    DropIndexQuery(DeclContext* declCtxt, std::string_view indexName);
```
rjb32, line 19: The constructor is private but not the destructor?

cyrusknopf: A mistake on my part: dstrtor should also be private, I'll fix that

cyrusknopf: Fixed

rjb32 (approved):

PR #572, opened 2026-04-09
Add support for setting properties to NULL

cyrusknopf:
Adds an explicit store of entities which are null-valued in a PropertyContainer.

Modifies `GetPropertiesWithNullIterator` to use the new `tryGetWithNull` method of `Property{Container, Manager}` to handle explicit nulls.

Add tests to ensure correctness.

### DIFF
--- a/storage/properties/PropertyContainer.h
+++ b/storage/properties/PropertyContainer.h
@@ -59,8 +59,12 @@ public:
     IDs& ids() { return _ids; }
     const IDs& ids() const { return _ids; }
 
+    IDs& nullIds() { return _nullIds; }
+    const IDs& nullIds() const { return _nullIds; }
+
 protected:
     IDs _ids;
+    IDs _nullIds;
     std::unordered_map<EntityID, size_t> _entityIndexMap;
     static constexpr size_t NULL_INDEX = std::numeric_limits<size_t>::max();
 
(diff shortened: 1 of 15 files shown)


rjb32 (dismissed):

storage/properties/PropertyContainer.h
```diff
@@ -246,6 +304,26 @@ class TypedPropertyContainer<types::String> : public PropertyContainer {
         return &(*it);
     }
 
+    std::optional<const types::String::Primitive*> tryGetWithNull(EntityID entityID) const {
```
rjb32, line 318: We will need to fix the existing iterators because most of them don't use that interface yet.

storage/writers/DataPartBuilder.cpp
rjb32: Does not use with null

storage/properties/PropertyContainer.h
cyrusknopf: Are there iterators which needed changing other than `GetPropertiesWithNullIterator`?

rjb32: The scans that combines scan nodes + property scan need to be checked

cyrusknopf:
@rjb32 This PR should be at its final revision

cyrusknopf:
@rjb32 New convergence point

rjb32:
Do you have a cypher-level test for SET to null, such as:
```
MATCH (p:Person) WHERE p.age = 32
SET p.age = null
```

cyrusknopf:
> Do you have a cypher-level test for SET to null, such as:
> 
> ```
> MATCH (p:Person) WHERE p.age = 32
> SET p.age = null
> ```

Explicit `= null` is not supported in this PR. This PR only implements the ability for the storage layer to represent an explicitly null property.

The only way to see this today is, e.g:

```
MATCH (n) RETURN n.age // = 32, 32, null, null
```

```
MATCH (n) CREATE (m:Clone {age:n.age})
COMMIT
MATCH (m:Clone) RETURN m.age // = 32, 32, null, null <- new m.age is set to NULL
```

setting explicitly to null requires analyzer changes which I was going to do in a separate PR.

rjb32:
> > Do you have a cypher-level test for SET to null, such as:
> > ```
> > MATCH (p:Person) WHERE p.age = 32
> > SET p.age = null
> > ```
> 
> Explicit `= null` is not supported in this PR. This PR only implements the ability for the storage layer to represent an explicitly null property.
> 
> The only way to see this today is, e.g:
> 
> ```
> MATCH (n) RETURN n.age // = 32, 32, null, null
> ```
> 
> ```
> MATCH (n) CREATE (m:Clone {age:n.age})
> COMMIT
> MATCH (m:Clone) RETURN m.age // = 32, 32, null, null <- new m.age is set to NULL
> ```
> 
> setting explicitly to null requires analyzer changes which I was going to do in a separate PR.

Fair enough, thanks.

rjb32 (approved):

PR #590, opened 2026-04-15
Implement a `ListBuffer` for storage of list elements

cyrusknopf:
Implements `ListBuffer` to store list elements, and associated views into that container.


samples/list-buffer/ListBuffer.h
```diff
@@ -0,0 +1,102 @@
...
+
+template <typename T>
+concept Listable = TypeConcepts::InTuple<std::decay_t<T>, decltype(ListableTypes)>;
+
+class ListBufferElementView;
+class ListView;
+
+class ListBuffer {
+public:
+    using Storage = std::deque<std::byte>;
```
rjb32: Dangerous, let's implement our own ChunkedVector?

storage/ListByteBuffer.cpp
```diff
@@ -0,0 +1,100 @@
...
+    _last->_next = newChunk;
+    _last = newChunk;
+    return newChunk;
+}
+
+template <size_t N>
+void ListByteBuffer<N>::reserveContiguous(size_t numBytes) {
+    // Ensure we can fit this many bytes in a single buffer
+    const bool exceedsChunk = numBytes > N;
+    if (exceedsChunk) {
```
rjb32: That's the only part that could be dynamic in size because otherwise we may have an error reported to the user that they can't really do anything with. Or for us with big lists of seed nodes

storage/ListByteBuffer.cpp
```diff
@@ -0,0 +1,100 @@
...
+    const size_t startingIndex = _last->_size;
+
+    std::byte* const startPtr = &buf[startingIndex];
+
+    std::byte* writePtr = startPtr;
+    static_assert(_tagSize == 1);
+
+    { // Write the tag
+        const auto* tagAddr = &tag;
+        std::memcpy(writePtr, tagAddr, _tagSize);
```
rjb32: It's a bit overkill to use memcpy for writing a tag, you could use a pointer

storage/ListByteBuffer.cpp
cyrusknopf: Currently the buffer is represented as a `std::array`, so the size can't be dynamic. Admittedly this does limit lists to a maximum size as you say.

If we were to change the chunk size to be dynamic, how would you see it working? Would we have a default size which is used unless the call to `reserveContiguous` requests more than the default size?

storage/ListByteBuffer.cpp
cyrusknopf: The design was intended to be defensive against the size of a tag changing (from 1 byte to something potentially larger), but depending on if we think that is likely to happen I'm happy to remove this memcpy call

rjb32: I see. It is also defensive if you cast an std::byte* pointer to a (T*) pointer given that you do size checks before

storage/ListByteBuffer.cpp
rjb32: Yes exactly. We would have a chunk of a fixed size upon allocation but that size could be decided dynamically in reserveContiguous, so that the allocation always succeeds (except OOM of course)

storage/ListByteBuffer.cpp
cyrusknopf: Where in the code are you referring to?

rjb32: It is also defensive in a way to do:
```c++
MyTagType* writeTagPtr = (MyTagType*)writePtr;
*writeTagPtr = tag;
``` 
provided that you do before:
```c++
assert(sizeof(MyTagType) <= someRemainingSize);
```

cyrusknopf: Ah I see yes, I'll change it

storage/ListByteBuffer.cpp
cyrusknopf: I have made both the ListByteBuffer and the ListElementViewBuffer be able to allocate arbitrarily large chunks

rjb32 (approved):

PR #591, opened 2026-04-15
Implement MERGE_DATAPARTS query since v2

rjb32:
Implement a MERGE_DATAPARTS query that calls the merging of data parts manually.

It was previously existing in TuringDB but got removed since query pipeline v2.


query/pipeline/processors/DataPartMergeProcessor.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "Graph.h"
+#include "PipelineException.h"
+
+#include "Profiler.h"
+#include "BioAssert.h"
+
+using namespace db;
+
+DataPartMergeProcessor::DataPartMergeProcessor() {
+}
```
cyrusknopf:
```suggestion
DataPartMergeProcessor::DataPartMergeProcessor()
{
}
```

query/pipeline/processors/DataPartMergeProcessor.cpp
```diff
@@ -0,0 +1,72 @@
...
+    return proc;
+}
+
+void DataPartMergeProcessor::prepare(ExecutionContext* ctxt) {
+    _ctxt = ctxt;
+    markAsPrepared();
+}
+
+void DataPartMergeProcessor::reset() {
+}
```
cyrusknopf, line 46:
```suggestion
void DataPartMergeProcessor::reset() {
    markAsReset();
}
```

query/parser/CypherParser.y
```diff
@@ -658,6 +663,10 @@ commitQuery
     : COMMIT { $$ = CommitQuery::create(ast); LOC($$, @$); }
     ;
 
+mergeDataPartsQuery
+    : MERGE DATA PARTS { $$ = MergeDataPartsQuery::create(ast); LOC($$, @$); }
+    ;
```
cyrusknopf, line 667: Purely stylistic: do we consider "DataPart" as one or two words?

rjb32: Fixed, it's now MERGE_DATA_PARTS in one keyword

cyrusknopf: No I meant "DataPart" as a single word. So the command would be "MERGE DATAPARTS". I don't know what our convention is but I always assumed that "DataPart" is really one word, and not "Data Part"

rjb32: Indeed, it's now MERGE_DATAPARTS

query/parser/CypherLexer.l
```diff
@@ -104,6 +104,7 @@ using Token = db::YCypherParser::token;
 "LIMIT" { KEYWORD(LIMIT) }
 "YIELD" { KEYWORD(YIELD) }
 "MATCH" { KEYWORD(MATCH) }
+"MERGE_DATAPARTS" { KEYWORD(MERGE_DATAPARTS) }
```
sulaimansuhas, line 107: Super tiny nit - I always assumed this list was sorted by token length - so maybe this should be at the top.

query/pipeline/processors/DataPartMergeProcessor.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+void DataPartMergeProcessor::execute() {
+    Profile profile("DataPartMergeProcessor::execute");
+
+    SystemManager* sysMan = _ctxt->getSystemManager();
+    bioassert(sysMan, "DataPartMergeProcessor: SystemManager must be set");
+
+    const std::string graphName(_ctxt->getGraphName());
+    Graph* graph = graphName.empty() ? sysMan->getDefaultGraph()
```
sulaimansuhas, line 55: Is this necessary? Do we not assume in the pipeline that we always get a valid graph name?

rjb32: Fixed in main

query/parser/CypherLexer.l
rjb32: Fixed in main

PR #593, opened 2026-04-16
Do string escaping for all endpoints and unescaping for string literal

### LINKED ISSUE #592: JSON API responses contain invalid escape sequences
## Summary

The HTTP API returns JSON with invalid escape sequences in string property values, causing `JSON.parse()` to fail in clients.

### 1. Invalid `\'` escape sequence (77 occurrences in `shipping` graph)

`\'` is not a valid JSON escape. Only `\"`, `\\`, `\/`, `\b`, `\f`, `\n`, `\r`, `\t`, and `\uXXXX` are valid.

**Example from `GET /get_nodes`:**
```json
{"name":"Lloyd\'s of London"}
```

**Expected:**
```json
{"name":"Lloyd's of London"}
```

Single quotes do not need escaping in JSON strings.

### 2. Unescaped `"` inside string values

Double quotes inside property values are not escaped, producing structurally invalid JSON.

**Example from `GET /get_nodes`:**
```json
{"iso_equipment_name":"40ft high cube container (9'6\")"}
```

The `"` after `6` is unescaped and terminates the string prematurely, causing a parse error:
```
SyntaxError: Expected ',' or '}' after property value in JSON at position 138005
```

**Expected:**
```json
{"iso_equipment_name":"40ft high cube container (9'6\\\")"}
```

## Reproduction

```bash
# Start turingdb with the shipping graph loaded
turingdb start -load shipping

# Fetch all nodes — response fails JSON.parse() in JavaScript
curl -s -X POST 'http://localhost:6666/get_nodes?graph=shipping' \
  -H 'Content-Type: application/json' \
  -d '{"nodeIDs":[0,1,2,3,4,5,6,7,8,9]}' | node -e "
    let t=''; process.stdin.on('data',d=>t+=d); process.stdin.on('end',()=>{
      try { JSON.parse(t); console.log('OK') }
      catch(e) { console.log(e.message) }
    })"
```

With enough nodes (e.g. all 3209 in the shipping graph), the response hits a property containing `\'` or unescaped `"`, and `JSON.parse()` throws.

## Impact

The visualizer's `MATCH (n) RETURN n LIMIT 10000` query returns an empty canvas because `getNodes()` fails silently on JSON parse, and React Query retries in a loop.

## Workaround

Client-side sanitization before parsing: `text.replace(/\\'/g, "'")` handles the `\'` case, but the unescaped `"` requires a more complex fix.

rjb32:
* The /get_nodes endpoint returns invalid JSON when string properties contain single quotes (\') or unescaped double quotes. Creates a minimal regress test to reproduce.
* Escaping strings in all REST endpoints
* Unescape strings that may be escaped during query parsing when creating StringLiteral

Fixes #592

### DIFF
--- a/query/AST/Literal.cpp
+++ b/query/AST/Literal.cpp
@@ -28,6 +28,29 @@ DoubleLiteral* DoubleLiteral::create(CypherAST* ast, double value) {
     return literal;
 }
 
+StringLiteral::StringLiteral(std::string_view raw) {
+    _value.reserve(raw.size());
+
+    for (size_t i = 0; i < raw.size(); i++) {
+        if (raw[i] == '\\' && i + 1 < raw.size()) {
+            char next = raw[i + 1];
+            switch (next) {
+                case '\\': _value.push_back('\\'); i++; break;
+                case '\'': _value.push_back('\''); i++; break;
+                case '"':  _value.push_back('"');  i++; break;
+                case 't':  _value.push_back('\t'); i++; break;
+                case 'n':  _value.push_back('\n'); i++; break;
+                case 'r':  _value.push_back('\r'); i++; break;
+                case 'b':  _value.push_back('\b'); i++; break;
+                case 'f':  _value.push_back('\f'); i++; break;
+                default:   _value.push_back(raw[i]); break;
+            }
+        } else {
+            _value.push_back(raw[i]);
+        }
+    }
+}
+
 StringLiteral* StringLiteral::create(CypherAST* ast, std::string_view value) {
     StringLiteral* literal = new StringLiteral(value);
     ast->addLiteral(literal);
--- a/query/AST/Literal.cpp
+++ b/query/AST/Literal.cpp
@@ -138,7 +138,8 @@ DoubleLiteral* DoubleLiteral::create(CypherAST* ast, double value) {
     return literal;
 }
 
-StringLiteral::StringLiteral() {
+StringLiteral::StringLiteral()
+{
 }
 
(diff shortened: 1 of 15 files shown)


query/AST/Literal.cpp
```diff
@@ -28,6 +28,29 @@ DoubleLiteral* DoubleLiteral::create(CypherAST* ast, double value) {
     return literal;
 }
 
+StringLiteral::StringLiteral(std::string_view raw) {
```
rjb32: Do this in the create function. Never do significant work in constructors

query/AST/Literal.cpp
```diff
@@ -28,8 +56,15 @@ DoubleLiteral* DoubleLiteral::create(CypherAST* ast, double value) {
     return literal;
 }
 
+StringLiteral::StringLiteral() {
+}
```
cyrusknopf:
```suggestion
StringLiteral::StringLiteral()
{
}
```

query/AST/Literal.cpp
```diff
@@ -4,6 +4,34 @@
...
 
+namespace {
+
+void unescapeString(std::string_view raw, std::string& out) {
+    out.clear();
+    out.reserve(raw.size());
+
+    for (size_t i = 0; i < raw.size(); i++) {
+        if (raw[i] == '\\' && i + 1 < raw.size()) {
+            char next = raw[i + 1];
```
sulaimansuhas: const here

cyrusknopf (approved):

sulaimansuhas (approved):

PR #599, opened 2026-04-17
Implement First Version Of Binary Protocol Client And Server

sulaimansuhas:
Run with  = `USE_TURING_PROTO=1` (sets the turing server to use the turing protocol over http, you shouldn't have to do this for the client instance)

connect from a client using `connect <address> <port>`
you can then `cd` to the graph of choice (as well all `checkout` to your preferred commit or change)

From when I last checked most if not all types that can be returned through the query language are returned.

The code is not in its final state (I have something I need to look over and maybe remove - especially the testing code). Feel free to review the code and test the functionality!


net/encoders/TuringProtoEncoder.h
```diff
@@ -0,0 +1,358 @@
...
+    ColumnHeaderWriter() = delete;
+    explicit ColumnHeaderWriter(net::proto::ColumnWireHeader& header)
+        : _header(header)
+    {
+    }
+
+    void writeColumnSchema(net::proto::ColumnInternalKind typeCode,
+                           net::proto::ColumnKind encoding) {
+        _header._typeCode = static_cast<uint32_t>(typeCode);
+        _header._encoding = static_cast<uint8_t>(encoding);
```
cyrusknopf: I would suggest using `std::to_underlying` instead of casting

net/encoders/TuringProtoEncoder.h
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    template <typename T>
+    void operator()(const ColumnVector<T>* col) {
+        if constexpr (std::is_same_v<T, NodeID>) {
+            writeColumnSchema(net::proto::ColumnInternalKind::NODE_ID, net::proto::ColumnKind::VECTOR);
+        } else if constexpr (std::is_same_v<T, EdgeID>) {
+            writeColumnSchema(net::proto::ColumnInternalKind::EDGE_ID, net::proto::ColumnKind::VECTOR);
+        } else if constexpr (std::is_same_v<T, EdgeTypeID>) {
```
cyrusknopf: What was the decision behind having a single `writeColumnSchema` function which takes the container kind as the second argument, instead of having `writeColumnVectorSchema`, etc..? It seems theres a lot of unneeded duplication in the seocnd argument.

Also it may be more maintainable to have a struct something like

~~~cpp
template <typename T>
struct TypeToColunmKind;

template <>
struct TypeToColunmKind<NodeID> {
   static constexpr auto enumValue = net::proto::ColumnInternalKind::NODE_ID;
};

... // other specialisations.

~~~

which means there is then

1. a single source of truth for the type -> enum mapping
2. you avoid the chain of `if constexpr` in all the `operator()` functions

which would make both each `operator()` overload a lot slimmer whilst also making it a lot harder for them to have discrepencies between them due to human error.

net/encoders/TuringProtoEncoder.h
```diff
@@ -0,0 +1,358 @@
...
+                _outBuf->copyFixedLenData(&columnByteSize, sizeof(columnByteSize));
+                _outBuf->copyVarLenData(val.data(), columnByteSize);
+            }
+        } else {
+            static_assert(std::is_trivially_copyable_v<T>,
+                          "TuringProtoEncoder only supports trivially copyable types and string");
+            const size_t columnByteSize = sizeof(T) * col->size();
+            _outBuf->copyVector<T>(col->data(), columnByteSize);
+        }
+    }
```
cyrusknopf: This is up to your taste but consider overloading each internal kind as its own overload, e.g.

~~~cpp
template <>
void operator()(const ColumnVector<types::Embedding::Primitive>* col) {
    ... // encode embedding
}

... // same for string, path, triv copyable, etc.
~~~

the trade off here is that we have more overloads to `operator()` but we gain clarity in the sense that each overload is very simple and focused (handles a single type, not all types with a constexpr switch inside.)

net/turing_proto_common/Codec.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "NetBuffer.h"
+#include "TuringProtoHeaders.h"
+#include "TuringProtoOutBuf.h"
+
+namespace net::proto {
+
+[[nodiscard]] constexpr size_t headerWireSize() {
+    return sizeof(MessageTypes) + sizeof(uint32_t);
+}
```
cyrusknopf: Can/should be `consteval`

net/encoders/TuringProtoEncoder.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    size_t chunkHeaderLen = sizeof(uint32_t);
+    for (const NamedColumn* namedCol : df->cols()) {
+        chunkHeaderLen += net::proto::ColumnWireHeader::wireSize();
+        chunkHeaderLen += namedCol->getName().size();
+    }
+
+    bioassert(chunkHeaderLen <= _outBuf->capacity(), "Dataframe schema exceeds maximum chunk size");
+
+    const uint32_t columnCount = static_cast<uint32_t>(df->cols().size());
```
cyrusknopf, line 27: Should this have a check to see if the size fits in a u32?

net/encoders/TuringProtoEncoder.cpp
```diff
@@ -0,0 +1,73 @@
...
+    _outBuf->copyFixedLenData(&columnCount, sizeof(columnCount));
+
+    for (const NamedColumn* namedCol : df->cols()) {
+        std::string_view colName = namedCol->getName();
+        net::proto::ColumnWireHeader columnHeader {
+            ._nameLen = static_cast<uint32_t>(colName.size()),
```
cyrusknopf, line 34: Possible need for magnitude check

net/encoders/TuringProtoEncoder.cpp
```diff
@@ -0,0 +1,73 @@
...
+            ._typeCode = 0,
+            ._encoding = static_cast<uint8_t>(net::proto::ColumnKind::VECTOR)
```
cyrusknopf: std::to_underlying

net/http_common/HTTPWriter.h
```diff
@@ -34,11 +35,9 @@ class NetWriter {
         EnumStringPair<Status::SendFailed, "Could not send data">,
         EnumStringPair<Status::SentZeroBytes, "Send zero bytes which is not supposed to happen">>;
 
-    explicit NetWriter(utils::DataSocket socket)
+    explicit HTTPWriter(int socket = 0)
```
cyrusknopf: What's the motivation for changing this to int

net/http_parser/HTTPParser.h
```diff
@@ -22,7 +22,77 @@ class HTTPParser : public AbstractHTTPParser {
...
+
+    void handleAnalyzeError(AnalyzeError error, AbstractTCPWriter& writer) override {
+        handleAnalyzeError(error, static_cast<HTTPWriter&>(writer));
+    }
+
+    void handleAnalyzeError(AnalyzeError error, HTTPWriter& httpWriter) {
+        switch (static_cast<HTTP::Error>(error)) {
+            case net::HTTP::Error::REQUEST_TOO_BIG: {
+                httpWriter.setFirstLine(net::HTTP::Status::CONTENT_TOO_LARGE);
+                break;
```
cyrusknopf:
```suggestion
            break;
```

net/tcp_server/TCPConnection.h
```diff
@@ -29,10 +29,13 @@ class TCPConnection {
 
     void setSocket(utils::DataSocket socket) {
         _socket = socket;
-        _writer.setSocket(socket);
+        if (_writer) {
+            _writer->setSocket(socket);
+        }
     }
```
cyrusknopf, line 35: Does it make sense to set socket if `_writer == nullptr`?

net/tcp_server/TCPConnection.h
```diff
@@ -42,19 +45,22 @@ class TCPConnection {
     bool isCloseRequired() const { return _closeRequired; }
     size_t getStorageIndex() const { return _storageIndex; }
     NetBuffer& getInputBuffer() { return _inputBuffer; }
-    NetWriter& getWriter() { return _writer; }
-    AbstractHTTPParser* getParser() { return _parser.get(); }
+    AbstractTCPWriter& getWriter() { return *_writer; }
```
cyrusknopf: Unless I'm missing something: the public API only has a function to dereference the `_writer` pointer, but not to check if it is valid. I think this should ideally be `tryGetWriter` which throws/returns an error if `_writer` is null, or at least we should have some function which checks if the writer is valid.

net/tcp_server/TCPConnectionStorage.h
```diff
@@ -19,13 +20,15 @@ class TCPConnectionStorage {
     TCPConnectionStorage& operator=(const TCPConnectionStorage&) = delete;
     TCPConnectionStorage& operator=(TCPConnectionStorage&&) = delete;
 
-    void initialize(CreateAbstractHTTPParserFunc&& createParser) {
+    void initialize(CreateAbstractTCPParserFunc&& createParser,
+                    CreateAbstractTCPWriterFunc&& createWriter) {
```
cyrusknopf: Why is this taken as rvalue ref

net/turing_proto_client/Decoders.h
```diff
@@ -0,0 +1,390 @@
...
+
+        T value {};
+        inBuf->readData(&value, sizeof(T));
+        (*typedCol)[rowIndex] = value;
+    }
+    return true;
+}
+
+// Optional vector: strings
+inline bool decodeOptVectorString(size_t& rowIndex,
```
cyrusknopf: What was the decision behind using different functions with different names instead of having a templated `decodeVector` function and then constraining different implemenations for different types?

net/turing_proto_client/Decoders.h
```diff
@@ -0,0 +1,390 @@
...
+
+        const bool hasValue = columnState->_bitMask.test(rowIndex);
+        if (!hasValue) {
+            inBuf->increaseReadOffset(sizeof(T));
+            continue;
+        }
+
+        T value {};
+        inBuf->readData(&value, sizeof(T));
+        (*typedCol)[rowIndex] = value;
```
cyrusknopf: What guarantees that `typedCol` is the appropriate size to be able to do this index with `[]`?

net/turing_proto_client/Decoders.h
```diff
@@ -0,0 +1,390 @@
...
+            inBuf->increaseReadOffset(sizeof(uint32_t));
+            continue;
+        }
+
+        uint32_t stringSize = 0;
+        inBuf->readData(&stringSize, sizeof(stringSize));
+
+        if (stringSize <= inBuf->readable()) {
+            //the whole string is in one buffer so let's allocate it in one go
+            (*typedCol)[rowIndex].emplace(inBuf->data() + inBuf->readOffset(), stringSize);
```
cyrusknopf: I think this should be split up onto a few lines

net/turing_proto_client/TuringClient.h
```diff
@@ -0,0 +1,71 @@
...
+
+	void recvAll(uint64_t recvLen);
+	ProtoHeader recvMsgHeader();
+
+	std::string _remoteAddress;
+	std::string _remotePort;
+	std::string _graphName {"default"};
+	db::CommitHash _commitHash {db::CommitHash::head()};
+	db::ChangeID _changeID {db::ChangeID::head()};
+	int _socket;
```
cyrusknopf:
```suggestion
	int _socket {-1};
```

net/turing_proto_client/TuringClient.h
```diff
@@ -0,0 +1,71 @@
...
+	db::LocalMemory* _localMem;
```
cyrusknopf:
```suggestion
	db::LocalMemory* _localMem {nullptr};
```

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+TuringClient::TuringClient(const std::string& remoteAddress,
+			   const std::string& remotePort,
+			   db::LocalMemory* localMem)
+	:_remoteAddress(remoteAddress),
+	_remotePort(remotePort),
+	_socket(-1),
+	_localMem(localMem),
+	_outBuf(1024ul * 1024),
+	_inBuf(1024ul * 1024)
```
cyrusknopf: Why not have these default constructed in the member declaration? Also is there any implications of automatically allocating such a buffer size instead of defaulting to 0 and then expanding if needed?

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+	if (resHeader._type != MessageTypes::IYI) {
+		throw TuringException("Invalid hello response type received from server");
+	}
+
+	if (resHeader._dataLen != sizeof(uint8_t)) {
+		throw TuringException("Invalid hello response payload size");
+	}
+
+	spdlog::info(_inBuf.dump());
```
cyrusknopf:
```suggestion
```

net/turing_proto_client/Decoders.h
```diff
@@ -0,0 +1,390 @@
...
+#include "TuringProtoInBuf.h"
+#include "columns/ColumnConst.h"
+#include "columns/ColumnOptVector.h"
+#include "columns/ColumnVector.h"
+
+namespace net::proto {
+
+// Optional vector: trivially copyable types
+template<typename T>
+inline bool decodeOptVectorTrivial(size_t& rowIndex,
```
cyrusknopf: Would it make sense to put these in a class? Seems like there is a lot of passing the same parameters in each function

server/TuringProtoServerProcessor.cpp
```diff
@@ -0,0 +1,166 @@
...
+                                                       net::TCPConnection& connection)
+    : _db(db),
+    _connection(connection)
+{
+}
+
+TuringProtoServerProcessor::~TuringProtoServerProcessor() = default;
+
+void TuringProtoServerProcessor::process(net::AbstractThreadContext* threadContext) {
+    _threadContext = static_cast<DBThreadContext*>(threadContext);
```
cyrusknopf, line 42: Any chance this could fail?

server/TuringProtoServerProcessor.cpp
```diff
@@ -0,0 +1,166 @@
...
+    auto& writer = _connection.getWriter<net::proto::TuringProtoWriter>();
+
+    try {
+        switch (parser.getHeader()._type) {
+            case net::proto::MessageTypes::NABER:
+                handleHello();
+                break;
+            case net::proto::MessageTypes::QUERY:
+                handleQuery();
+                break;
```
cyrusknopf:
```suggestion
            break;
            case net::proto::MessageTypes::QUERY:
                handleQuery();
            break;
```

server/TuringProtoServerProcessor.cpp
```diff
@@ -0,0 +1,166 @@
...
+    });
+
+    callbacks.setOnError([&] (const QueryStatus& status) {
+        if (writer.errorOccured()) {
+            return;
+        }
+        writer.writeError(&status);
+    });
+
+    spdlog::info("[server] handleQuery: query='{}' graph='{}'", transactionInfo.query, transactionInfo.graphName);
```
cyrusknopf:
```suggestion
```

server/TuringProtoServerProcessor.cpp
```diff
@@ -0,0 +1,166 @@
+#include "TuringProtoServerProcessor.h"
```
cyrusknopf, line 1: Think this file is quite pivotal and would benefit from some comments

query/pipeline/procedures/HistoryProcedure.cpp
```diff
@@ -98,7 +98,7 @@ void HistoryProcedure::registerProcedure(ProcedureNamespace* ns) {
 void HistoryProcedure::execute(ProcedureState* proc) {
     Data& data = proc->data<Data>();
     ExecutionContext* ctxt = proc->getContext();
-    const std::string& graphName = ctxt->getGraphName().data();
+    const std::string graphName(ctxt->getGraphName());
```
cyrusknopf: Why is it no longer a reference?

query/pipeline/procedures/DescribeCommitProcedure.cpp
```diff
@@ -138,7 +138,8 @@ void DescribeCommitProcedure::execute(ProcedureState* proc) {
             bioassert(rawCommitCol, "db.describeCommit: must be provided a commit hash");
 
             const SystemManager* sysMan = ctxt->getSystemManager();
-            const Graph* graph = sysMan->getGraph(ctxt->getGraphName().data());
+            const std::string graphName(ctxt->getGraphName());
+            const Graph* graph = sysMan->getGraph(graphName);
```
cyrusknopf: Can be reference?

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+		if (bytesRead < 0) {
+			throw TuringException(std::string("Failed to read response: ") + std::strerror(errno));
+		}
+
+		if (bytesRead == 0) {
+			throw TuringException("Connection closed before a response was received");
+		}
+
+		totalBytesRead += static_cast<size_t>(bytesRead);
```
cyrusknopf: Is this cast necessary?

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+	queryHeader.copyToBuffer(payload.data(), offset);
+	std::memcpy(payload.data() + offset, _graphName.data(), _graphName.length());
+	offset += _graphName.length();
+	std::memcpy(payload.data() + offset, query.data(), query.length());
+	offset += query.length();
+
+	// Testing hook: uncomment to append a stray byte to the QUERY payload and
+	// trip the server's "Incoming query payload size is inconsistent" protocol
+	// error. Verifies the PROTOCOL_ERROR round-trip end-to-end.
+	// payload.push_back('\x00');
```
cyrusknopf: Leftover?

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+	_embeddingBuffer.clear();
+	db::DataframeManager dfMan;
+	db::QueryStatus res;
+	std::vector<TuringProtoDecoder::DecodedColumnSchema> colSchemas;
+	db::Dataframe df;
+	ProtoHeader responseHeader = send();
+	TuringProtoDecoder decoder(_localMem, &dfMan, &_embeddingBuffer);
+	while(true) {
+		switch (responseHeader._type) {
```
cyrusknopf: Formatting of this switch is off

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+				{
+					callback(&df);
+					//Ideally we find a better place to reset this.
+					//the problem with resetting after the for loops
+					//is the code wont get hit if the last element is
+					//a buffer crossing item (this kind of seems like a code smell tbh)
+					for(auto& schema : colSchemas) {
+						schema._colState.reset();
+					}
+					decoder.reset();
```
cyrusknopf: Does the comment refer to finding a better place to reset the decoder? or the `_colState`?

net/turing_proto_client/TuringClient.h
```diff
@@ -0,0 +1,71 @@
...
+	void setRemoteAddress(const std::string& remoteAddress) { _remoteAddress = remoteAddress; }
+	void setRemotePort(const std::string& remotePort) { _remotePort = remotePort; }
+	void setGraphName(std::string graphName) { _graphName = std::move(graphName); }
+	void setCommitHash(db::CommitHash commitHash) { _commitHash = commitHash; }
+	void setChangeID(db::ChangeID changeID) { _changeID = changeID; }
+	bool isConnected() const { return _socket >= 0; }
+	const std::string& getRemoteAddress() const { return _remoteAddress; }
+	const std::string& getRemotePort() const { return _remotePort; }
+	EmbeddingBuffer& getEmbeddingBuffer() { return _embeddingBuffer; }
+	const std::string& getGraphName() const { return _graphName; }
```
cyrusknopf: Why not string views?

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+            return fn.template operator()<db::ChangeID>(encoding);
+        default:
+            break;
+    }
+
+    throw TuringException("Unsupported incoming column type");
+}
+
+struct MakeColumnFn {
+    db::LocalMemory* _localMem;
```
cyrusknopf:
```suggestion
    db::LocalMemory* _localMem {nullptr};
```

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    template <typename T>
+    db::Column* operator()(net::proto::ColumnKind encoding) const {
+        switch (encoding) {
+            case net::proto::ColumnKind::VECTOR: {
+                if constexpr (SupportedColumnVectorTypes<T>) {
+                    return static_cast<db::Column*>(_localMem->alloc<db::ColumnVector<T>>());
```
cyrusknopf: Are the casts necessary? Or will a `ColumnVector<T>*` bind to the `Column*` automatically due to return signature?

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+                throw TuringException("Unsupported incoming column kind");
+        }
+    }
+};
+
+struct CastColumnFn {
+    TuringProtoDecoder* _decoder;
+    db::Column* _column;
+    TuringProtoInBuf* _inBuf;
+    TuringProtoDecoder::DfColumnState* _columnState;
```
cyrusknopf: Initialise all to nullptr

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+    if (colState->_numRows == 0) {
+        if (inBuf->readable() < sizeof(colState->_numRows)) {
+            return false;
+        }
+        inBuf->readData(&colState->_numRows, sizeof(colState->_numRows));
+    }
+
+    if (_rowIndex == 0) {
+        col->clear();
+        col->resize(colState->_numRows);
```
cyrusknopf: Clear followed by resize is redundant, unless you must have all values default initialised.

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+        colState->_bitMask.resize(colState->_numRows);
+
+        const size_t bytesToCopy = std::min(inBuf->readable(), colState->_bitMask.byteSize());
+        inBuf->readData(colState->_bitMask.data(), bytesToCopy);
+
+        if (bytesToCopy != colState->_bitMask.byteSize()) {
+            _bufferState._start = reinterpret_cast<char*>(colState->_bitMask.data());
+            _bufferState._len = colState->_bitMask.byteSize();
+            _bufferState._offset = bytesToCopy;
+            return false;
```
cyrusknopf, line 201: I think this could benefit from some comments, not immediately clear what the logic is to me

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+            col->set(std::string(inBuf->data() + inBuf->readOffset(), stringSize));
+            inBuf->increaseReadOffset(stringSize);
+        } else {
+            std::string val(stringSize, '\0');
+            const size_t numBytesToRead = inBuf->readable();
+            inBuf->readData(val.data(), numBytesToRead);
+            _bufferState._start = val.data();
+            _bufferState._len = val.size();
+            _bufferState._offset = numBytesToRead;
+            col->set(std::move(val));
```
cyrusknopf: To me it looks like there is no rvalue reference overload for `ColumnConst::set`. I think it will perform a copy without it.

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+            return false;
```
cyrusknopf: I'm assuming this returns false because only a section of the string was read? Where does the remainder of the string get read?

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+        return decodeConstPath(inBuf, col, &_bufferState);
+    } else if constexpr (std::is_same_v<T, db::types::Embedding::Primitive>) {
+        return decodeConstEmbedding(inBuf, col, &_bufferState, _embeddingBuffer);
+    } else {
+        throw TuringException("unsupported constant column type received");
+    }
+}
+
+template <typename T>
+bool TuringProtoDecoder::decodeColumn(db::ColumnConst<std::optional<T>>* col, TuringProtoInBuf* inBuf) {
```
cyrusknopf: The control flow in this function is quite complicated. Do you think you could reduce nesting/complexity by rearranging branches to utilise early returns better?

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+    bioassert(df->size() == colSchemas.size(), "Dataframe should have the same number of columns as we have read from the packet");
+
+    while( _colIndex < df->size()) {
+        auto& schema = colSchemas[_colIndex];
+        auto* col = df->cols()[_colIndex]->getColumn();
+
+        bool completed = dispatchColumnType(
+            static_cast<net::proto::ColumnInternalKind>(schema._header._typeCode),
+            static_cast<net::proto::ColumnKind>(schema._header._encoding),
+            CastColumnFn {this, col, inBuf, &schema._colState});
```
cyrusknopf: Is static cast not overkill here? can you not just construct the `net::proto::` enums directly?

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+        if(!completed) {
+            break;
+        }
```
cyrusknopf: A comment here to say what completed being false actually means would be useful

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+                                                   TuringProtoInBuf* inBuf,
+                                                   std::vector<DecodedColumnSchema>& colSchemas) {
+    bioassert(df, "decodeIncomingChunkHeader called with null dataframe");
+
+    uint32_t columnCount = 0;
+    inBuf->readData(&columnCount, sizeof(columnCount));
+
+    // Preallocate so emplace_back-induced reallocations can't invalidate pointers
+    // into schema._colName (the tag manager aliases those bytes via string_view).
+    colSchemas.resize(columnCount);
```
cyrusknopf: To which calls to `emplace_back` is this referring to? And by the wording of the comment I would have assumed this would be `reserve` not a `resize`

net/turing_proto_client/TuringProtoDecoder.h
```diff
@@ -0,0 +1,86 @@
...
+	bool decodeColumn(db::ColumnConst<std::optional<T>>* col, TuringProtoInBuf* inBuf);
+
+private:
+	db::LocalMemory* _localMem;
+	db::DataframeManager* _dfMan;
+	size_t _colIndex = 0;
+	size_t _rowIndex= 0;
+
+	InterruptedBufferState _bufferState;
+	EmbeddingBuffer* _embeddingBuffer;
```
cyrusknopf: nullptr initialisation; use brace initialisation

net/turing_proto_client/TuringProtoDecoderConcepts.h
```diff
@@ -0,0 +1,68 @@
...
+|| std::is_same_v<T, db::Path>
+|| std::is_same_v<T, std::string>;
+
+template<typename T>
+concept SupportedColumnOptVectorTypes = std::is_same_v<T, db::types::UInt64::Primitive>
+|| std::is_same_v<T, db::types::Int64::Primitive>
+|| std::is_same_v<T, db::types::Double::Primitive>
+|| std::is_same_v<T, db::types::Bool::Primitive>
+|| std::is_same_v<T, db::types::Embedding::Primitive>
+|| std::is_same_v<T, std::string>;
```
cyrusknopf, line 52: Why do we use non-owning types for embeddings, but owning types for strings?

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+        inBuf->readHeader(&(schema._header));
+
+        inBuf->ensureReadable(schema._header._nameLen);
+        schema._colName = std::string(inBuf->data() + inBuf->readOffset(), schema._header._nameLen);
+        inBuf->increaseReadOffset(schema._header._nameLen);
+
+        auto* column = dispatchColumnType(
+            static_cast<net::proto::ColumnInternalKind>(schema._header._typeCode),
+            static_cast<net::proto::ColumnKind>(schema._header._encoding),
+            MakeColumnFn {_localMem});
```
cyrusknopf, line 275: Same static cast remark

net/turing_proto_common/EmbeddingBuffer.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <stddef.h>
+
+#include <span>
+#include <vector>
+
+namespace net::proto {
+
+class EmbeddingBuffer {
```
cyrusknopf, line 22: From what I can see this class is designed to give stable references to spans of embeddings. However that intent of reference stability is not declared anywhere, and I thnk it needs documenting otherwise this would be easy to misuse and potentially break all spans

net/turing_proto_common/TuringProtoHeaders.h
```diff
@@ -0,0 +1,135 @@
...
+	END_CHUNK,
+	END,
+	ERROR,
+	PROTOCOL_ERROR,
+	_SIZE
+};
+
+struct ProtoHeader {
+	MessageTypes _type;
+	uint32_t _dataLen;
```
cyrusknopf: default initialise

net/turing_proto_common/TuringProtoHeaders.h
```diff
@@ -0,0 +1,135 @@
...
+
+	static constexpr size_t wireSize() { return sizeof(_type) + sizeof(_dataLen); }
```
cyrusknopf: consteval

net/turing_proto_common/TuringProtoHeaders.h
```diff
@@ -0,0 +1,135 @@
...
+	OPTIONAL_VECTOR,
+	CONSTANT,
+	OPTIONAL_CONSTANT,
+};
+
+// Only exists for the first chunk.
+struct ColumnWireHeader {
+	uint32_t _nameLen;
+	uint32_t _typeCode;
+	uint8_t _encoding;
```
cyrusknopf: default initialise

net/turing_proto_common/TuringProtoOutBuf.h
```diff
@@ -0,0 +1,79 @@
...
+namespace net::proto {
+
+struct ProtoHeader;
+struct ColumnWireHeader;
+struct QueryWireHeader ;
+
+class TuringProtoOutBuf{
+public:
+
+    explicit TuringProtoOutBuf(std::size_t capacity);
```
cyrusknopf: This file uses std::size_t, but I think our convention is to use size_t

common/Bitmask.h
```diff
@@ -34,3 +37,220 @@ struct bitmask {
         return v;
     }
 };
+
+template<std::unsigned_integral Type, size_t Size>
+struct LargeBitMask {
```
rjb32: This is becoming large, please make it a class

common/Bitmask.h
```diff
@@ -34,3 +37,220 @@ struct bitmask {
...
+
+    template <testable... Flags>
+    static constexpr auto create(Flags... flags) {
+        constexpr size_t count = sizeof...(Flags);
+        static_assert(count <= Size);
+
+        LargeBitMask mask;
+        size_t i = 0;
+        ((mask.set(i++, static_cast<bool>(flags))), ...);
+        return mask;
```
rjb32: Don't return the mask by value but use a result reference

common/Bitmask.h
```diff
@@ -34,3 +37,220 @@ struct bitmask {
         return v;
     }
 };
+
+template<std::unsigned_integral Type, size_t Size>
+struct LargeBitMask {
+    static constexpr size_t bitsPerWord = sizeof(Type) * 8;
+    static constexpr size_t arraySize = (Size + bitsPerWord - 1) / bitsPerWord;
+
+    std::array<Type, arraySize> _v {};
```
rjb32: This should be private and we should have a getter returning a reference to it

common/Bitmask.h
```diff
@@ -34,3 +37,220 @@ struct bitmask {
...
+        ((mask.set(i++, static_cast<bool>(flags))), ...);
+        return mask;
+    }
+
+    template <size_t N>
+    static constexpr auto create(const std::array<bool, N>& flags) {
+        static_assert(N <= Size);
+
+        LargeBitMask mask;
+        for (size_t i = 0; i < N; ++i) {
```
rjb32: Can't we just do assignment or a copy using std::copy here?

net/encoders/TuringProtoEncoder.h
```diff
@@ -0,0 +1,358 @@
...
+    explicit TuringProtoEncoder(net::proto::TuringProtoOutBuf* outBuf);
+
+    void writeDataframeHeader(const Dataframe* df);
+    void writeDataframe(const Dataframe* df);
+    void writeError(const QueryStatus* status);
+    void writeProtocolError(std::string_view message);
+    void writeEnd(db::QueryCallbacks::ExecTimeMilliseconds milliseconds);
+
+private:
+
```
rjb32: Remove extra blank line

net/tcp_server/TCPConnection.h
```diff
@@ -42,19 +45,22 @@ class TCPConnection {
...
+    WriterT& getWriter() { return *static_cast<WriterT*>(_writer.get()); }
+
 private:
     utils::DataSocket _socket {0};
     TCPConnectionStorage* _storage {nullptr};
     size_t _storageIndex {0};
     NetBuffer _inputBuffer;
-    NetWriter _writer {_socket};
-    std::unique_ptr<AbstractHTTPParser> _parser {nullptr};
+    std::unique_ptr<AbstractTCPWriter> _writer {nullptr};
```
rjb32: Not necessary to initialise explicitly to nullptr for unique pointers. They are default initialised to nullptr already

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+#include "TuringException.h"
+#include "dataframe/Dataframe.h"
+#include "dataframe/DataframeManager.h"
+
+#include "BioAssert.h"
+#include "spdlog/spdlog.h"
+
+using namespace net::proto;
+
+TuringClient::TuringClient(const std::string& remoteAddress,
```
rjb32, line 21: Fix alignment of parameters

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+	if (response > 1) {
+		throw TuringException("Invalid hello response payload value");
+	}
+
+	return response != 0;
+}
+
+ProtoHeader TuringClient::sendHello() {
+	uint8_t protocolVersion = 1;
```
rjb32: Can these be const?

net/turing_proto_client/TuringClient.cpp
```diff
@@ -0,0 +1,362 @@
...
+		return;
+	}
+
+	::freeaddrinfo(result);
+	throw TuringException(std::string("Failed to connect to ")
+		       + _remoteAddress + ":" + _remotePort + ": "
+		       + std::strerror(lastErrno));
+}
+
+db::QueryStatus TuringClient::sendQuery(const std::string& query,
```
rjb32, line 228: Align parameters

net/turing_proto_client/TuringProtoDecoder.cpp
```diff
@@ -0,0 +1,398 @@
...
+    template <typename T>
+    db::Column* operator()(net::proto::ColumnKind encoding) const {
+        switch (encoding) {
+            case net::proto::ColumnKind::VECTOR: {
+                if constexpr (SupportedColumnVectorTypes<T>) {
+                    return static_cast<db::Column*>(_localMem->alloc<db::ColumnVector<T>>());
+                } else {
+                    throw TuringException("Unsupported internal kind for type:Vector");
+                }
+            }
```
rjb32, line 85: Better to delimit these cases with break statements

net/encoders/TuringProtoEncoder.h
sulaimansuhas: Once I refactored to have some sort of InternalKindToEnum map (that was a good idea!) the need to have separate writeColumnVectorSchema functions seems a bit unnecessary.

net/http_common/HTTPWriter.h
sulaimansuhas: not sure think it was claude that this - good catch!

net/tcp_server/TCPConnection.h
sulaimansuhas: I think we shouldn't make much assumptions about what the higher level protocol is implementing here. So I don't think we can definitively say it is wrong to set the _socket if `_writer == nullptr`.

sulaimansuhas: hmm actually - after reading some more code and the next comment - I realise we do assume the existence of the _writer.

net/turing_proto_client/Decoders.h
sulaimansuhas: we resize the column to columnRows->_numRows in TuringProtoDecoder.cpp

net/turing_proto_client/TuringClient.cpp
sulaimansuhas: Good catch on having the default construction in the member declaration. But I don't  see what is the need now to default to 0 and then expand as needed - these buffers should be fixed size throughout their existence.

net/tcp_server/TCPConnection.h
cyrusknopf: I'm not sure that assuming a pointer to be null counts as "assuming what the higher level protocol is implementing". A pointer can be null no matter what the protocol implements

net/turing_proto_client/TuringClient.cpp
cyrusknopf: Ah okay if they are all fixed size throughout then all good. Probably should define the size as a constexpr static variable since the same size is used in multiple places (`_outBuf` and `_inBuf` at least)

query/pipeline/procedures/HistoryProcedure.cpp
sulaimansuhas: Constructing a string ref from char* creates a temporary string anyways. This is clearer in what is happening and helped me deal with a bug I had.

net/turing_proto_client/TuringClient.cpp
sulaimansuhas: leftover

net/turing_proto_client/TuringProtoDecoder.cpp
sulaimansuhas: Yes the decode functions return false if the decode wasn't completed and we are in an interrupted state. In that case we read from the packet directly to the bufferState in the `decodeIncomingChunk()` function.

net/turing_proto_client/TuringProtoDecoder.cpp
sulaimansuhas: I have moved this into the `decoders.h` file

net/turing_proto_client/TuringProtoDecoder.cpp
sulaimansuhas: think this is a leftover

net/turing_proto_client/TuringProtoDecoderConcepts.h
sulaimansuhas: For the simplicity of implementation I have only included owning strings for now. But in the future some sort of StringBuffer would be interesting to have as well.

net/turing_proto_common/EmbeddingBuffer.h
sulaimansuhas: I have added some documentation for this - but just to clarify it is just meant to be a simple (stable) owning class for the floats the embeddings reference. I looked at using the existing embedding buckets we have but they were a bit too complex for the use case - they are specific to embedding size if I am recalling correctly.

server/TuringProtoServerProcessor.cpp
sulaimansuhas: Not now - but this is not ideal code. It is something I have just copied from the existing http server implementation. This is something we should look into in the future.

net/turing_proto_client/TuringProtoDecoderConcepts.h
cyrusknopf: But why are the strings owning whilst embeddings non-owning? Was the embeddings buffer easier to implement?

sulaimansuhas: More so we had an existing String Owning type already - so I made use of that. I didn't want to add a new type of owning embeddings in this PR.

cyrusknopf: We do have `types::Embedding::OwningPrimitive`, in case you want to use that

common/Bitmask.h
```diff
@@ -34,3 +37,91 @@ struct bitmask {
...
 };
+
+template <std::unsigned_integral Type>
+class DynamicLargeBitMask {
+public:
+    static constexpr size_t bitsPerWord = sizeof(Type) * 8;
+
+    explicit DynamicLargeBitMask(size_t bitCount)
+        : _bitCount(bitCount),
+          _v((bitCount + bitsPerWord - 1) / bitsPerWord, 0) {
```
rjb32: Align with ":" in initialiser list

common/Bitmask.h
```diff
@@ -34,3 +37,91 @@ struct bitmask {
...
+    }
+
+    template <testable... Flags>
+    static void create(DynamicLargeBitMask& out, Flags... flags) {
+        out.resize(sizeof...(Flags));
+        size_t i = 0;
```
rjb32: What is going on here? This is a bit unclear

common/ProtocolException.h
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * @brief Exception class to denote a wire-protocol failure: malformed framing,
+ * bad handshake payload, unsupported message type, truncated query payload, etc.
+ * @detail Carries no extra state beyond the message string. Catching this
+ * distinctly from other TuringExceptions lets the server emit a PROTOCOL_ERROR
+ * packet to the client instead of silently closing the connection.
+ */
+class ProtocolException : public TuringException {
+public:
+    ProtocolException();
```
rjb32: Is the no-message constructor really necessary?

net/encoders/TuringProtoEncoder.h
```diff
@@ -0,0 +1,345 @@
...
+#include "Bitmask.h"
+#include "OutputValues.h"
+#include "TuringProtoOutBuf.h"
+#include "columns/ColumnVector.h"
+#include "dataframe/Dataframe.h"
+#include "metadata/PropertyType.h"
+#include "TuringProtoHeaders.h"
+#include "QueryCallbacks.h"
+
+using ProtoEncoderSupportedTypes = std::tuple<
```
rjb32, line 21: This should be put inside the db namespace

net/encoders/TuringProtoEncoder.h
```diff
@@ -0,0 +1,345 @@
...
+    db::Path,
+    db::EntityList,
+    db::PropertyNull,
+    db::ValueType>;
+
+namespace db {
+
+class QueryStatus;
+
+struct InternalKindToEnum {
```
rjb32: This struct will exist for the entire namespace so the name needs to be made more specific to this area

net/encoders/TuringProtoEncoder.h
```diff
@@ -0,0 +1,345 @@
...
+    db::TemplateCommitHash<0>,
+    db::TemplateCommitHash<1>,
+    size_t,
+    std::string,
+    db::Path,
+    db::EntityList,
+    db::PropertyNull,
+    db::ValueType>;
+
+namespace db {
```
rjb32: Why are we not actually using net::proto here?

storage/columns/ColumnConst.h
```diff
@@ -48,6 +51,7 @@ class ColumnConst : public Column {
     }
 
     size_t size() const override { return _empty ? 0 : 1; }
+    void clear() noexcept override { _empty = true; }
```
cyrusknopf: Why is this noexcept

storage/columns/ColumnConst.h
```diff
@@ -92,7 +97,9 @@ class ColumnConst : public Column {
     static constexpr auto _staticKind = ColumnKind::code<ColumnConst<T>>();
 };
 
+template <>
+ColumnConst<PropertyNull>::ColumnConst();
```
cyrusknopf, line 101: Is this needed for the specialisations defined in the .cpp? If not then should remove

net/turing_proto_client/Decoders.h
```diff
@@ -0,0 +1,525 @@
...
+
+template <>
+inline bool decodeVector<std::string>(TuringProtoDecoder::DecodeContext* ctx,
+                                      db::ColumnVector<std::string>* typedCol,
+                                      TuringProtoDecoder::DfColumnState* columnState) {
+    if (ctx->_rowIndex == 0) {
+        typedCol->reserve(columnState->_numRows);
+    }
+
+    for (; ctx->_rowIndex < columnState->_numRows; ++ctx->_rowIndex) {
```
rjb32: It is better style to use a local variable index for the loop instead on relying on something in a context structure directly

storage/columns/ColumnMask.h
```diff
@@ -123,7 +123,7 @@ class ColumnMask : public Column {
         _data.assign(otherStart, otherStart + rowCount);
     }
 
-    void clear() { _data.clear(); }
+    void clear() noexcept override { _data.clear(); }
```
cyrusknopf: What is the rationale behind marking this as noexcept? std::vector::clear already is noexcept

storage/columns/ColumnSet.h
```diff
@@ -74,7 +74,7 @@ class ColumnSet : public Column {
         bioassert(false, "ColumnSet::assignFromLine: not implemented for ColumnSet");
     }
 
-    void clear() noexcept { _data.clear(); }
+    void clear() noexcept override { _data.clear(); }
```
cyrusknopf: Same noexcept comment

storage/columns/ColumnVector.h
```diff
@@ -108,7 +108,7 @@ class ColumnVector : public Column {
         _data.assign(otherStart, otherStart + rowCount);
     }
 
-    void clear() { _data.clear(); }
+    void clear() noexcept override { _data.clear(); }
```
cyrusknopf: Same noexcept comment

net/turing_proto_common/EmbeddingBuffer.cpp
```diff
@@ -0,0 +1,39 @@
...
+using namespace net::proto;
+
+EmbeddingBuffer::EmbeddingBuffer() {
+}
+
+EmbeddingBuffer::~EmbeddingBuffer() {
+}
+
+std::vector<float>& EmbeddingBuffer::ensureCapacity(size_t count) {
+    if (_chunks.empty() || _chunks.back().size() + count > _chunks.back().capacity()) {
```
rjb32: Break the condition

storage/dataframe/Dataframe.cpp
```diff
@@ -16,6 +16,12 @@ Dataframe::Dataframe() = default;
 
 Dataframe::~Dataframe() = default;
 
+void Dataframe::clear() const noexcept {
+    for (const NamedColumn* ncols : _cols) {
+        ncols->getColumn()->clear();
+    }
+}
```
cyrusknopf: Marking this function as const is misleading. Also why is it noexcept

net/turing_proto_common/ProtocolException.h
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * @brief Exception class to denote a wire-protocol failure: malformed framing,
+ * bad handshake payload, unsupported message type, truncated query payload, etc.
+ * @detail Carries no extra state beyond the message string. Catching this
+ * distinctly from other TuringExceptions lets the server emit a PROTOCOL_ERROR
+ * packet to the client instead of silently closing the connection.
+ */
+class ProtocolException : public TuringException {
+public:
+    ProtocolException();
```
rjb32: There is already a ProtocolException header defined in common

net/turing_proto_common/TuringProtoHeaders.h
```diff
@@ -0,0 +1,154 @@
...
+    CHUNK,
+    END_CHUNK,
+    END,
+    ERROR,
+    PROTOCOL_ERROR,
+    _SIZE
+};
+
+struct ProtoHeader {
+    MessageTypes _type;
```
rjb32: Define default values

net/turing_proto_common/TuringProtoHeaders.h
```diff
@@ -0,0 +1,154 @@
...
+enum class ColumnKind : uint8_t {
+    VECTOR = 0,
+    OPTIONAL_VECTOR,
+    CONSTANT,
+    OPTIONAL_CONSTANT,
+};
+
+// Only exists for the first chunk.
+struct ColumnWireHeader {
+    uint32_t _nameLen;
```
rjb32: Define default values

net/turing_proto_common/TuringProtoHeaders.h
```diff
@@ -0,0 +1,154 @@
...
+        offset += sizeof(_encoding);
+    }
+};
+
+// Commit and change IDs are sent as raw uint64 values. By convention,
+// std::numeric_limits<uint64_t>::max() is the "head" sentinel, matching the
+// in-memory representation of TemplateCommitHash::head(). All other values
+// in [1, UINT64_MAX - 1] are valid IDs; 0 is reserved.
+struct QueryWireHeader {
+    uint64_t _commitHash;
```
rjb32: Define default values

rjb32 (approved):

net/turing_proto_common/TuringProtoOutBuf.cpp
```diff
@@ -0,0 +1,89 @@
...
+    copy(&header->_changeID, sizeof(header->_changeID));
+    copy(&header->_graphNameLen, sizeof(header->_graphNameLen));
+    copy(&header->_queryLen, sizeof(header->_queryLen));
+}
+
+void TuringProtoOutBuf::copyVarLenData(const void* data, size_t len) {
+    size_t offset = 0;
+    while (offset < len) {
+        const size_t toCopy = std::min(len - offset, remaining());
+        copy(static_cast<const char*>(data) + offset, toCopy);
```
cyrusknopf: What is this cast for?

net/turing_proto_common/HexDump.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace net::proto {
+
+[[nodiscard]] inline std::string hexDump(const char* data, size_t len) {
+    static constexpr char kHex[] = "0123456789abcdef";
+
+    std::string out;
+    out.reserve(len * 3);
+    for (size_t i = 0; i < len; ++i) {
+        const unsigned char byte = static_cast<unsigned char>(data[i]);
```
cyrusknopf, line 14: Does reinterpret cast make more sense here? I am not actually sure what the difference in behaviour/generated assembly would be

common/Bitmask.h
```diff
@@ -34,3 +37,91 @@ struct bitmask {
...
+    }
+
+    void resize(size_t bitCount) {
+        _bitCount = bitCount;
+        _v.assign((bitCount + bitsPerWord - 1) / bitsPerWord, 0);
+    }
+
+    void clear() {
+        std::fill(_v.begin(), _v.end(), static_cast<Type>(0));
+    }
```
cyrusknopf: I think it is generally good practice to use `std::vector::assign` here because the member methods have more context of the underlying container than an alien std::fill, for example

cyrusknopf: Also the name of this method might be a bit misleading, since you have `resize` right above one might assume that `clear` resets the size of the bitmask to 0, when actually it "resets" it.

common/Bitmask.h
```diff
@@ -1,9 +1,12 @@
...
+#include <optional>
 #include <stddef.h>
 #include <stdint.h>
 #include <array>
 #include <type_traits>
+#include <concepts>
+#include <vector>
 
 template <typename T>
 concept testable = std::is_integral_v<T>
```
cyrusknopf, line 12: `is_integral_v` is a superset of `is_same_v<bool>`, so you only need the former. What is the need for the is_pointer_v?

common/Bitmask.h
sulaimansuhas: actually I think this function is pretty impractical I will remove it for now

common/ProtocolException.h
sulaimansuhas: deleted

net/encoders/TuringProtoEncoder.h
sulaimansuhas: fixed - I realised I was accidentally using the db namespace here instead of the net namespace

storage/columns/ColumnConst.h
sulaimansuhas: It's needed because the callers need to know not to rely on the default template and to go look for the declaration of the specialisation.

storage/columns/ColumnSet.h
sulaimansuhas: will remove these as they seem a bit confusing

net/turing_proto_common/TuringProtoOutBuf.cpp
sulaimansuhas: data is a `void*` that we can't do pointer arithmetic on - so I needed to cast to char so I can increase the offset.

rjb32: It's clearer to use uint8_t* for this, or uintptr_t

storage/columns/ColumnConst.h
cyrusknopf: Ah cool thanks

net/turing_proto_common/TuringProtoOutBuf.cpp
cyrusknopf: Ah I see thanks

net/turing_proto_common/HexDump.h
sulaimansuhas: It doesn't make sense to use reinterpret_cast on non pointer values, I think it won't compile even.

net/http_common/HTTPWriter.h
```diff
@@ -34,11 +36,9 @@ class NetWriter {
         EnumStringPair<Status::SendFailed, "Could not send data">,
         EnumStringPair<Status::SentZeroBytes, "Send zero bytes which is not supposed to happen">>;
 
-    explicit NetWriter(utils::DataSocket socket)
+    explicit HTTPWriter(net::utils::DataSocket socket = 0)
```
rjb32: Why this socket is default initialized to zero?

rjb32 (approved):

PR #604, opened 2026-04-23
Implement `ListLiterals` in the query language

cyrusknopf:
This PR integrates the previously implemented `ListBuffer` into `LocalMemory`, allowing for the construction and use of list literals.

Replaces uses of `ListExpr` which should have been a literal type; replaced with the new `ListLiteral` subclass.

Allows for the use of literal lists, where all elements of the list are literals themselves.

Example use:

~~~
RETURN [10, true, "i love turingdb", 3.14]
~~~

~~~
MATCH (n) RETURN n, ["one list for each n", true]
~~~


memory/LocalMemory.h
```diff
@@ -117,11 +123,15 @@ class LocalMemory {
 
     void clear() {
         _pools.transform<ClearTransform>();
+        _listBuffer.clear();
```
cyrusknopf, line 126: @rjb32 @sulaimansuhas I would be interested in what you guys think of

a) the implementation of the `clear()` method; and 

b) the placement of the list buffer in `LocalMemory`

memory/LocalMemory.h
```diff
@@ -19,13 +19,18 @@
 #include "ID.h"
 #include "versioning/ChangeID.h"
 
+#include "ListBuffer.h"
```
sulaimansuhas: Maybe these includes belong above the metadata/ID/versioning includes?

query/analyzer/CypherAnalyzer.cpp
```diff
@@ -259,6 +262,13 @@ void CypherAnalyzer::analyze(const ReturnStmt* returnSt) {
 
         bioassert(!name.empty(), "All declared variable must have a name.");
 
+        // For exprs with explicit AS aliases, update their decl to be named
+        if (hasExplicitAlias) {
+            VarDecl* namedDecl =
+                _ctxt->getOrCreateNamedVariable(_ast, item->getType(), name);
+            item->setExprVarDecl(namedDecl);
+        }
+
```
sulaimansuhas, line 271: Is this relevant to this PR?

cyrusknopf: It is required for  `test/query-test-suite/tests/return-symbol-non-node-edge.json` to ensure that aliased list items are treated correctly.

memory/LocalMemory.h
```diff
@@ -19,13 +19,18 @@
...
+#include "ListView.h"
+
 namespace db {
 
 class CommitBuilder;
 class Change;
 
 class LocalMemory {
 public:
+    using DefaultListBuffer = ListBuffer<>;
```
rjb32, line 32: Is this really useful to define a type alias for the default list buffer?

storage/ListElementViewBuffer.cpp
```diff
@@ -51,6 +51,21 @@ void ListElementViewBuffer<N>::write(ListElementView view) {
...
+    // Delete all chunks
+    auto* cur = _first;
+    while (cur) {
+        auto* next = cur->_next;
+        delete cur;
+        cur = next;
+    }
+
+    // Create new first chunk
+    _first = new Chunk;
```
rjb32: Please use the new Chunk() style with parenthesis

memory/LocalMemory.h
rjb32: The clear() method seems good to me.
I don't have a strong opinion or a better idea for the placement of the list buffer right now. @sulaimansuhas any opinion?

rjb32 (approved):

memory/LocalMemory.h
cyrusknopf: I did it to avoid having to give empty template argument `<>`, but I can remove it if you prefer

rjb32: Ah fair enough ok, let's keep it

memory/LocalMemory.h
sulaimansuhas: I think list buffer placement seems appropriate for now!

memory/LocalMemory.h
cyrusknopf: Good point thanks

storage/ListElementViewBuffer.cpp
cyrusknopf: Fixed

PR #611, opened 2026-04-28
Analyzer/generator refactor; `ShortestPathStmt` made a read statement

cyrusknopf:
This PR migrates `ShortestPathStmt` as a more general `ReadStmt`, instead of having specialised logic for analysis/generation.

This slims down the `SinglePartQuery` class, as well as moving analysis/generation to `ReadStmtAnalyzer/Generator`.

This PR also slims the logic for adding a query limit by moving logic for setting the query limit to an anonymous function, instead of inline.

### DIFF
--- a/query/plan/PlanGraphGenerator.cpp
+++ b/query/plan/PlanGraphGenerator.cpp
@@ -22,9 +22,7 @@
 #include "nodes/CommitNode.h"
 #include "Literal.h"
 #include "stmt/Limit.h"
-#include "stmt/OrderBy.h"
 #include "stmt/ReturnStmt.h"
-#include "stmt/ShortestPathStmt.h"
 #include "stmt/Skip.h"
 #include "stmt/StmtContainer.h"
 #include "views/GraphView.h"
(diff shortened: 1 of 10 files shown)


query/plan/PlanGraphGenerator.cpp
```diff
@@ -252,30 +266,6 @@ void PlanGraphGenerator::generateSinglePartQuery(const SinglePartQuery* query) {
         // Place joins based on procedures calls
         readGenerator.placeJoinsOnProcedures();
 
-        // Insert ShortestPath Node
-        if (shortestPathStmt) {
```
cyrusknopf, line 256: @sulaimansuhas Do you think there is a potential problem with this change?

Currently, the shortest path is always generated *after* all read statements are generated (the if clause comes after the for loop), however now the shortest path statement is generated in the loop, with the other statements.

sulaimansuhas: If the read statements are processed in the order they are added. Doesn't the grammar in the parser guarantee that the shortest path statement is the last read stmt - so it will be processed after all the other read statements anyways?

cyrusknopf: Yes currently but if the grammar were to change, allowing the shortest path to not be last, could that cause problems? (Maybe undecidable for now)

rjb32:
I think it's a good idea in general. Let's approve if there are no more concerns

query/plan/PlanGraphGenerator.cpp
sulaimansuhas: It could cause problems if the statements that the shortest path statement needs to read from come after the shortest path statement. I will look into this but the PR looks good to me for now!

rjb32 (approved):

PR #614, opened 2026-05-01
System manager refactor

rjb32:
Refactoring of SystemManager:
* Move all subsystems and subcomponents to SystemManager. TuringDB is just an outer layer with the query function.
* Subsystems are instantiated by values in SystemManager, no more artificial unique_ptr
* Just use directly SystemManager pointer instead of passing a lot manager objects in contexts
* Move ProcedureManager code and procedures into their own turing_db_procedure_s library outside of pipeline

SystemManager is still ugly for now, this is a first step before the new accessor system.


db/TuringDB.h
```diff
@@ -26,30 +18,16 @@ class TuringDB {
 
     void init();
 
-    void stop();
-
     QueryStatus query(std::string_view query, const QueryState& state);
 
     const QueryConfig& getDefaultQueryConfig() const { return _defaultQueryConfig; }
```
cyrusknopf, line 23: Why is this getter/member called "default" query config, when presumably it is not the "default", but the config of the current TuringDB object, that passed in the constructor. I would've thought the "default" config would be a static member

jobs/JobSystem.cpp
```diff
@@ -13,18 +13,25 @@ namespace {
...
 constexpr const char* THREAD_NAME = "turingdb.worker";
 
+size_t getThreadCount(size_t requestedThreads) {
+    const size_t hardwareThreads = std::max<size_t>(1, std::thread::hardware_concurrency());
+    if (requestedThreads == 0) {
+        return hardwareThreads;
+    } else {
+        return requestedThreads;
+    }
+}
```
cyrusknopf, line 22: Whats the rationale behind returning the number of avaiable threads if `requestedThreads == 0`?  Also when can `std::thread::hardware_concurrency()` return something less than 1 (motivating the call to std::max)?

cyrusknopf (approved):

db/TuringDB.h
rjb32: So it is the default config of the queries in the sense that is the one that can be passed to QueryState and it contains some good defaults. It is not a static parameter because it probes the environment variables for activating or not some features.

jobs/JobSystem.cpp
rjb32: It is to be defensive on the number of requestedThreads. So if requestedThreads == 0 it means that the value is erroneous somehow but is possible because zero is in the range of size_t. But this is mostly historical.
I removed the part about taking the max on hardware_concurrency, it just means that the OS is deeply wrong anyway.

cyrusknopf: But I can see that it is not purely erroneous because you use `getThreadCount(0)` in the default constructor of `JobSystem`, so is it to say that historically 0 is used to return the maximum available?

rjb32: Yes indeed

PR #622, opened 2026-05-06
Fix dependency and versioning issues on older builds

cyrusknopf:
- Adds required include of `<sstream>`
- Removes use of C++23 monadic `std::optional<T>::and_then`

CI Full matrix passing: https://github.com/turing-db/turingdb/actions/runs/25428939392

### DIFF
--- a/query/plan/PipelineGenerator.cpp
+++ b/query/plan/PipelineGenerator.cpp
@@ -1948,8 +1948,8 @@ PipelineOutputInterface* PipelineGenerator::translateUnwindNode(UnwindNode* node
 
     const PipelineValuesOutputInterface* unwindOut {nullptr};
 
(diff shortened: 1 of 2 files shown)


query/plan/PipelineGenerator.cpp
```diff
@@ -1948,7 +1948,9 @@ PipelineOutputInterface* PipelineGenerator::translateUnwindNode(UnwindNode* node
 
     const PipelineValuesOutputInterface* unwindOut {nullptr};
 
-    std::optional<ValueType> maybeHomogeneity = node->homogeneity().and_then(toValueType);
+    std::optional<EvaluatedType> maybeEvalHomogeneity = node->homogeneity();
```
rjb32: Can this be const?

cyrusknopf: Fixed thanks

rjb32 (approved):

PR #628, opened 2026-05-07
Use dense bitmap for visited set of `PlanGraphTopology`

cyrusknopf:
`PlanGraphTopology` needs a set of visited nodes when exploring the `PlanGraph`. This set can be replaced with a dense bitmap now that each `PlanGraphNode` has a unique ID.

### DIFF
--- a/query/plan/PlanGraphTopology.h
+++ b/query/plan/PlanGraphTopology.h
@@ -1,16 +1,18 @@
 #pragma once
 
 #include <unordered_map>
-#include <unordered_set>
+
+#include "VisitedNodeSet.h"
 
 namespace db {
 
 class VarNode;
 class PlanGraphNode;
+class PlanGraph;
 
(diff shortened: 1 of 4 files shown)


query/plan/PlanGraphTopology.h
```diff
@@ -31,7 +33,8 @@ class PlanGraphTopology {
 
 private:
     /// Visited set used by the algorithms. Clear at the beginning of each algorithm
-    std::unordered_set<PlanGraphNode*> _visited;
+    VisitedNodeSet _visited;
+    const PlanGraph* _tree {nullptr};
```
rjb32: _tree should be the first parameter before the visited set

cyrusknopf: Fixed

rjb32 (approved):

PR #631, opened 2026-05-08
Fix invalid HashJoins

cyrusknopf:
- Update `PipelineBuilder::addHashJoin` to use new column dispatcher, which strictly enforces types declaratively in `AllowedKinds`.
- Perform plan-time type checking of join arguments to avoid invalid joins which slip past analysis stage

### DIFF
--- a/query/plan/nodes/JoinNode.cpp
+++ b/query/plan/nodes/JoinNode.cpp
@@ -1,0 +1,10 @@
+#include "JoinNode.h"
+#include "decl/EvaluatedType.h"
+
+using namespace db;
+
+// NOTE: This should be synced with @ref ValueHashJoinPairs
+bool JoinNode::joinableTypes(EvaluatedType a, EvaluatedType b) {
+    const bool same = a == b;
+    return same;
+}
--- a/storage/columns/BinaryPredicates.h
+++ b/storage/columns/BinaryPredicates.h
@@ -21,6 +21,10 @@ namespace {
 struct TuringEqual;
 struct TuringNotEqual;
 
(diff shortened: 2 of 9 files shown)


query/plan/nodes/JoinNode.cpp
```diff
@@ -0,0 +1,10 @@
+#include "JoinNode.h"
+#include "decl/EvaluatedType.h"
+
+using namespace db;
+
+// NOTE: This should be synced with @ref ValueHashJoinPairs
+bool JoinNode::joinableTypes(EvaluatedType a, EvaluatedType b) {
+    const bool same = a == b;
```
sulaimansuhas, line 8: why not just return a == b?

cyrusknopf: I'm making this change to be able to properly support `NodeID == int64`, for example, where both types may not be the same, so in future there will be more conditions which result in true. In the meantime this will be optimised down to the equivalent of `a == b`

storage/columns/BinaryPredicates.h
```diff
@@ -21,6 +21,10 @@ namespace {
 struct TuringEqual;
 struct TuringNotEqual;
 
+// Static storage duration sentinals to avoid reconstructing each iteration
+static constexpr CustomBool sentinalFalse(false);
```
sulaimansuhas: Spelling should be sentinel

sulaimansuhas (approved):

storage/columns/BinaryPredicates.h
cyrusknopf: Haha yes you're right

rjb32 (approved):

PR #633, opened 2026-05-11
Feature/binary protocol py sdk

sulaimansuhas:
This PR creates a python sdk for the binary client. We use nanobind to create a wrapper over an underlying C++ client. 

## Modifications to the regress tests
- Added the ability to set the sdk protocol type using an env var to avoid having to duplicate our code to switch between test types
- Removed redundant authentication code in the http client and regress tests
- Modified the interface of the general turingdb client to have a reconnect op to support the stateful nature of 
   the binary protocol


python/turingdb/_binary/PyTuringClient.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include <nanobind/nanobind.h>
```
rjb32: That include should come after the standard includes

python/turingdb/_binary/PyTuringClient.h
```diff
@@ -0,0 +1,49 @@
...
+#include <string>
+
+#include "TuringClient.h"
+
+namespace db {
+class LocalMemory;
+}
+
+namespace net::proto {
+class TuringClient;
```
rjb32: What's the point of forward declaring TuringClient here if you include the header anyway?

python/turingdb/_binary/NanobindUtils.h
```diff
@@ -0,0 +1,192 @@
+#pragma once
+
+#include <nanobind/nanobind.h>
```
rjb32: These includes should come after standard includes

python/turingdb/_binary/PyTuringClient.h
sulaimansuhas: Sorry I think this was a leftover from a migration.

python/turingdb/_binary/PyTuringClient.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "dataframe/Dataframe.h"
+#include "dataframe/DataframeManager.h"
+
+#include "TuringException.h"
+
+namespace pybindings {
+
+PyTuringClient::PyTuringClient(const std::string& host, const std::string& port)
+    : _localMem(std::make_unique<db::LocalMemory>()),
+      _client(std::make_unique<net::proto::TuringClient>(host, port, _localMem.get()))
```
rjb32: Align the _client initialisation with ":"

rjb32 (approved):

PR #637, opened 2026-05-13
Implement In Process TuringDB

sulaimansuhas:
Implementation Of TuringDB running locally in a python process.

- TuringDB client class has been renamed to TuringClient
- we can start an in in process TuringDB instance with `db = TuringDB()`


python/turingdb/_local/PyTuringDB.h
```diff
@@ -0,0 +1,53 @@
...
+class TuringDB;
+}
+
+namespace pybindings {
+
+namespace nb = nanobind;
+
+class PyTuringDB {
+public:
+    PyTuringDB();
```
rjb32: Make this one explicit too

rjb32 (approved):

qdrddr:
Pleas enote that BSL license might be a barrier to adopting TuringDB as an in-process (embeddable) module/library.

rjb32:
Hi @qdrddr , indeed, BSL is the base license that we chose for now for TuringDB. We may reconsider at some point

PR #638, opened 2026-05-14
Implement BumpPtrAllocator

rjb32:
Implement a BumpPtrAllocator to allocate mixed C++ types in the flat buffers by "slabs" or by chunks.

This is to allocate and free a bunch of objects together in big chunks, to reduce the number of true allocations using malloc and improve memory locality.

### DIFF
--- a/common/BumpPtrAllocator.cpp
+++ b/common/BumpPtrAllocator.cpp
@@ -1,0 +1,164 @@
+#include "BumpPtrAllocator.h"
+
+#include <new>
+#include <stdlib.h>
+#include <memory>
+
+namespace {
+
+constexpr size_t SLAB_SAFETY_MARGIN = 64;
+
+}
+
+class BumpPtrAllocatorDestructor {
+public:
+    friend BumpPtrAllocator;
+
+    BumpPtrAllocatorDestructor(void* ptr,
+                               const BumpPtrAllocator::DestructFunc& destructFunc,
+                               BumpPtrAllocatorDestructor* prev)
+        : _prev(prev),
+        _ptr(ptr),
+        _destructFunc(destructFunc)
+    {
+    }
+
+    ~BumpPtrAllocatorDestructor() = default;
+
+    void destroy();
+
+private:
+    BumpPtrAllocatorDestructor* _prev {nullptr};
+    void* _ptr {nullptr};
+    BumpPtrAllocator::DestructFunc _destructFunc;
+};
+
+class BumpPtrAllocatorSlab {
+public:
+    friend BumpPtrAllocator;
+
+    explicit BumpPtrAllocatorSlab(size_t slabSize, BumpPtrAllocatorSlab* prev);
+
+    BumpPtrAllocatorSlab(const BumpPtrAllocatorSlab&) = delete;
+    BumpPtrAllocatorSlab& operator=(const BumpPtrAllocatorSlab&) = delete;
+
+    ~BumpPtrAllocatorSlab();
+
+    void init();
+
+    void reset();
+
+private:
+    BumpPtrAllocatorSlab* _prev {nullptr};
+    uint8_t* _data {nullptr};
+    uint8_t* _start {nullptr};
+    size_t _size {0};
+    const size_t _capacity {0};
+};
+
+// BumpPtrAllocator
+
+BumpPtrAllocator::BumpPtrAllocator(size_t defaultSlabSize)
+    : _defaultSlabSize(defaultSlabSize)
+{
+}
+
+BumpPtrAllocator::~BumpPtrAllocator() {
+    reset();
+
+    // Destroy slabs
+    BumpPtrAllocatorSlab* currentSlab = _freeSlab;
+    while (currentSlab) {
+        BumpPtrAllocatorSlab* prev = currentSlab->_prev;
+        delete currentSlab;
+        currentSlab = prev;
+    }
+}
+
+void* BumpPtrAllocator::allocate(size_t size, size_t align) {
+    const size_t requiredSlabSize = size+align+SLAB_SAFETY_MARGIN;
+
+    if (!_lastSlab || _lastSlab->_size < requiredSlabSize) {
+        // Check first free slab if any
+        if (_freeSlab && _freeSlab->_capacity >= requiredSlabSize) {
+            // If it fits in the first free slab, move it from the free list
+            // to the in-use list.
+            BumpPtrAllocatorSlab* reused = _freeSlab;
+            _freeSlab = _freeSlab->_prev;
+            reused->reset();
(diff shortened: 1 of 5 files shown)


common/BumpPtrAllocator.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+namespace {
+
+constexpr size_t SLAB_SAFETY_MARGIN = 64;
+
+}
+
+class BumpPtrAllocatorDestructor {
+public:
+    friend BumpPtrAllocator;
```
sulaimansuhas, line 15: where do we preferring putting the friends in our coding style? I see that in some places it is public and some places it is private?

rjb32: I prefer to put friend directives at the beginning of the public section

common/BumpPtrAllocator.h
```diff
@@ -0,0 +1,119 @@
...
+// CAUTION: please use bumpPtrAlloc.create<T>() when possible
+// because it calls the destructor for you.
+//
+// Usage:
+//    T* obj = new (bumpPtrAlloc) (arg1, arg2);
+// 
+// and don't forget to call delete yourself for each object so
+// that the destructor is called. 
+//
+// This must be done before the BumpPtrAllocator goes itself is deleted
```
sulaimansuhas:
```suggestion
// This must be done before the BumpPtrAllocator itself is deleted
```

common/BumpPtrAllocator.h
```diff
@@ -0,0 +1,119 @@
...
+    T* createWithNoDestruct(Args&&... args) {
+        return ::new (*this) T(std::forward<Args>(args)...);
+    }
+
+    void* allocate(size_t size, size_t align);
+
+    void reset();
+
+private:
+    static constexpr size_t DEFAULT_SLAB_SIZE = 4*1024;
```
cyrusknopf, line 63: It would be useful to have this as a template argument

common/BumpPtrAllocator.h
```diff
@@ -0,0 +1,119 @@
...
+    void* ptr = bumpPtrAlloc.allocate(size, static_cast<size_t>(align));
+    if (!ptr) {
+        throw std::bad_alloc();
+    }
+    return ptr;
+}
+
+inline void* operator new(size_t size,
+                          BumpPtrAllocator& bumpPtrAlloc) {
+    return operator new(size, std::align_val_t(alignof(std::max_align_t)), bumpPtrAlloc);
```
cyrusknopf, line 108: Why the use of max_align_t here?

common/BumpPtrAllocator.cpp
```diff
@@ -0,0 +1,164 @@
...
+        _lastSlab->_data += size;
+        _lastSlab->_size -= size;    
+        return result;
+    }
+
+    return nullptr;
+}
+
+void BumpPtrAllocator::registerDestructor(void* ptr, const DestructFunc& destructFunc) {
+   BumpPtrAllocatorDestructor* destr = new (*this) BumpPtrAllocatorDestructor(ptr, destructFunc, _lastDestruct);
```
cyrusknopf, line 107: Does this allocate the destructor object in the slab owned by this allocator?

common/BumpPtrAllocator.h
```diff
@@ -0,0 +1,119 @@
...
+
+    void reset();
+
+private:
+    static constexpr size_t DEFAULT_SLAB_SIZE = 4*1024;
+
+    const size_t _defaultSlabSize {DEFAULT_SLAB_SIZE};
+    BumpPtrAllocatorSlab* _lastSlab {nullptr};
+    BumpPtrAllocatorSlab* _freeSlab {nullptr};
+    BumpPtrAllocatorDestructor* _lastDestruct {nullptr};
```
cyrusknopf, line 68: Why is this called "last"Destruct?

common/BumpPtrAllocator.cpp
```diff
@@ -0,0 +1,164 @@
...
+        currentSlab = prev;
+    }
+}
+
+void* BumpPtrAllocator::allocate(size_t size, size_t align) {
+    const size_t requiredSlabSize = size+align+SLAB_SAFETY_MARGIN;
+
+    if (!_lastSlab || _lastSlab->_size < requiredSlabSize) {
+        // Check first free slab if any
+        if (_freeSlab && _freeSlab->_capacity >= requiredSlabSize) {
```
sulaimansuhas, line 83: why just check the first free slab, and not walk the free list?

common/BumpPtrAllocator.h
rjb32: Because in this version of the new operator we are not given the alignment value and the C++ type. So to satisfy alignment rules in a conservative way we can only use the alignment of the "worst" alignment possible.

common/BumpPtrAllocator.cpp
rjb32: Yes exactly. We take advantage of the allocator being able to allocate heterogeneous objects, including bookkeeping records, which is nice to not have extra allocations outside of it.

common/BumpPtrAllocator.cpp
rjb32: Because it may be expensive and thus counterproductive if a lot of slabs were freed in a previous reset cycle. And anyway the slabs should have a default "big enough" minimum size. So if the first one does not satisfy this allocation it means that it may be better to just allocate a fresh new slab directly.

common/BumpPtrAllocator.h
rjb32: Because we have a couple of linked lists in the allocator, the list of slabs has an anchor which is called "_lastSlab" for being the last element of the linked list of slabs, and the list of destructor records is anchored by its last element called "_lastDestruct". Traditionally anchors of intrusive linked lists may be called first, last, prev or next.

common/BumpPtrAllocator.h
cyrusknopf: cppreference says `std::max_align_t` is the maximum alignment for any 'scalar' type - does this imply a non-scalar type could have greater alignment requirements? If so then it would not align correctly.

Do we need this overload of new? Both `create` and `createWithNoDestruct` are templated on `T`, so can we not get the required alignment via `alignof(T)` and use the `operator new(size, align, ... )` overload? meaning we never waste extra alignment bytes (by falling back to `std::max_align_t`), and can handle arbitrarily large alignment requirements.

rjb32: The usage of max_align_t in that context is what seems to be recommended on the internet in the folklore. I assume that it is the worst alignment for that purpose.
I think it's a good idea for uniformity to define basically the 2 forms of placement new for completeness sake.

PR #642, opened 2026-05-14
Implement a streaming parquet parser

rjb32:
First version of a streaming parquet parser.
Can be used to read chunks of columns in a parquet file. 
It exposes a SAX-like interface where we can define callbacks to react to various events as we receive chunks.

Added a sample that reads and count the rows of an extract of 100 rows of a parquet file from OptimusKG


io/parquet/ParquetReader.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+#include <spdlog/fmt/fmt.h>
+
+#include "TuringException.h"
+
+using namespace db;
+
+ParquetSaxVisitor::~ParquetSaxVisitor()
+{
+}
```
cyrusknopf:
```suggestion
ParquetSaxVisitor::~ParquetSaxVisitor() {
}
```

io/parquet/ParquetReader.cpp
```diff
@@ -0,0 +1,319 @@
...
+            break;
+            case parquet::Type::BOOLEAN:
+                return readBoolSlice(columnReader, scratch, columnIndex, batchRows);
+            break;
+            case parquet::Type::BYTE_ARRAY:
+                return readByteArraySlice(columnReader, scratch, columnIndex, batchRows);
+            break;
+            default:
+                // INT96 and FIXED_LEN_BYTE_ARRAY not yet supported; skip silently.
+                return true;
```
cyrusknopf: Why fail silently?

io/parquet/ParquetReader.cpp
```diff
@@ -0,0 +1,319 @@
...
+}
+
+bool ParquetReader::readColumnSlice(size_t projectionIndex,
+                                    size_t columnIndex,
+                                    int64_t batchRows) {
+    parquet::ColumnReader* columnReader = _columnReaders[projectionIndex].get();
+    const parquet::SchemaDescriptor* schema = _fileMetadata->schema();
+    const parquet::ColumnDescriptor* descriptor = schema->Column(static_cast<int>(columnIndex));
+
+    std::vector<uint8_t>& scratch = _scratch[projectionIndex];
```
cyrusknopf, line 179: What is the guarantee this is a valid index?

io/parquet/ParquetReader.cpp
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    try {
+        _fileReader = parquet::ParquetFileReader::OpenFile(_path.get());
+    } catch (const parquet::ParquetException& e) {
+        throw TuringException(fmt::format("Parquet: opening {}: {}", _path.get(), e.what()));
+    }
+
+    const auto fileMetadataSharedPtr = _fileReader->metadata();
+    _fileMetadata = fileMetadataSharedPtr.get();
```
cyrusknopf, line 44: Do we need to check this is not nullptr?

io/parquet/ParquetReader.cpp
```diff
@@ -0,0 +1,319 @@
...
+}
+
+bool ParquetReader::readInt32Slice(parquet::ColumnReader* columnReader,
+                                   std::vector<uint8_t>& scratch,
+                                   size_t columnIndex,
+                                   int64_t batchRows) {
+    auto* typed = static_cast<parquet::Int32Reader*>(columnReader);
+    int32_t* buffer = reinterpret_cast<int32_t*>(scratch.data());
+    int64_t valuesRead = 0;
+    typed->ReadBatch(batchRows, nullptr, nullptr, buffer, &valuesRead);
```
cyrusknopf: How does `ReadBatch` handle if `batchRows` is negative?

io/parquet/ParquetReader.cpp
```diff
@@ -0,0 +1,319 @@
...
+    const std::span<const int32_t> values(buffer, static_cast<size_t>(valuesRead));
+    if (!_visitor.onInt32Values(columnIndex, values)) {
```
cyrusknopf: What does this condition mean?

io/parquet/ParquetReader.cpp
```diff
@@ -0,0 +1,319 @@
...
+        _aborted = true;
+        return false;
+    }
+    return true;
+}
```
cyrusknopf: Would we rather template all these functions i.e. `readSlice<T>` instead of duplicating the same logic for all types?

io/parquet/ParquetReader.cpp
rjb32: Indeed, I added support for these 2 types and throw an exception in default, even if the current enum is completely covered now

io/parquet/ParquetReader.cpp
rjb32: It does not. That's just how Parquet API is currently. I pushed the usage of size_t in our helpers and did the casting to int64 in the call to ReadBatch

io/parquet/ParquetReader.cpp
rjb32: It is checking the stopping condition of the visitor. As soon as the visitor returns false, the reading is stopped

io/parquet/ParquetReader.cpp
rjb32: When the file opens successfully, metadata() is guaranteed to return a valid shared pointer. It's very unfortunate that the API is forcing us to have shared pointers here.

io/parquet/ParquetReader.cpp
rjb32: Thanks, done

io/parquet/ParquetReader.cpp
rjb32: Thansks, I added a bioassert

io/parquet/ParquetReader.cpp
cyrusknopf: Okay makes sense

PR #651, opened 2026-05-20
Strip binaries in wheel building

rjb32:
Strip binaries and shared libraries when building the wheel to remove debug symbols

### DIFF
--- a/build_backend.py
+++ b/build_backend.py
@@ -4,6 +4,7 @@ import os
 import shlex
 import shutil
 import subprocess
+import sys
 import tempfile
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


build_backend.py
```diff
@@ -63,6 +64,16 @@ def _get_package_extensions_dir() -> Path:
     return _get_project_root() / "python" / "turingdb" / "lib" / "turingdb" / "extensions"
 
 
+def _strip_binary(path: Path) -> None:
+    """Strip debug symbols from a built binary or shared library."""
+    if sys.platform == "darwin":
```
sulaimansuhas, line 69: Why not strip on MacOs?

rjb32: Because strip on macos also happens to remove code signing so everything need to be resigned with a custom signature after strip it seems. So not sure of the impact of that.

PR #653, opened 2026-05-21
Implement Turing Proto Over HTTP

sulaimansuhas:
Replace The Raw Binary Protocol with a version running over HTTP/1.1.

Benchmark numbers show that this protocol has a similar performance to the raw binary protocol, and will provide us the ecosystem benefits of HTTP.


net/http_common/HTTPUtils.cpp
```diff
@@ -0,0 +1,111 @@
+#include "HTTPUtils.h"
+
+#include <string.h>
+#include <string>
+
+namespace net::http {
+
+void writeHexU32(uint32_t num, char* dst) {
+    for (int i = 3; i >= 0; --i) {
```
rjb32: Prefer for size_t

net/http_common/HTTPUtils.cpp
```diff
@@ -0,0 +1,111 @@
+#include "HTTPUtils.h"
+
+#include <string.h>
+#include <string>
+
+namespace net::http {
```
rjb32: Please do using namespace in the cpp file

net/turing_proto_server/TuringProtoWriter.cpp
```diff
@@ -1,99 +1,70 @@
...
 #include "BioAssert.h"
 
 using namespace net::proto;
 
 TuringProtoWriter::TuringProtoWriter(size_t bufferCapacity)
-    : _buffer(bufferCapacity),
-    _headerBuf(ProtoHeader::wireSize())
+    : _buffer(bufferCapacity)
 {
+    net::http::writeChunkTrailer(_trailerBuffer.data());
```
rjb32: What is it doing?

net/turing_proto_server/TuringProtoWriter.cpp
```diff
@@ -1,99 +1,70 @@
...
-    _errorOccured = false;
-    _hasPendingPacket = false;
-}
+void TuringProtoWriter::startResponse(net::ConnectionHeader connection) {
+    if (errorOccured()) {
+        return;
+    }
+
+    std::array<char, 256> buffer;
+    size_t offset = 0;
```
rjb32, line 57: It's dangerous to rely on these offsets to not go beyond 256 and smash the stack. Each of these write functions could take a bound for the buffer size.

net/turing_proto_server/TuringProtoWriter.cpp
```diff
@@ -1,99 +1,70 @@
...
-    _pendingBytes = 0;
-    _errorOccured = false;
-    _hasPendingPacket = false;
-}
+void TuringProtoWriter::startResponse(net::ConnectionHeader connection) {
+    if (errorOccured()) {
+        return;
+    }
+
+    std::array<char, 256> buffer;
```
rjb32: Make the size a constexpr constant

net/turing_proto_server/TuringProtoWriter.cpp
sulaimansuhas: Ah thanks for catching this! It's just instantiating the buffer values which will be reused every time. I have changed the array to be initialised with these values.

net/http_common/HTTPUtils.h
```diff
@@ -0,0 +1,114 @@
...
+
+// Default stack buffer size for assembling a response's status line + headers.
+// Sized to fit the status line plus a handful of typical header lines plus
+// the terminating CRLF. Callers needing more headers should pick a larger
+// array and pass its size through to each write*().
+inline constexpr size_t RESPONSE_HEADER_BUFFER_SIZE = 256;
+
+// 256-entry table of two-char lowercase hex byte renderings (00..ff). Lets
+// writeHexU32 be a branch-free 4-byte to 8-char conversion: each byte indexes
+// directly into a 2-char slot.
```
rjb32: Better to put this in the cpp file to avoid binary bloat

net/http_common/HTTPUtils.h
```diff
@@ -0,0 +1,114 @@
...
+void writeChunkHeaderLine(uint32_t chunkSize, char* dst, size_t dstSize);
+
+void writeChunkTrailer(char* dst, size_t dstSize);
+
+// ---- HTTP header lines -----------------------------------------------------
+// The header-line strings here match the on-wire bytes HTTPWriter already
+// emits, so HTTPWriter can be refactored to use these helpers later without
+// changing what goes out on the socket.
+
+inline constexpr std::string_view STATUS_LINE_VERSION = "HTTP/1.1 ";
```
rjb32: Same with the definition of all the strings

net/turing_proto_client/TuringClient.cpp
```diff
@@ -36,268 +41,404 @@ TuringClient::~TuringClient() {
...
+    headers += "Content-type: application/turing-proto\r\n";
+    headers += "Content-Length: ";
+    headers += std::to_string(query.size());
+    headers += "\r\n";
+    headers += "Connection: Keep-Alive\r\n";
+    headers += "\r\n";
+
+    std::array<iovec, 2> iovecs = {{
+        {const_cast<char*>(headers.data()), headers.size()},
+        {const_cast<char*>(query.data()), query.size()},
```
cyrusknopf: What's the need for const_cast here?

net/turing_proto_client/TuringClient.cpp
```diff
@@ -36,268 +41,404 @@ TuringClient::~TuringClient() {
...
+        }
 
-    if (resHeader._dataLen != sizeof(uint8_t)) {
-        throw TuringException("Invalid hello response payload size");
-    }
+        for (size_t i = 0; i + 3 < _scratchTail; ++i) {
+            const bool match = _httpScratch[i] == '\r'
+                            && _httpScratch[i + 1] == '\n'
+                            && _httpScratch[i + 2] == '\r'
+                            && _httpScratch[i + 3] == '\n';
```
cyrusknopf, line 294: Feels like it should be broken down or put in its own function

net/turing_proto_client/TuringClient.cpp
```diff
@@ -36,268 +41,404 @@ TuringClient::~TuringClient() {
...
+            const std::string_view codeStr = (space2 == std::string_view::npos)
+                                               ? afterVersion
+                                               : afterVersion.substr(0, space2);
+
+            int statusCode = 0;
+            for (const char c : codeStr) {
+                if (c < '0' || c > '9') {
+                    throw TuringException("Malformed HTTP status code");
+                }
+                statusCode = statusCode * 10 + (c - '0');
```
cyrusknopf: Feels like it should be in its own function, with a bit of explanation

net/turing_proto_client/TuringClient.cpp
```diff
@@ -36,268 +41,404 @@ TuringClient::~TuringClient() {
...
-    const uint8_t protocolVersion = 1;
-    const uint8_t keepAlive = static_cast<uint8_t>(true);
-    const uint8_t timeout = 0;
+            if (statusCode != 200) {
+                throw TuringException("Server returned HTTP " + std::to_string(statusCode));
+            }
 
-    const size_t payloadSize =
-        sizeof(protocolVersion) + sizeof(keepAlive) + sizeof(timeout);
+            _scratchHead = i + 4;
```
cyrusknopf: 4 is a bit magic

net/turing_proto_client/TuringClient.cpp
```diff
@@ -36,268 +41,404 @@ TuringClient::~TuringClient() {
...
+        if (c >= '0' && c <= '9') {
+            digit = static_cast<size_t>(c - '0');
+        } else if (c >= 'a' && c <= 'f') {
+            digit = static_cast<size_t>(c - 'a') + 10;
+        } else if (c >= 'A' && c <= 'F') {
+            digit = static_cast<size_t>(c - 'A') + 10;
+        } else {
+            throw TuringException("Invalid character in chunk size line");
+        }
+        size = (size << 4) | digit;
```
cyrusknopf: Would be better in its own function. Also can you reduce the number of branches by doing a tolower(c) first so you dont have the 3rd clause?

net/turing_proto_common/TuringProtoHeaders.cpp
```diff
@@ -35,16 +35,20 @@ void frameMessage(MessageTypes type,
...
     const ProtoHeader header {
         ._type = type,
         ._dataLen = static_cast<uint32_t>(dataBuf->size())};
-    headerBuf->copyHeader(&header);
 
-    iovecs[0] = {headerBuf->data(), headerBuf->size()};
+    size_t offset = 0;
+    memcpy(headerSpan.data() + offset, &header._type, sizeof(header._type));
+    offset += sizeof(header._type);
+    memcpy(headerSpan.data() + offset, &header._dataLen, sizeof(header._dataLen));
```
cyrusknopf: Do you need to define offset as 0? Or can the line with `offset+=` turn into `size_t offset = `?

sulaimansuhas: It could but I feel like it's a bit clearer to have it with the offset increasing from 0?

net/turing_proto_client/TuringClient.cpp
sulaimansuhas: Fixed and added a comment

net/turing_proto_common/TuringProtoHeaders.cpp
sulaimansuhas: actually realised it was a bit unnecessary actually, have refactored quiet a bit more.

net/http_common/HTTPUtils.h
sulaimansuhas: im moved most of the strings - but I did some reading and inline should let the compiler mark the symbols as weak and then the linker will do weak symbol resolution leaving no extra bloat (if I've understood correctly).

rjb32 (approved):

PR #654, opened 2026-05-27
Implement multi-source shortest path query support

### LINKED ISSUE #617: Feature: Multi-Source Shortest Path Algorithm
## Summary

Implement a `multiSourceShortestPath` processor that computes shortest paths from a set of source nodes to all reachable target nodes, returning one row per (source, target) pair. This extends the existing `shortestPath` processor, which collapses results to a single best pair.

## Motivation

The current `shortestPath` accepts multiple sources/targets but returns only the single minimum-distance result. Many analytical workloads need the full distance matrix — e.g., finding the nearest facility from every customer, or computing pairwise distances between a set of important nodes. A multi-source variant avoids issuing N separate queries.

## Proposed Syntax

```cypher
MATCH (n:City), (m:City)
WHERE n.name = 'CityA' OR n.name = 'CityB'
multiSourceShortestPath(n, m, distance, src, tgt, dist, path)
RETURN src, tgt, dist, path
```

**Parameters (7):**
1. `n` — source node set (from MATCH)
2. `m` — target node set (from MATCH)
3. `distance` — edge weight property name
4. `src` — output variable: the source node for this row
5. `tgt` — output variable: the target node reached
6. `dist` — output variable: shortest distance
7. `path` — output variable: the path

## Expected Output

One row per reachable (source, target) pair:

| src | tgt | dist | path |
|-----|-----|------|------|
| CityA | CityC | 30.0 | (0)-[1]->(1)-[2]->(2) |
| CityA | CityD | 80.0 | (0)-[1]->(1)-[2]->(2)-[3]->(3) |
| CityB | CityC | 20.0 | (1)-[2]->(2) |
| CityB | CityD | 70.0 | (1)-[2]->(2)-[3]->(3) |

## Algorithm

Run Dijkstra independently from each source node, exploring until all reachable target nodes are settled. Each source produces one row per reachable target with the shortest distance and reconstructed path.

## Implementation Scope

Follows the same layered pattern as the existing `shortestPath`:

- **Lexer/Parser:** New `MULTISOURCESHORTESTPATH` token and grammar rule (7 parameters)
- **AST:** `MultiSourceShortestPathStmt` node
- **Analyzer:** Validate edge property exists and is numeric; register 4 output variables (src, tgt, dist, path)
- **Plan:** `MultiSourceShortestPathNode` with new `MULTI_SOURCE_SHORTEST_PATH` opcode
- **Processor:** `MultiSourceShortestPathProcessor<T>` — templated over weight type (UInt64, Int64, Double)
- **Pipeline Builder/Generator:** Wiring methods analogous to existing shortest path
- **Tests:** Query-level and processor-level tests

## Constraints

- Rejects negative edge weights (same as existing `shortestPath`)
- Returns 0 rows if no paths exist between any source-target pair
- Supports the same weight types as `shortestPath`: UInt64, Int64, Double

patelchaitany:
Add multiSourceShortestPath processor that computes shortest paths from a set of source nodes to all reachable targets, returning one row per (source, target) pair with distance and path — extends the existing shortestPath which only returns a single best pair. Closes #617


patelchaitany:
@rjb32 if you get time pls review this PR.

rjb32:
> @rjb32 if you get time pls review this PR.

Thank you very much @patelchaitany! I will review it on Monday

patelchaitany:
Hey @rjb32! Whenever you get a moment, I'd love to get your thoughts on this PR. Thanks!

sulaimansuhas:
Hey @patelchaitany - I have looked at the PR and had a few overall comments:
- Can you think of anyways we can re-use explored paths between source nodes? This would help make the algorithm a lot more efficient I feel.
- I think there is a lot of overlap between the ShortestPath processor code and the multi-path shortest path processor code. It would be nice to wrap up common logic and data structures into a single ShortestPathUtils.h/.cpp file!
- I think we need a lot more comments in your code - especially around the shortest path algorithm.

Overall the code looks good and conforms to our standards well:)!

query/pipeline/processors/MultiSourceShortestPathProcessor.cpp
```diff
@@ -0,0 +1,244 @@
...
+    finish();
+}
+
+template <SupportedType T>
+void MultiSourceShortestPathProcessor<T>::runDijkstra(NodeID sourceNode,
+                                                      ColumnVector<NodeID>* sourceOutputCol,
+                                                      ColumnVector<NodeID>* targetOutputCol,
+                                                      ColumnVector<EdgePropType>* distCol,
+                                                      ColumnVector<Path>* pathCol) {
+    DijkstraHeap heap;
```
rjb32: All these structures are going to be recreated per runDijkstra call, you could save them as members of the processor class and clear them on entry in the function

query/pipeline/processors/MultiSourceShortestPathProcessor.h
```diff
@@ -0,0 +1,123 @@
...
+    NodeID prevNode;
+    EdgeID edge;
+    T distance {0};
+};
+
+template <SupportedType T>
+class MultiSourceShortestPathProcessor final : public Processor {
+public:
+    using EdgePropType = T::Primitive;
+    using DijkstraHeap = std::priority_queue<MultiSourceDijkstraNode<EdgePropType>,
```
rjb32: Probably better to extract all these type definitions and the runDijkstra mechanics into some utility class that would wrap up everything related to Dijkstra algorithm

query/plan/PipelineGenerator.cpp
```diff
@@ -1629,6 +1634,59 @@ PipelineOutputInterface* PipelineGenerator::translateShortestPathNode(ShortestPa
...
+                                                                pathCol);
+            sourceOutputCol->rename(node->getSourceOutput()->getName());
+            targetOutputCol->rename(node->getTargetOutput()->getName());
+            distCol->rename(node->getDistance()->getName());
+            pathCol->rename(node->getPath()->getName());
+        } else {
+            throw PlannerException("Unsupported Edge Weight Type");
+        }
+    };
+    ValueTypeDispatcher {edgeType._valueType}.execute(process);
```
rjb32: Coding style: we prefer to use the parenthesis for constructors instead of braces, such as ValueTypeDispatcher(edgeType._valueType)

patelchaitany:
@rjb32 @sulaimansuhas Feedback addressed! Please take another look and let me know if any further changes are needed. Happy to revise.

sulaimansuhas:
Hey @patelchaitany. I looked over your changes and it mostly looks good! I have two comments:
- I think the Dijkstra Search Algorithm itself should be pulled out into a common function, as there is a lot of overlapping logic that would be painful to keep in sync between these two implementations.
- I feel that there is scope to optimise this code - we should be able to cache already explored paths in previous runs with different source nodes. Please give me your thoughts on this (even why it wouldn't be feasible).

thank you for your good work!

patelchaitany:
I'm going to expand `ShortestPathUtils` into a `DijkstraRunner<T>` class that owns the heap and value map as members (reused across calls instead of reallocated) and holds the core Dijkstra loop - expansion, stale filtering, relaxation, and path reconstruction. A `stopAtFirst` flag on `run()` handles both use cases: `ShortestPathProcessor` stops at the first settled target, `MultiSourceShortestPathProcessor` keeps going until all are found, so both processors shrink down to just port wiring and result extraction. On caching across source nodes - the shortest-path trees are source-dependent so intermediate results don't transfer cleanly; the early-exit we already have cuts out the main waste, and anything beyond that (bidirectional Dijkstra, landmarks) feels like a separate effort. Will also fix the `ValueTypeDispatcher` brace style Remy flagged.

sulaimansuhas:
Hey @patelchaitany - your code changes sound good, I'll be waiting to review! As for your comments on the shortest path optimisations:

If we take the shortest path between node a and z to be a->p->x->z, then for that path we are guaranteed that every subpath of the path is the optimal shortest path from the source node of that subpath to the target. So from this single traversal we can infer that the shortest path from p to z is p->x->z.

Given this finding, two optimisations I can think of off the top of my head:
- if p was in the source node set we already have a shortest path to a node in the target set (z).
- if in any other traversal we pop p from the top of the stack we immediately can know that there is a path to a target node and the size of the path.

these are the ideas that gave me the intuition for optimisation in this case. I think it'd be a shame to get this PR through without exploring these kind of optimisations as I think they have the potential to reduce a lot of computation.

I'd like to hear your thoughts on this, and thank you for all your good work!

patelchaitany:
@sulaimansuhas - subpath caching is in. After each run we cache intermediate-to-target subpaths, then before the next source we check the cache and skip Dijkstra if we already have the answer. Mid-traversal hits are handled as pending results that only finalize when the heap confirms optimality.

Right now the cache is query-scoped - thinking a cross-query version keyed to commit hashes (with eviction like `ShardCache` does for vectors) would be a natural follow-up. Thoughts?

patelchaitany:
Hey @rjb32 , @sulaimansuhas Is there any changes are required.

sulaimansuhas:
Hey @patelchaitany, sorry for taking so long! I looked through the PR it looks pretty good. I'll have some final comments that I'll add later today.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+    _propertyIndices = propertyIndices;
+    _properties = properties;
+    _getPropertiesWriter = getPropertiesWriter;
+}
+
+template <SupportedType T>
+void DijkstraRunner<T>::run(const DijkstraHeap<EdgePropType>& initialHeap,
+                            const DijkstraValueMap<EdgePropType>& initialValues,
+                            const std::unordered_set<NodeID>& targetNodes,
+                            bool stopAtFirst,
```
sulaimansuhas, line 39: I don't like having the bool flag, I think it should be a number. This would make it trivial to create a top K shortest paths algorithm in the future.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+    _heapValueMap = initialValues;
+    _settledTargets.clear();
+    _results.clear();
+    _pendingResults.clear();
+
+    while (!_heap.empty()) {
+        const DijkstraNode<EdgePropType> val = _heap.top();
+        _heap.pop();
+
+        const auto it = _heapValueMap.find(val.id);
```
sulaimansuhas, line 51: Please add a comment explaining that we are removing the stale values here.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+        if (it != _heapValueMap.end() && it->second.distance != val.distance) {
+            continue;
+        }
+
+        // Finalize pending cache-hit results whose distance cannot be beaten
+        // by any future path (all remaining nodes have distance >= val.distance).
+        for (auto pendingIt = _pendingResults.begin(); pendingIt != _pendingResults.end(); ) {
```
sulaimansuhas, line 58: Nice! Please add more comments explaining what is happening here. And expand on the why - these are complex algorithms that we want to as readable as possible.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+            if (stopAtFirst) {
+                break;
+            }
+
+            if (_settledTargets.size() == targetNodes.size()) {
+                break;
+            }
+        }
+
+        // Consult the subpath cache: if this settled node has known shortest
```
sulaimansuhas, line 87: This comment is not clear at all, please clearly explain the logic behind this cache search better. Something like:
If the current popped node is found in the cache then we know we have the shortest path from our source to the Target Node in the cache entry.

sulaimansuhas: Also please clarify the explanation on why we need the pending result sets. It is useful but it took me a while to understand it just based on the code.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+        // paths to unsettled targets, record them as pending results.
+        if (cache) {
+            const auto cacheIt = cache->find(val.id);
+            if (cacheIt != cache->end()) {
+                for (const SubpathCacheEntry<EdgePropType>& entry : cacheIt->second) {
+                    if (!targetNodes.contains(entry.targetNode)) {
```
sulaimansuhas, line 93: both these if conditions should be one statement.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+                        continue;
+                    }
+                    if (_settledTargets.contains(entry.targetNode)) {
+                        continue;
+                    }
+
+                    const EdgePropType candidateDistance = val.distance + entry.distance;
+
+                    const auto existingIt = _pendingResults.find(entry.targetNode);
```
sulaimansuhas, line 102: I don't think this conditional is needed, as for the given target node the `candidateDistance` will be the shortest distance anyways.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+        const auto it = _heapValueMap.find(val.id);
+        if (it != _heapValueMap.end() && it->second.distance != val.distance) {
+            continue;
+        }
+
+        // Finalize pending cache-hit results whose distance cannot be beaten
+        // by any future path (all remaining nodes have distance >= val.distance).
+        for (auto pendingIt = _pendingResults.begin(); pendingIt != _pendingResults.end(); ) {
+            if (pendingIt->second.distance <= val.distance) {
```
sulaimansuhas, line 59: we would want to check stop at first (or the the size of results against a size parameter passed to the function here)

query/pipeline/processors/ShortestPathUtils.cpp
sulaimansuhas: so we can mark the target as settled here.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+                    _pendingResults[entry.targetNode] = pending;
+                }
+            }
+        }
+
+        expandNode(val);
+    }
+
+    // Finalize any remaining pending results (heap exhausted, no shorter path exists).
+    for (auto& [targetNode, pending] : _pendingResults) {
```
sulaimansuhas, line 121: Would we even need this? If we are going to be exhausting the heap or finding the k shortest paths. We've either found our K shortest paths and do not need to add any more values. Or if we have legitimately exhausted the heap and thus explored the graph wouldn't we have come across paths >= in distance to those in `_pendingResults` and would have added them already?

query/pipeline/processors/ShortestPathUtils.h
```diff
@@ -0,0 +1,119 @@
...
+             bool stopAtFirst,
+             const SubpathCache<EdgePropType>* cache = nullptr);
+
+    const std::vector<DijkstraResult<EdgePropType>>& results() const { return _results; }
+    const DijkstraValueMap<EdgePropType>& getValueMap() const { return _heapValueMap; }
+
+private:
+    DijkstraHeap<EdgePropType> _heap;
+    DijkstraValueMap<EdgePropType> _heapValueMap;
+    std::unordered_set<NodeID> _settledTargets;
```
sulaimansuhas, line 97: I think it makes more sense to pass the constructed target nodes to the Utils class as a const, and have the settledTargets passed to the function itself. This way we don't have to create a copy of the targetNodes set every time. This would be especially useful if the target nodes set itself is quite large.

query/pipeline/processors/ShortestPathUtils.h
```diff
@@ -0,0 +1,119 @@
...
+    std::vector<DijkstraResult<EdgePropType>> _results;
+    std::unordered_map<NodeID, DijkstraResult<EdgePropType>> _pendingResults;
+
+    ColumnNodeIDs* _inputNodes {nullptr};
```
sulaimansuhas, line 101: We can forward declare all these pointer members.

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+        if (!_settledTargets.contains(targetNode)) {
+            _settledTargets.insert(targetNode);
+            _results.push_back(pending);
+        }
+    }
+    _pendingResults.clear();
+}
+
+template <SupportedType T>
+void DijkstraRunner<T>::expandNode(const DijkstraNode<EdgePropType>& node) {
```
sulaimansuhas, line 131: please add back all the comments that were removed here ( see original algo)

query/pipeline/processors/ShortestPathUtils.cpp
```diff
@@ -0,0 +1,204 @@
...
+    auto edge = settledNode.edge;
+    while (lastNode.isValid()) {
+        outputPath.push_back(edge.getValue());
+        outputPath.push_back(lastNode.getValue());
+
+        const auto& pathInfo = _heapValueMap[lastNode];
+        lastNode = pathInfo.prevNode;
+        edge = pathInfo.edge;
+    }
+}
```
sulaimansuhas, line 181:
```suggestion
    // Start with the cached path suffix without the starting node: [target, edge, ..., edge]
    outputPath = cacheEntry.pathSuffix | ranges::views::drop_last(1);
    reconstructPath(settledNode, outputPath);
    }
}
```

query/pipeline/processors/ShortestPathUtils.h
```diff
@@ -0,0 +1,119 @@
...
+    NodeID targetNode;
+    T distance {0};
+    Path path;
+};
+
+template <typename T>
+struct SubpathCacheEntry {
+    NodeID targetNode;
+    T distance {0};
+    Path pathSuffix;
```
sulaimansuhas, line 62: pathSuffix is not the correct name for this. Path would be fine.

sulaimansuhas (changes requested):
Hey @patelchaitany, the overall logic of your change is sound. I have left quite a few comments explaining what would be needed to make this mergeable. I think we need a lot more comments around the algorithm - in places my previous comments have been removed where my code was copied. If you are using AI coding tools - I request that you write the comments yourself because we generally find the AI generated comments to not be that clear to human readers, and it will also help you reason about the logic itself.

patelchaitany:
@sulaimansuhas, Okay I will address your comments and not try to use AI for comments

PR #656, opened 2026-06-01
Fix Rule of 5 violations with ListBuffer and co.

cyrusknopf:
`ListByteBuffer` and `ListElementViewBuffer` manually managed memory, but had implicitly defined copy and move constructors/assignment which did not serve that memory management. Deleted to prevent misuse.

### DIFF
--- a/storage/list/ListElementViewBuffer.h
+++ b/storage/list/ListElementViewBuffer.h
@@ -26,6 +26,11 @@ public:
     /// Deallocates all owned chunks
     ~ListElementViewBuffer();
 
+    ListElementViewBuffer(const ListElementViewBuffer&) = delete;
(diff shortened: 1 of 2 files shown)


rjb32 (approved):

storage/list/ListElementViewBuffer.h
```diff
@@ -91,6 +96,11 @@ class ListElementViewBuffer<N>::Chunk {
         delete[] _buf;
     }
 
+    Chunk(const Chunk&) = delete;
+    Chunk(Chunk&&) = delete;
+    Chunk& operator=(const Chunk&) = delete;
+    Chunk& operator=(Chunk&&) = delete;
```
rjb32, line 102: Move constructors are harmless in general

PR #660, opened 2026-06-01
Move Graphs to GraphManager and use an ObjectMap

rjb32:
* Wraps the map of graphs that is in SystemManager to a GraphManager
* Manage graphs with ObjectMap #658 

Caution: yes the calls in GraphManager are still a bit ugly, to be fixed later


system/GraphManager.cpp
```diff
@@ -0,0 +1,351 @@
...
+GraphManager::GraphManager(const TuringConfig* config)
+    : _config(config)
+{
+}
+
+GraphManager::~GraphManager() {
+}
+
+Graph* GraphManager::getGraph(std::string_view graphName) const {
+    const auto* slot = _graphs.getObject(graphName);
```
rjb32: Don't use auto here

system/GraphManager.cpp
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    if (!graph) {
+        throw FatalException("Could not initialise the default graph");
+    }
+
+    _defaultGraph.store(graph);
+}
+
+bool GraphManager::importGraph(std::string_view graphName, const fs::Path& filePath, JobSystem& jobSystem) {
```
rjb32: Align each argument one below each other

system/GraphManager.cpp
rjb32: Done in 71c540835 — spelled out the type as `const ObjectMap<Graph>::ObjectSlot*`.

system/GraphManager.cpp
rjb32: Done in 71c540835 — each argument now on its own line, aligned under the first.

storage/Graph.cpp
```diff
@@ -10,6 +10,22 @@
 
 using namespace db;
 
+Graph::Graph()
+    : _graphName("default"),
+    _graphPath("/dev/null"),
```
cyrusknopf, line 15: Why not initialise these in the member declaration?

storage/Graph.cpp
```diff
@@ -10,6 +10,22 @@
...
+    _versionController(new VersionController {this}),
+    _serializer(new GraphSerializer {this})
```
cyrusknopf, line 17: I know this is not new to this change, but why are both of these heap allocated?

system/GraphManager.cpp
```diff
@@ -0,0 +1,353 @@
...
+}
+
+Graph* GraphManager::createGraph(std::string_view name) {
+    const fs::Path path = _config->getGraphsDir() / name;
+
+    if (_config->isSyncedOnDisk() && path.exists()) {
+        throw TuringException(fmt::format("Graph '{}' already exists", name));
+    }
+
+    auto reservation = _graphs.reserve(name);
```
cyrusknopf: Explicit type would aid readability

system/SystemManager.cpp
```diff
@@ -154,325 +150,54 @@ void SystemManager::initSystemEvents() {
...
-    // Search if a graph with the same name exists
-    const auto it = _graphs.find(name);
-    if (it != _graphs.end()) {
-        return false;
-    }
-
-    _graphs[name] = std::move(graph);
-
-    return true;
+    return _graphManager.createGraph(name);
```
cyrusknopf: Can all the methods in SystemManager be changed to take a string_view without too much pain?

storage/Graph.cpp
rjb32: I don't like to initialise strings in the header files because of the risk of binary bloat. At least in the cpp file I am sure that the string will only be stored uniquely in one object file and not in each object file where the class is instantiated.

storage/Graph.cpp
rjb32: This is bad I will eliminate it in my next pass.

system/SystemManager.cpp
rjb32: Good one thanks

system/GraphManager.cpp
rjb32: Done

system/SystemManager.cpp
rjb32: Done

storage/ChangeManager.h
sulaimansuhas, line 1: Why is this in the storage directory now? I think it's more appropriate to have it in system?

rjb32: It's a leftover of my previous design. But if changes belong to graphs, is that really a system consideratIon?

sulaimansuhas: For the change class it makes sense to be in storage. But my feeling is any manager class should come under system right (it belongs to graph manager anyways right?).

sulaimansuhas: or are you referring to the fact that the change manager will be a per graph object at some point?

rjb32: It's not because an object is a manager object that has to be in system. It depends on the nature of the object being managed. Is it a top level system concern? Is it an internal "storage" concern?
In the case of changes, this is just an intermediary PR and things are very much in flux right now. 
The goal is to move changes into graphs as they should be.

PR #664, opened 2026-06-04
Remove GraphSerializer

rjb32:


storage/dump/GraphDumper.cpp
```diff
@@ -17,6 +19,8 @@ using namespace db;
 DumpResult<void> GraphDumper::dump(const Graph& graph, const fs::Path& graphDir) {
```
sulaimansuhas: Can we change this to a pointer - to fit our style as well as match the `GraphDumper::load` pattern.

rjb32: Done

PR #666, opened 2026-06-05
Implement List Types Over Binary Protocol

sulaimansuhas:


net/encoders/TuringProtoEncoder.h
```diff
@@ -1,23 +1,57 @@
...
+    std::span<const db::ListElementView>::iterator currIt;
+    std::span<const db::ListElementView>::iterator endIt;
+};
+
+/// Dispatched on a list element's runtime tag to return the size of the object
+/// the tag maps to, so the client knows how much to allocate. For String and
+/// Embedding this is the size of the view object (std::string_view / std::span),
+/// not the length of the data it points to.
+struct ListElementByteSizeVisitor {
+    template <typename T>
```
rjb32, line 35: Isn't this a bit weird that there is only one function?

net/encoders/TuringProtoEncoder.h
```diff
@@ -1,23 +1,57 @@
...
+
+// Writes one list as [count][listByteSize] followed by its [tag][value] /
+// [tag][numBytes][data] elements. Shared by the vector and constant paths.
+// Σ sizeof(value) over the elements — the deserialized footprint the decoder reserves.
+inline uint32_t computeListByteSize(std::span<const db::ListElementView> elements) {
+    const ListElementByteSizeVisitor sizeVisitor;
+
+    size_t totalSize = 0;
+    for (const auto& elem : elements) {
+        totalSize += db::ListTagDispatcher {elem.getTag()}.execute(sizeVisitor, elem);
```
rjb32: Let's break dispatcher construction and execute call into 2 separate lines. Do not construct objects with braces { } style.

net/encoders/TuringProtoEncoder.h
```diff
@@ -133,6 +173,63 @@ struct ColumnHeaderWriter {
...
+            const uint32_t elementCount = static_cast<uint32_t>(value.size());
+            const uint32_t listByteSize = computeListByteSize(value);
+
+            _outBuf->checkRemainingAndFlush(tagSize + sizeof(elementCount) + sizeof(listByteSize));
+            _outBuf->copyFixedLenData(&tag, tagSize);
+            _outBuf->copyFixedLenData(&elementCount, sizeof(elementCount));
+            _outBuf->copyFixedLenData(&listByteSize, sizeof(listByteSize));
+
+            _stack->push({value.begin(), value.end()});
+        } else {
```
rjb32, line 225: Is there a risk here in case the set of what is StringLike grows in the future and we fall in this else branch unexpectedly?

net/encoders/TuringProtoEncoder.h
```diff
@@ -333,6 +445,55 @@ class DataWriter {
     }
 
 private:
+    // Writes each element's [tag][value] / [tag][numBytes][data].
+    void writeListElementValues(std::span<const db::ListElementView> elements) {
+        std::stack<NestedListStackElement> stack;
```
rjb32: To what extent could be reuse and clear this stack?

net/turing_proto_common/Decoders.h
```diff
@@ -1,20 +1,187 @@
...
+ * For each top-level element completed (one written through the column's own cursor, i.e. stack
+ * depth 1 — nested elements live deeper and are not column rows), @param onTopLevelElement is
+ * invoked with its 0-based index and view. ListElementView columns store the view; ListView
+ * columns pass a no-op.
+ */
+template <typename OnTopLevelElement>
+inline bool drainListStack(TuringProtoDecoder::DecodeContext* ctx,
+                           const OnTopLevelElement& onTopLevelElement) {
+    db::ListElementView view;
+    const ListElementReadVisitor readVisitor {ctx, &view};
```
rjb32, line 149: Coding style: don't use brace syntax { } for constructing objects

net/encoders/TuringProtoEncoder.h
sulaimansuhas: It's just to fit into the existing dispatcher code - rather adding more boilerplate.

net/turing_proto_common/Decoders.h
sulaimansuhas: I am leaving these in as they are structs?

net/encoders/TuringProtoEncoder.h
sulaimansuhas: I've added a static assert like I have in other parts of this codebase

storage/list/ListWriteCursor.cpp
```diff
@@ -0,0 +1,45 @@
...
+using namespace db;
+
+ListElementView ListWriteCursor::writeRaw(const void* data, size_t numBytes) {
+    std::byte* start = _elementWritePtr;
+
+    std::memcpy(start, data, numBytes);
+    _elementWritePtr += numBytes;
+
+    return recordView(start);
+}
```
cyrusknopf, line 18: Can this be templated instead of `void*`? If so please do, so that you can `static_assert` that the type is trivially copyable and preferably also that it is `Listable`

storage/list/ListWriteCursor.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+template <typename T>
+ListElementView ListWriteCursor::writeValue(ListBufferTypeTag tag, const T& value) {
+    constexpr size_t tagSize = sizeof(ListBufferTypeTag);
```
cyrusknopf, line 22: Please add a static_assert that this `tagSize = 1` with a message that the tag size has changed if it fails

storage/list/ListWriteCursor.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    return recordView(start);
+}
+
+ListElementView ListWriteCursor::recordView(const std::byte* elementStart) {
+    const ListElementView view {elementStart};
+
+    *_viewWritePtr = view;
+    ++_viewWritePtr;
+    ++_written;
```
cyrusknopf, line 40: Could you explain what `*_viewWritPtr = view` is doing? and the following line, not sure I understand

storage/columns/ColumnDispatcher.h
```diff
@@ -183,6 +184,8 @@ struct ColumnTypeFromKind {
         CONST_COL_CASE(ColumnVector<const CommitBuilder*>)                               \
         CONST_COL_CASE(ColumnVector<const Change*>)                                      \
         CONST_COL_CASE(ColumnVector<EntityList>)                                         \
+        CONST_COL_CASE(ColumnVector<ListView>)                                           \
+        CONST_COL_CASE(ColumnVector<ListElementView>)                                    \
```
cyrusknopf: We should stop using this, what callsite needed this added?

net/encoders/TuringProtoEncoder.h
```diff
@@ -133,10 +174,70 @@ struct ColumnHeaderWriter {
...
+            bioassert(value.size() <= std::numeric_limits<uint32_t>::max(), "List element count exceeds maximum wire size");
+            const uint32_t elementCount = static_cast<uint32_t>(value.size());
+            const uint32_t listByteSize = computeListByteSize(value);
+
+            _outBuf->checkRemainingAndFlush(tagSize + sizeof(elementCount) + sizeof(listByteSize));
+            _outBuf->copyFixedLenData(&tag, tagSize);
+            _outBuf->copyFixedLenData(&elementCount, sizeof(elementCount));
+            _outBuf->copyFixedLenData(&listByteSize, sizeof(listByteSize));
+
+            _stack->push({value.begin(), value.end()});
```
cyrusknopf: use emplace

net/encoders/TuringProtoEncoder.h
```diff
@@ -133,10 +174,70 @@ struct ColumnHeaderWriter {
...
+/// fixed framing are kept within a single chunk (checkRemainingAndFlush) so the decoder,
+/// which reads each unit atomically, never sees them split across a packet boundary; the
+/// variable payload that follows still streams across chunks via copyVarLenData.
+struct ListElementWriteVisitor {
+    net::proto::TuringProtoOutBuf* _outBuf {nullptr};
+    std::stack<NestedListStackElement>* _stack {nullptr};
+
+    template <typename T>
+    void operator()(const db::ListElementView elem) const {
+        constexpr size_t tagSize = sizeof(db::ListBufferTypeTag);
```
cyrusknopf, line 190: Pls add a static assert that tagsize=1 with a message

net/encoders/TuringProtoEncoder.h
```diff
@@ -188,9 +289,17 @@ class DataWriter {
...
                 }
             }
+        } else if constexpr (db::IsListView<T>) {
+            for (const auto& listView : *col) {
+                writeListView(listView);
+            }
+        } else if constexpr (db::IsListElement<T>) {
+            // One element per row: the row count (already written above) is the element
+            // count, so only [listByteSize] + the elements follow.
+            writeListElements(std::span<const db::ListElementView>(col->data(), col->size()));
```
cyrusknopf: Can you not just pass col.getRaw() directly and the vector will implict convert to span, without the need of this explicit std::span initialisation?

net/encoders/TuringProtoEncoder.h
```diff
@@ -243,7 +352,7 @@ class DataWriter {
                 _outBuf->copyFixedLenData(&columnByteSize, sizeof(columnByteSize));
                 _outBuf->copyVarLenData(val->data(), columnByteSize);
             }
-        } else if constexpr (db::IsList<T>) {
+        } else if constexpr (db::IsEntityList<T>) {
             // EntityList is only used as ColumnVector<EntityList>, never optional
```
cyrusknopf, line 357: Can we have an exception here then?

net/encoders/TuringProtoEncoder.h
```diff
@@ -282,8 +391,15 @@ class DataWriter {
...
-            // EntityList is only used as ColumnVector<EntityList>, never const
+        } else if constexpr (db::IsEntityList<T>) {
+            // It is not used/returned anywhere in the codebase so we disable support for
+            // it now. We can renable it if ever needed.
+            throw FatalException("ColumnConst<EntityList> is not supported");
+        } else if constexpr (db::IsListView<T>) {
+            writeListView(col->at(0));
+        } else if constexpr (db::IsListElement<T>) {
+            const db::ListElementView element = col->at(0);
+            writeListElements(std::span<const db::ListElementView>(&element, 1));
```
cyrusknopf, line 404: Use getRaw, not at(0)

net/encoders/TuringProtoEncoder.h
```diff
@@ -333,7 +449,56 @@ class DataWriter {
     }
 
 private:
+    // Writes each element's [tag][value] / [tag][numBytes][data].
+    void writeListElementValues(std::span<const db::ListElementView> elements) {
+        const ListElementWriteVisitor writeVisitor {_outBuf, &_stack};
+
+        _stack.push({elements.begin(), elements.end()});
```
cyrusknopf: Emplace

net/encoders/TuringProtoEncoder.h
```diff
@@ -333,7 +449,56 @@ class DataWriter {
...
+        while (!_stack.empty()) {
+            auto& frame = _stack.top();
+            if (frame.currIt == frame.endIt) {
+                _stack.pop();
+                continue;
+            }
+
+            const auto it = frame.currIt;
+            ++frame.currIt;
+            db::ListTagDispatcher {it->getTag()}.execute(writeVisitor, *it);
```
cyrusknopf: A structured binding here on `auto& [current, end] = _stack.top()` would aid readability

net/encoders/TuringProtoEncoder.h
```diff
@@ -333,7 +449,56 @@ class DataWriter {
...
+    // Helper function to write ColumnVectors or Column Const List Element Views.
+    void writeListElements(std::span<const db::ListElementView> elements) {
+        const uint32_t listByteSize = computeListByteSize(elements);
+
+        _outBuf->checkRemainingAndFlush(sizeof(listByteSize));
+        _outBuf->copyFixedLenData(&listByteSize, sizeof(listByteSize));
+
+        writeListElementValues(elements);
+    }
+
```
cyrusknopf, line 501: I am wildly out of context here, but does `checkRemainingAndFlush` mean that the size of the list in bytes might be in a different packet than the elements themselves? and is that fine? (Do correct me if I my understanding is wrong here)

net/turing_proto_client/TuringClient.h
```diff
@@ -52,7 +53,9 @@ class TuringClient {
     std::string_view getRemotePort() const { return _remotePort; }
     std::string_view getGraphName() const { return _graphName; }
 
-    net::proto::EmbeddingBuffer& getEmbeddingBuffer() { return _embeddingBuffer; }
+    net::proto::ChunkedBuffer<float>& getEmbeddingBuffer() { return _embeddingBuffer; }
+    net::proto::ChunkedBuffer<char>& getStringBuffer() { return _stringBuffer; }
+    db::ListBuffer<>& getListBuffer() { return _listBuffer; }
```
cyrusknopf: We have `DefaultListBuffer` as a typedef for the default list buffer, see if you'd prefer to/can use that here

net/turing_proto_common/TuringProtoDecoder.h
```diff
@@ -58,19 +61,31 @@ class TuringProtoDecoder {
...
-        EmbeddingBuffer* _embeddingBuffer {nullptr};
+        ChunkedBuffer<float>* _embeddingBuffer {nullptr};
+        ChunkedBuffer<char>* _stringBuffer {nullptr};
+        db::ListBuffer<>* _listBuffer {nullptr};
+
+        // Stack of write cursors into the preallocated list buffers. We always write the
+        // latest element received on the wire through the cursor at the top of the stack.
+        // If we find a nested list we push a child cursor; once a cursor's elements are all
+        // written we pop it and resume writing through the one beneath. Lives here (rather
+        // than as a local) so a list resumes across buffer boundaries.
```
cyrusknopf, line 72: Is this still correct? I thought I read somewhere that a list can't cross buffer boundaries?

net/turing_proto_common/Decoders.h
```diff
@@ -74,6 +241,76 @@ inline bool decodeVector<std::string>(TuringProtoDecoder::DecodeContext* ctx,
...
+            if (ctx->_inBuf->readable() < 2 * sizeof(uint32_t)) {
+                return false;
+            }
+
+            uint32_t elementCount = 0;
+            uint32_t listByteSize = 0;
+            ctx->_inBuf->readData(&elementCount, sizeof(elementCount));
+            ctx->_inBuf->readData(&listByteSize, sizeof(listByteSize));
+
+            stack.push(ctx->_listBuffer->reserveList(elementCount, listByteSize));
```
cyrusknopf: Probably emplace here

net/turing_proto_common/Decoders.h
```diff
@@ -74,6 +241,76 @@ inline bool decodeVector<std::string>(TuringProtoDecoder::DecodeContext* ctx,
...
+    // our encoded ListView and need to allocate the list buffer itself
+    if (ctx->_rowIndex == 0) {
+        if (ctx->_inBuf->readable() < sizeof(uint32_t)) {
+            return false;
+        }
+
+        uint32_t listByteSize = 0;
+        ctx->_inBuf->readData(&listByteSize, sizeof(listByteSize));
+
+        stack.push(ctx->_listBuffer->reserveList(columnState->_numRows, listByteSize));
```
cyrusknopf: Probably empalce here

net/turing_proto_common/Decoders.h
```diff
@@ -74,6 +241,76 @@ inline bool decodeVector<std::string>(TuringProtoDecoder::DecodeContext* ctx,
...
+        typedCol->resize(columnState->_numRows);
+        ctx->_rowIndex = 1;
+    }
+
+    // One column row per top-level element; its view is stored in the matching row.
+    return drainListStack(ctx, [typedCol](size_t index, const db::ListElementView& view) {
+        typedCol->data()[index] = view;
+    });
```
cyrusknopf: This is very hard to parse, can you save the lambda as a variable with a name and pass it to the `drainListStack`? Also saving the result of teh function call in a meaningfully named boolean would also help

net/turing_proto_common/Decoders.h
```diff
@@ -437,6 +674,57 @@ inline bool decodeConst<db::types::Embedding::Primitive>(TuringProtoDecoder::Dec
...
+template <>
+inline bool decodeConst<db::ListView>(TuringProtoDecoder::DecodeContext* ctx,
+                                      db::ColumnConst<db::ListView>* typedCol) {
+    auto& stack = ctx->_listStack;
+
+    // Const columns have no row dimension, so reuse _rowIndex as a 0/1 "list started"
+    // marker (the driver resets it to 0 before this column). 1 means we have read the
+    // header, reserved the space, and set the (initially unfilled) ListView on the column.
+    if (ctx->_rowIndex == 0) {
+        if (ctx->_inBuf->readable() < 2 * sizeof(uint32_t)) {
```
cyrusknopf: I see uint32_t a lot in assertions about the size of the buffers for the protocol. Is that a hard limit of the protocol? if so I'd rather we have a typedef somewhere in the network code which  we can use instead, so that if the limit changes we don't need to search for all these checks based on uint32_t

net/turing_proto_common/Decoders.h
```diff
@@ -437,6 +674,57 @@ inline bool decodeConst<db::types::Embedding::Primitive>(TuringProtoDecoder::Dec
...
+            return false;
+        }
+
+        uint32_t elementCount = 0;
+        uint32_t listByteSize = 0;
+        ctx->_inBuf->readData(&elementCount, sizeof(elementCount));
+        ctx->_inBuf->readData(&listByteSize, sizeof(listByteSize));
+
+        stack.push(ctx->_listBuffer->reserveList(elementCount, listByteSize));
```
cyrusknopf: Probably emplace

net/turing_proto_common/Decoders.h
```diff
@@ -437,6 +674,57 @@ inline bool decodeConst<db::types::Embedding::Primitive>(TuringProtoDecoder::Dec
...
+        typedCol->set(stack.top().getView());
+        ctx->_rowIndex = 1;
+    }
+
+    return drainListStack(ctx, [](size_t, const db::ListElementView&) {});
```
cyrusknopf: Same comment on saving the lambda, saving the bool

net/turing_proto_common/Decoders.h
```diff
@@ -437,6 +674,57 @@ inline bool decodeConst<db::types::Embedding::Primitive>(TuringProtoDecoder::Dec
...
+        ctx->_inBuf->readData(&listByteSize, sizeof(listByteSize));
+
+        stack.push(ctx->_listBuffer->reserveList(1, listByteSize));
+        ctx->_rowIndex = 1;
+    }
+
+    // The constant is the single top-level element; store its view on the column.
+    return drainListStack(ctx, [typedCol](size_t, const db::ListElementView& view) {
+        typedCol->set(view);
+    });
```
cyrusknopf: Same comment

net/encoders/TuringProtoEncoder.h
cyrusknopf: Actually I just realised that getRaw and at(0) are not equivalent. If the column has `_empty = true` then `at` will throw whilst `getRaw` will return the value. I think we probably need to check before hand.

storage/list/ListWriteCursor.cpp
sulaimansuhas: Why does asserting that the tag size stay the same matter here, I'm not sure I understand?

storage/list/ListWriteCursor.cpp
sulaimansuhas: I'll add some comments but it is copying a ListElementView to our preserved ListElementView Buffer, and incrementing the pointer to the start of the next ListElementView object.

storage/columns/ColumnDispatcher.h
sulaimansuhas: These get hit by DataFrameToNumpy dispatcher code. What do you mean we should stop using this -is there something we should use instead?

net/encoders/TuringProtoEncoder.h
sulaimansuhas: yes that's fine, the binary format is designed such that the information needed to construct the object on the decoder need to be sent together - but there is no requirement that the data itself be sent in the same packet. We can just treat it like any variable length data split across packets ( in this case the offset of data received would just be 0).

net/turing_proto_common/TuringProtoDecoder.h
sulaimansuhas: No list and any variable length data can cross buffer boundaries. Generally it's fixed length individual components that are not split up - so say [tag][int].

storage/columns/ColumnDispatcher.h
cyrusknopf: Yes the `ColumnOperatorDispatcher` should be used, the one you use here can only support vectors

storage/list/ListWriteCursor.cpp
cyrusknopf: Whilst generally we have been defensive in being agnostic of the tag size, there are some assumptions that the tag size is 1 in the implementation. If it ever changes then everywhere the static assert fails is a place to check if that assumption being violated requires changes at that location

net/turing_proto_common/TuringProtoDecoder.h
cyrusknopf: So the comment is correct? And do you mean [tag] [int] can or cannot be split?

net/encoders/TuringProtoEncoder.h
```diff
@@ -333,7 +454,56 @@ class DataWriter {
...
+        while (!_stack.empty()) {
+            auto& [currIt, endIt] = _stack.top();
+            if (currIt == endIt) {
+                _stack.pop();
+                continue;
+            }
+
+            const auto it = currIt;
+            ++currIt;
+            db::ListTagDispatcher {it->getTag()}.execute(writeVisitor, *it);
```
cyrusknopf: How about 
```suggestion
            db::ListTagDispatcher {it->getTag()}.execute(writeVisitor, *currIt);
            ++currIt;
```
instead?

net/turing_proto_common/TuringProtoDecoder.h
sulaimansuhas: Yes the comment is correct. To clarify the splitting of data across packet boundaries a bit more:
- [tag][value] is always sent in one packet
- if [value] is a fixed length data type - int/bool/double/etc. it is sent in one packet.
- if [value] is a variable length data type - strings/embeddings/lists - which are generally of the form [datalengthinfo][...data...], then the guarantee is that [tag][datalengthinfo] are sent in one packet.

storage/list/ListWriteCursor.cpp
sulaimansuhas: but do you think that this is a location where the a change in the assumption of the size would require a change? Since we are just memcpying whatever size the type tags are defined to be.

storage/list/ListWriteCursor.cpp
sulaimansuhas: This function is meant to write raw bytes values read from the binary protocol so we don't really have any type info we can pass into here.

net/turing_proto_common/TuringProtoDecoder.h
cyrusknopf: Makes sense thanks

storage/list/ListWriteCursor.cpp
cyrusknopf: Is this method still used?

cyrusknopf (approved):
Just check if `ListWriteCursor::writeRaw` is still used, and then all good from me

storage/list/ListWriteCursor.cpp
sulaimansuhas: yes it is used

rjb32 (approved):

PR #668, opened 2026-06-08
Fix segfault when full-dumping a loaded graph

rjb32:
## Problem

When dumping a graph from scratch and some of its commits are not hydrated. Segfault in GraphDumper on the null WeakARCs.

## Fix

If we want to dump from scratch a loaded graph in another location, load all the commits when dumping.

### DIFF
--- a/storage/dump/GraphDumper.cpp
+++ b/storage/dump/GraphDumper.cpp
@@ -16,10 +16,29 @@
 
 using namespace db;
 
-DumpResult<void> GraphDumper::dump(const Graph* graph, const fs::Path& graphDir) {
+DumpResult<void> GraphDumper::dump(Graph* graph, const fs::Path& graphDir) {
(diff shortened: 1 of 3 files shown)


storage/dump/GraphDumper.cpp
```diff
@@ -16,11 +16,30 @@
 
 using namespace db;
 
-DumpResult<void> GraphDumper::dump(const Graph* graph, const fs::Path& graphDir) {
```
sulaimansuhas, line 19: This seems like a bit of code smell, that we are removing the constness of `Graph*` before we dump it. Perhaps the hydration step should be carried out before we call the dump() function?

rjb32: I agree. It's also a bit weird that it's not self-contained from a segfault-sensitivity perspective.

PR #669, opened 2026-06-08
Bulk-write property columns when dumping

rjb32:
Write properties by spans instead of one element at a time, while respecting page boundaries.

## Problem

We are currently dumping property containers and IDs element by element, which adds some call overhead for large graphs such as OGBN 100M.

## Approach

We dump by spans of values and IDs as much as we can:
- **Embedding floats**: one write per page, coalescing physically-contiguous embedding views and splitting only at bucket boundaries.
- **IDs** (trivial, string and embedding dumpers): one bulk write per page. We add static_asserts to check that EntityID does not add weird padding that would make this not possible.
- **Strings**: one bulk write per block. `StringLimits` is a packed `{uint32,uint32}` (cf. `StringBucket::create`'s `memcpy`).

## Measurements

OGBN-papers100M + 256-dim embedding vectors per node: Dump is 52% faster


REVIEW_FINDINGS.md
sulaimansuhas, line 1: Looks to be a left over file

rjb32: Thanks removed

PR #672, opened 2026-06-09
Implement an MLIR assembler for TuringDB dialects

rjb32:
* Added an assembler of MLIR in textual format into an MLIR module 
* Merged into the mlir sample tool
* mlir -d pretty prints the assembled module
* Just gives the "hello MLIR" module if no arguments


query/ir/dialect/db/CMakeLists.txt
```diff
@@ -42,9 +42,9 @@ add_mlir_library(DBDialect
...
-target_include_directories(DBDialect PUBLIC
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
-  $<BUILD_INTERFACE:${MLIR_INCLUDE_DIRS}>
-  $<BUILD_INTERFACE:${LLVM_INCLUDE_DIRS}>
+target_include_directories(turing_db_ir_dbdialect_s PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}
+  ${CMAKE_CURRENT_BINARY_DIR}
+  ${MLIR_INCLUDE_DIRS}
+  ${LLVM_INCLUDE_DIRS}
```
cyrusknopf, line 49: What's the difference between having `BUILD_INTERFACE` and not having it?

rjb32: That's additional categorization of the include directories that can be used in cmake but I think it's an unnecessary complication for us. It's better to have the simplest cmake file that does the job.

cyrusknopf: Yep agree looks good

query/ir/as/IRAssembler.cpp
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        // A parsed module is already verified to have unique symbol names, so a single
+        // file cannot redefine a function. Guard only against collisions across files.
+        const auto functions = llvm::to_vector(fileMod->getBody()->getOps<mlir::func::FuncOp>());
+        for (mlir::func::FuncOp func : functions) {
+            const llvm::StringRef name = func.getSymName();
+            const bool alreadyDefined = moduleSymbols.lookup(name) != nullptr;
+            if (alreadyDefined) {
+                throw CompilerException(fmt::format("Function '{}' from MLIR file {} is already defined in the module",
```
cyrusknopf: Should this be the new IRException? Or what was the intended use for the new exception

cyrusknopf (approved):
Looks good thanks for this

query/ir/as/IRAssembler.cpp
rjb32: Thanks, fixed

PR #673, opened 2026-06-09
First implementation of SystemAccessor

rjb32:
* Route the SystemManager API through SystemAccessor 
* Remove direct subsystems access in SystemManager
* Remove GraphSerializer

Besides introducing SystemAccessor as a proxy, this PR keeps intact the current API of SystemManager. The graph level and the change level will be done in another PR.


system/SystemAccessor.h
```diff
@@ -0,0 +1,113 @@
+#pragma once
+
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "Accessor.h"
+
+#include "GraphFileType.h"
```
rjb32: Fix order of includes, put this group last

PR #675, opened 2026-06-10
Support embeddings natively when importing from JSONL

cyrusknopf:
Before this PR, array properties in a JSONL file were converted to strings for storage when importing to TuringDB. This PR allows fully-numeric arrays to be imported as the `Embedding` property type via extra arguments to the `LOAD JSONL` query:

```
LOAD JSONL "reactome.jsonl" AS "reactome" WITH EMBEDDINGS [{"embeddingProperty", 3}]
```
which will attempt to load the property `embeddingProperty` as an embedding property with dimension `3` if seen in the JSONL file provided.


import/jsonl/JsonlParser.cpp
```diff
@@ -47,9 +51,43 @@ namespace {
...
 
+bool isEmbedding(const std::vector<json>& arr) {
+    // Embeddings must be at least size 2 (enforced in grammar)
+    return arr.size() > 1
+        && std::ranges::all_of(arr, [](const json& v) { return v.is_number(); });
 }
 
-namespace db {
+JsonlImportResult<void> tryFillEmbedding(const json& arr,
+                      std::vector<float>& storage,
```
rjb32: Align each argument with the the first letter of the first one

cyrusknopf: Fixed thanks

import/jsonl/JsonlParser.cpp
```diff
@@ -47,11 +51,49 @@ namespace {
...
+        storage.push_back(v);
+    }
+
+    return {};
 }
 
-namespace db {
+std::optional<size_t> tryGetEmbDim(std::string_view name,
+                                   const json& value,
+                                   const std::unordered_map<std::string_view, size_t>& embeddingSpecs) {
```
rjb32: is it really that interesting to use a hash table for embedding specs given that there will be 1 embedding property most of the time, 4 typically at most?

import/jsonl/JsonlParser.cpp
```diff
@@ -115,14 +159,23 @@ JsonlImportResult<void> JsonlParser::parse(ChangeAccessor& change, std::istream&
                 if (properties != obj.end()) {
                     for (const auto& [key, value] : properties->items()) {
                         ptName = key;
+
                         ValueType vt = ValueType::Invalid;
+                        std::optional<size_t> embDim = std::nullopt;
```
rjb32, line 166: embDim can be const

system/SystemManager.h
```diff
@@ -98,7 +100,7 @@ class SystemManager {
     void setDefaultGraph(std::string_view name);
 
     // Import graph from file
-    Graph* importGraph(const fs::Path& path, std::string_view graphName);
+    Graph* importGraph(const fs::Path& path, std::string_view graphName, const std::unordered_map<std::string_view, size_t>& embeddingSpecs);
```
rjb32: Independently of the data structure being used, it's better to give the type a name and make it its own class since it will be passed through layers

import/jsonl/JsonlParser.cpp
cyrusknopf: It is overwritten below

import/jsonl/JsonlParser.cpp
cyrusknopf: Each time an array property is encountered in the JSONL file, it is checked against this datastructure to see if its name exists in the map. For OGBN this is potentially 100 million times. I'm not sure if it is worth using a hash table as opposed to 100 million linear searches involving 3-4 string comparisons, but the only serious alternative is `vector<pair<string_view, size_t>`. Not sure what you think

system/SystemManager.h
cyrusknopf: I'm not sure where we can typedef this so that it is consistently used across all use cases. For example it is referenced in `query/plan/nodes`, `import/jsonl/`, etc. and I don't think there is a common parent. Where would you suggest typedeffing this?

import/jsonl/JsonlParser.cpp
rjb32: I see. Ok let's do hashing as a good default.

system/SystemManager.h
rjb32: I think it would be appropriate to put it in some import/ hierarchy. Maybe some import/common actually

cyrusknopf: I have changed all references of the `unordered_map` to the `EmbeddingsSpec` class, found at `import/common`

rjb32 (approved):

PR #676, opened 2026-06-10
MLIR: Add filtered columns variadic argument to db.get_out_edges

rjb32:
When we do `MATCH (a)-->(b)-->(c) RETURN a` in Cypher we don't return the result a of the scan nodes but the set of NodeIDs a such that they have successors b which in turn have successors c. So nodes are filtered at each get_out_edges or get_in_edges operation along the chain.

As discussed, add a variadic argument to pass the columns to filter by a `db.get_out_edges` operation in the db dialect.
```
func.func @main() {
    // MATCH (a)->(b)->(c): scan `a`, then two get_out_edges hops.
    %a = db.scan_nodes()

    // First hop a->b, nothing carried yet so the carry set is `{}`.
    %a1, %e0, %et0, %b = db.get_out_edges(%a, {})

    // Second hop b->c, carrying %a1 so it comes back as %a2 (the `a` that reach a `c`).
    %b2, %e1, %et1, %c, %a2 = db.get_out_edges(%b, {%a1})

    return
}
```

### DIFF
--- a/samples/mlir/sample.mlir
+++ b/samples/mlir/sample.mlir
@@ -1,7 +1,14 @@
 module {
   func.func @main() {
-    %0 = "db.scan_nodes"() : () -> !db.column<"scan">
-    %1:4 = "db.get_out_edges"(%0) : (!db.column<"scan">) -> (!db.column<"srcs">, !db.column<"eids">, !db.column<"etypes">, !db.column<"tgts">)
+    // MATCH (a)->(b)->(c): scan `a`, then two get_out_edges hops.
+    %a = db.scan_nodes() : !db.column<"a">
+
+    // First hop a->b, nothing carried yet so the carry set is `{}`.
+    %a1, %e0, %et0, %b = db.get_out_edges(%a, {}) : (!db.column<"a">) -> (!db.column<"a1">, !db.column<"e0">, !db.column<"et0">, !db.column<"b">)
+
+    // Second hop b->c, carrying %a1 so it comes back as %a2 (the `a` that reach a `c`).
+    %b2, %e1, %et1, %c, %a2 = db.get_out_edges(%b, {%a1}) : (!db.column<"b">, !db.column<"a1">) -> (!db.column<"b2">, !db.column<"e1">, !db.column<"et1">, !db.column<"c">, !db.column<"a2">)
+
     return
   }
 }
--- a/query/ir/dialect/db/DBOps.td
+++ b/query/ir/dialect/db/DBOps.td
@@ -12,16 +12,43 @@ include "DBTypes.td"
 def ScanNodes : TuringOp<"scan_nodes", [Pure]> {
   let summary = "Produce all nodes in the graph";
   let results   = (outs Column:$result);
-  let assemblyFormat = "`(`operands`)` attr-dict `:` type($result)";
+  // `qualified(...)` keeps the full `!db.column<...>` mnemonic on the result type
+  // rather than the stripped `<...>` form, matching how get_out_edges prints.
+  let assemblyFormat = "`(`operands`)` attr-dict `:` qualified(type($result))";
 }
 
 /**
 * GetOutEdges
 */
 def GetOutEdges : TuringOp<"get_out_edges", [Pure]> {
-  let summary = "Fetch node and edge successors of input";
-  let arguments = (ins Column:$input_nodes);
-  let results   = (outs Column:$srcids, Column:$eids, Column:$etypes, Column:$tgtids);
+  let summary = "Fetch node and edge successors of input, filtering carried columns alongside";
+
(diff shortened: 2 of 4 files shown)


samples/mlir/sample.mlir
```diff
@@ -1,7 +1,14 @@
...
-    %0 = "db.scan_nodes"() : () -> !db.column<"scan">
-    %1:4 = "db.get_out_edges"(%0) : (!db.column<"scan">) -> (!db.column<"srcs">, !db.column<"eids">, !db.column<"etypes">, !db.column<"tgts">)
+    // MATCH (a)->(b)->(c): scan `a`, then two get_out_edges hops.
+    %a = db.scan_nodes() : !db.column<"a">
+
+    // First hop a->b, nothing carried yet so the carry set is `{}`.
+    %a1, %e0, %et0, %b = db.get_out_edges(%a, {}) : (!db.column<"a">) -> (!db.column<"a1">, !db.column<"e0">, !db.column<"et0">, !db.column<"b">)
+
+    // Second hop b->c, carrying %a1 so it comes back as %a2 (the `a` that reach a `c`).
+    %b2, %e1, %et1, %c, %a2 = db.get_out_edges(%b, {%a1}) : (!db.column<"b">, !db.column<"a1">) -> (!db.column<"b2">, !db.column<"e1">, !db.column<"et1">, !db.column<"c">, !db.column<"a2">)
```
cyrusknopf, line 10: What do you think about a syntax such as 
```
%b2, %e1, %et1, %c, {%a2} = db.get_out_edges(%b, {%a1})
```
to make it explicit which return values the carried filter elements? In this case it is obvious because they are both named `aX`, and there is only one element in the `{}` context, but I imagine the context will grow quite large in some queries and so a braced return value syntax might aid readability.

cyrusknopf (approved):
Looks good. Do we have any idea how we will implement this implicit filtering on previous columns for `GetOutEdges` and beyond? We discussed how the filtering of the input source node column can be done, but not an arbitrary list of other columns as far as I remember.

samples/mlir/sample.mlir
rjb32: Yes that's a good idea

rjb32: Actually MLIR controls the printing of the everything left of the "=" sign so we can't write a custom format for the carried columns in the results. The solution would be to write an entirely custom pretty printer of the IR. And we would also have to write a custom parser as well if we want to be able to ingest that custom syntax in textual form.

rjb32:
> Looks good. Do we have any idea how we will implement this implicit filtering on previous columns for `GetOutEdges` and beyond? We discussed how the filtering of the input source node column can be done, but not an arbitrary list of other columns as far as I remember.

I think what will happen is that the filtered columns need to have the same length as the column being expanded.
That's why we pass %a1 in the second get_out_edges on %b, because %a1 is the srcIDs of the operation that produced %b, so %a1 has the same length as %b here.
```
// First hop a->b, nothing carried yet so the carry set is `{}`.
  %a1, %e0, %et0, %b = db.get_out_edges(%a, {})

  // Second hop b->c, carrying %a1 so it comes back as %a2 (the `a` that reach a `c`).
  %b2, %e1, %et1, %c, %a2 = db.get_out_edges(%b, {%a1})

```  
But is that always true or can we make it always true? I am not sure...

samples/mlir/sample.mlir
cyrusknopf: Okay if it is not as straight forward as just setting the assembly format I think we can forget it for now, and revisit if it becomes a readability problem

cyrusknopf:
> > Looks good. Do we have any idea how we will implement this implicit filtering on previous columns for `GetOutEdges` and beyond? We discussed how the filtering of the input source node column can be done, but not an arbitrary list of other columns as far as I remember.
> 
> I think what will happen is that the filtered columns need to have the same length as the column being expanded. That's why we pass %a1 in the second get_out_edges on %b, because %a1 is the srcIDs of the operation that produced %b, so %a1 has the same length as %b here.
> 
> ```
> // First hop a->b, nothing carried yet so the carry set is `{}`.
>   %a1, %e0, %et0, %b = db.get_out_edges(%a, {})
> 
>   // Second hop b->c, carrying %a1 so it comes back as %a2 (the `a` that reach a `c`).
>   %b2, %e1, %et1, %c, %a2 = db.get_out_edges(%b, {%a1})
> ```
> 
> But is that always true or can we make it always true? I am not sure...

They would be the same length as a result of the filtering, but I am wondering how we will implement that filtering. It would have to be something on top of `GetOutEdges` as that has no notion of filtering other columns. It might end up needing to be using the indices generated and applying those as a materialise to all columns in the context

PR #678, opened 2026-06-10
MLIR: First version of nested loop dialect

rjb32:
First version of the nested loop MLIR dialect corresponding to the db dialect. Heavily inspired by the dsa dialect of LingoDB.
We will use it to lower the db dialect that uses dataflow semantics to the nested loop dialect which is an actual imperative loop-based implementation.

```
MATCH (a)-->(b)-->(c) RETURN a,b,c
```

db dialect:
```
func.func @main() {
  // MATCH (a)->(b)->(c): scan `a`, then two get_out_edges hops.
  %a = db.scan_nodes() 

  // First hop a->b, nothing carried yet so the carry set is `{}`.
  %a1, %e0, %et0, %b = db.get_out_edges(%a, {})

  // Second hop b->c, carrying %a1 so it comes back as %a2 (the `a` that reach a `c`).
  %b2, %e1, %et1, %c, %a2 = db.get_out_edges(%b, {%a1})

  return
}
```

Nested loop implementation:
```
func.func @nested_loop() {
    %nodes = nl.scan_nodes()

    nl.for %chunk in %nodes {

        // MATCH (a)->(b)->(c)
        // First hop a->b with an empty carry set
        %h1 = nl.get_out_edges(%chunk, {})

       // %hsrcs = a, %hb = b
        nl.for %hsrcs, %he0, %het0, %hb in %h1 {
            // Second hop b->c
            %h2 = nl.get_out_edges(%hb, {%hsrcs})

            // %h2srcs = b, %hc = c, %hafilt = a
            nl.for %h2srcs, %he1, %het1, %hc, %hafilt in %h2 {
                nl.output(%hafilt, %h2srcs, %hc)
            }
        }
    }

    return
}
```

-> Nice assembly format for `nl.for` loops so that we can write `nl.for %chunk in %nodes`with a for..in syntax

## Types

* The db.dialect uses named db.column types such as in `%a = db.scan_nodes() : !db.column<"a">`
* For now, nl dialect uses parametric chunk types `nl.chunk<!nl.node_id>` 💀 maybe we should have the column names as well here

### DIFF
--- a/samples/mlir/main.cpp
+++ b/samples/mlir/main.cpp
@@ -85,10 +85,26 @@ void addNestedLoopFunction(mlir::OpBuilder& builder, mlir::ModuleOp& module) {
     auto nodeLoop = builder.create<mlir::nl::For>(loc, nodes.getResult());
     builder.create<mlir::func::ReturnOp>(loc);
 
-    // Fill the node loop body: iterate the out-edges of each chunk of nodes
+    // Fill the node loop body: walk each chunk of nodes both forwards
+    // (out-edges) and backwards (in-edges)
     builder.setInsertionPointToStart(nodeLoop.getBody());
-    auto edges = builder.create<mlir::nl::GetOutEdges>(loc, nodeLoop.getBody()->getArgument(0));
-    builder.create<mlir::nl::For>(loc, edges.getResult());
+    const mlir::Value nodeChunk = nodeLoop.getBody()->getArgument(0);
+
+    // Out-edges: send each step's target (successor) node IDs to the result.
+    // The edge loop binds the chunks in order: sources, edge IDs, edge type
+    // IDs, targets - so argument 3 is the targets column.
+    auto outEdges = builder.create<mlir::nl::GetOutEdges>(loc, nodeChunk);
+    auto outLoop = builder.create<mlir::nl::For>(loc, outEdges.getResult());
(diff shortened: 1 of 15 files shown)


samples/mlir/main.cpp
```diff
@@ -12,33 +12,130 @@
...
+
+    // MATCH (a)->(b)->(c): two out-edge hops where the second carries the `a`
+    // of the current step. First hop a->b with an empty carry set; its loop
+    // binds sources(=a) at argument 0 and targets(=b) at argument 3.
+    builder.setInsertionPointAfter(inLoop);
+    auto firstHop = builder.create<mlir::nl::GetOutEdges>(loc, nodeChunk, mlir::ValueRange {});
+    auto firstHopLoop = builder.create<mlir::nl::For>(loc, firstHop.getResult());
+    builder.setInsertionPointToStart(firstHopLoop.getBody());
+    const mlir::Value aChunk = firstHopLoop.getBody()->getArgument(0);
+    const mlir::Value bChunk = firstHopLoop.getBody()->getArgument(3);
```
cyrusknopf: What does getBody on this type return?

rjb32: It returns the first block of the region of the loop. I extracted the result of getBody() to not repeat the getter calls

PR #681, opened 2026-06-11
NL Subroutine Threaded interpreter

rjb32:
* NLTranslator: translated MLIR nl dialect modules into an NLProgram that contains loops of function pointers. This is to avoid doing dynamic dispatch of the MLIR operations at each iteration.
* NLExecutor: executes NLProgram using function pointers
* NLInterpreter: takes an MLIR module in nl dialect and wraps NLTranslator + NLExecutor

MLIR sample:
We can specify -exec and -graph to the mlir sample to execute an mlir program now on a graph:
```
./mlir -exec -graph $HOME/.turing/graphs/simpledb/ ~/turingdb4/samples/mlir/nested_loop.mlir
```

### DIFF
--- a/query/ir/interpreter/NLExecutor.cpp
+++ b/query/ir/interpreter/NLExecutor.cpp
@@ -1,4 +1,4 @@
-#include "NLInterpreter.h"
+#include "NLExecutor.h"
 
 #include <type_traits>
 
@@ -73,7 +73,7 @@ void runEdgeLoopSteps(NLExecutionContext* context,
 
 }
 
-NLInterpreter::NLInterpreter(const GraphView* view,
+NLExecutor::NLExecutor(const GraphView* view,
                              const NLProgram* prog,
                              NLOutputSink* sink)
     : _ctxt(view, sink, prog->getChunkSize()),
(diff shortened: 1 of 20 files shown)


query/ir/interpreter/NLExecutor.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+        // Gather either source or target if we are get_out or get_in (the source side)
+        gatherColumn<NodeID>(loopData->getInput(), indices, gatheredNodeIDs);
+
+        // Transform all the columns in the carried set according to indices
+        for (const NLCarriedColumn& carriedColumn : loopData->carriedColumns()) {
+            const auto gatherFunc = carriedColumn.getGatherFunc();
+            gatherFunc(carriedColumn.getInput(), indices, carriedColumn.getOutput());
+        }
+
```
cyrusknopf, line 69: What is the notion of "gather", and why do we materialise all columns w.r.t indices in every step?

rjb32: gather is basically the name of the materialise operation, it's the canonical name for people doing MapReduce. We will do better, this is simply to go somewhere and chunk the work

PR #688, opened 2026-06-18
Add get_node_properties and get_edge_properties to the db dialect and lower them

rjb32:
Adds db.get_node_properties/get_edge_properties ops that resolve a property name to its value type against the graph schema during lowering and fetch a nullable value chunk in the nl dialect.


query/ir/interpreter/NLProgram.h
```diff
@@ -169,6 +170,30 @@ class NLEdgeLoopData : public NLFunctionData {
...
+    }
+
+    const Column* getInput() const { return _input; }
+    Column* getOutput() const { return _output; }
+    PropertyTypeID getPropertyTypeID() const { return _propertyTypeID; }
+
+private:
+    const Column* _input {nullptr};
+    Column* _output {nullptr};
+    PropertyTypeID _propertyTypeID {0};
```
cyrusknopf: Should this be left to default initialise as an INVALID ID rather than 0?

rjb32: Indeed thanks

cyrusknopf (approved):

PR #689, opened 2026-06-19
Implement the `avg()` aggregate function

cyrusknopf:
Implements the `avg()` aggregate function in the pipeline.

### DIFF
--- a/query/pipeline/processors/AvgProcessor.cpp
+++ b/query/pipeline/processors/AvgProcessor.cpp
@@ -9,7 +9,6 @@
 
 #include "columns/AllowedKinds.h"
 #include "columns/ColumnOperatorDispatcher.h"
-#include "TypeUtils.h"
 
 #include "PipelineException.h"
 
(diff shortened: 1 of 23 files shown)


query/pipeline/processors/AvgProcessor.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    // e.g. AVG(n.score) where score is a non-nullable numeric column: accumulate all values
+    template <typename T>
+    void operator()(const ColumnVector<T>* typed) {
+        for (const T& item : *typed) {
+            _sum += static_cast<AvgProcessor::AvgType>(item);
+            _count++;
+        }
+    }
```
cyrusknopf, line 96: @rjb32 There is the possibility that this local sum overflows, even when the average value itself is representable. I'm not sure if we want to handle this/to what granularity

rjb32: For now it's ok for as long as the behavior is known. It seems to be current practice in analytical DBs such as Clickhouse or DuckDB to implenment average as a sum and count like this, acknowledging that overflow can happen.

rjb32 (approved):

query/pipeline/processors/AvgProcessor.cpp
cyrusknopf: Okay sounds good

PR #690, opened 2026-06-22
Implement Authorization Mechanism For TuringDB server

sulaimansuhas:
Implement a simple authorisation mechanism for TuringDB.

### DIFF
--- a/auth/Authenticator.cpp
+++ b/auth/Authenticator.cpp
@@ -1,0 +1,45 @@
+#include "Authenticator.h"
+
+#include <stddef.h>
+
+using namespace db;
+
+namespace {
+
+// Length-independent (modulo the early size check) comparison, to avoid leaking
+// the configured key through timing.
+bool constantTimeEquals(std::string_view a, std::string_view b) {
+    if (a.size() != b.size()) {
+        return false;
+    }
+
+    unsigned char diff = 0;
+    for (size_t i = 0; i < a.size(); i++) {
+        diff |= static_cast<unsigned char>(a[i]) ^ static_cast<unsigned char>(b[i]);
+    }
+
+    return diff == 0;
+}
+
+}
+
+Authenticator::Authenticator(const std::string& apiKey)
+    : _apiKey(apiKey),
+    _enabled(!apiKey.empty())
+{
+}
+
+Authenticator::~Authenticator() {
+}
+
+bool Authenticator::check(std::string_view presentedKey) const {
+    if (!_enabled) {
+        return true;
+    }
+
+    if (presentedKey.empty()) {
+        return false;
+    }
+
+    return constantTimeEquals(presentedKey, _apiKey);
+}
--- a/net/http_parser/HTTPParser.h
+++ b/net/http_parser/HTTPParser.h
@@ -5,6 +5,7 @@
 #include "UriParser.h"
 #include "NetBuffer.h"
 #include "HTTPWriter.h"
+#include "HTTPUtils.h"
 
 namespace net {
 
@@ -81,6 +82,7 @@ public:
         _payloadSize = 0;
         _payloadBegin = nullptr;
         _parsedHeader = false;
+        _contentLengthSeen = false;
     }
 
 private:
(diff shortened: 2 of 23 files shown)


sulaimansuhas:
@rjb32 A very simple implementation for now. I'm not sure if we wanted something a bit more complex as a first step?

rjb32:
> @rjb32 A very simple implementation for now. I'm not sure if we wanted something a bit more complex as a first step?

I think it looks good and that it's the most minimal solution for now until we get more complex requirements for it.

auth/Authenticator.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <stddef.h>
+
+using namespace db;
+
+namespace {
+
+// Length-independent (modulo the early size check) comparison, to avoid leaking
+// the configured key through timing.
+bool constantTimeEquals(std::string_view a, std::string_view b) {
```
rjb32, line 11: Good one

rjb32 (approved):

tools/turingdb/StartCmd.cpp
```diff
@@ -225,6 +227,23 @@ int StartCmd::execute() {
         LocalMemory mem;
         TuringDB turingDB(&config);
 
+        // Authentication is opt-in via -auth-on. When enabled, the API key is
+        // read from the TURINGDB_AUTH_TOKEN environment variable; clients
+        // present it directly in the Authorization header. The Authenticator
+        // must outlive the server, so it lives in this scope.
+        std::unique_ptr<Authenticator> authenticator;
```
rjb32: Authenticator is declared after TuringDB so on destruction it will be destroyed before TuringDB and the server so it will technically not outlive them

rjb32:
Please rebase against latest main locally to avoid the commit that removed the gitlink for external/llvm-project

net/http_parser/HTTPParser.h
```diff
@@ -219,91 +222,113 @@ class HTTPParser : public AbstractTCPParser {
...
-            _currentPtr++;
+            _contentLengthSeen = true;
+
+            // Parse digits with bounds checking (no strtoull — buffer is not
+            // null-terminated). Horner's rule; the overflow guard keeps
+            // _payloadSize * 10 well below the size_t bound because BUFFER_SIZE
+            // is far smaller.
+            _payloadSize = 0;
+            for (const char c : value) {
+                if (!isdigit(static_cast<unsigned char>(c))) {
```
rjb32, line 315: If content length has a non-digit value we break without error and _payloadSize will be zero

net/turing_proto_client/TuringClient.cpp
```diff
@@ -177,6 +185,13 @@ void TuringClient::sendRequest(const std::string& query) {
     headers += "Content-Length: ";
     headers += std::to_string(query.size());
     headers += "\r\n";
+
+    if (!_authToken.empty()) {
```
rjb32, line 189: The authToken is concatenated without validation so the user can inject a string that can corrupt the header

server/TuringProtoServerProcessor.cpp
```diff
@@ -44,6 +45,10 @@ void TuringProtoServerProcessor::process(net::AbstractThreadContext* threadConte
             throw ProtocolException("Only POST is supported by the binary protocol");
         }
 
+        if (!isRequestAuthorized(_db.getAuthenticator(), httpInfo)) {
```
rjb32, line 48: This throws ProtocolException after startResponse has been called

rjb32:
Please add unit tests for AuthGate

sulaimansuhas:
> Please rebase against latest main locally to avoid the commit that removed the gitlink for external/llvm-project

@rjb32 I think you're getting confused with the other PR (which has been fixed), this one doesn't do anything with external/llvm-project

server/TuringProtoServerProcessor.cpp
sulaimansuhas: so this is a design decision actually - right now I am returning 200 http ok response, but it will carry the turingdb error within it. Would you prefer this or a 401 Unauthorised response.

PR #692, opened 2026-06-22
MLIR: db.cross_product op and its lowering to the nl dialect

rjb32:
Add the db.cross_product op and lower it to a nested-loop. The nl.cross_product operation implements the cross product of chunks at the innermost point of the loop nest.

### DIFF
--- a/query/ir/interpreter/NLExecutor.cpp
+++ b/query/ir/interpreter/NLExecutor.cpp
@@ -203,11 +203,13 @@ void NLExecutor::runCrossProduct(NLExecutionContext* context, NLFunctionData* da
     const size_t innerRowCount = innerColumns.front().getInput()->size();
 
     for (const NLCrossColumn& column : outerColumns) {
-        column.getBroadcast()(column.getInput(), innerRowCount, column.getOutput());
+        const NLBroadcastFunction broadcast = column.getBroadcast();
+        broadcast(column.getInput(), innerRowCount, column.getOutput());
     }
 
     for (const NLCrossColumn& column : innerColumns) {
-        column.getBroadcast()(column.getInput(), outerRowCount, column.getOutput());
+        const NLBroadcastFunction broadcast = column.getBroadcast();
+        broadcast(column.getInput(), outerRowCount, column.getOutput());
     }
 }
 
--- a/query/ir/interpreter/NLExecutor.cpp
+++ b/query/ir/interpreter/NLExecutor.cpp
@@ -60,8 +60,6 @@ void blockRepeatColumn(const Column* input, size_t factor, Column* output) {
     auto& outputRaw = typedOutput->getRaw();
     outputRaw.resize(inputRaw.size() * factor);
 
-    // fill_n each input value into its `factor`-wide run; vectorises where the
-    // push_back loop would not.
     auto outputIt = outputRaw.begin();
     for (const ElementType& value : inputRaw) {
         std::fill_n(outputIt, factor, value);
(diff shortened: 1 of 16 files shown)


query/ir/interpreter/NLExecutor.cpp
```diff
@@ -45,6 +46,47 @@ void gatherColumn(const Column* input,
...
+    const ColumnVector<ElementType>* typedInput = static_cast<const ColumnVector<ElementType>*>(input);
+    ColumnVector<ElementType>* typedOutput = static_cast<ColumnVector<ElementType>*>(output);
+
+    typedOutput->clear();
+    typedOutput->reserve(typedInput->size() * factor);
+
+    auto& outputRaw = typedOutput->getRaw();
+    for (const ElementType& value : typedInput->getRaw()) {
+        for (size_t repeat = 0; repeat < factor; repeat++) {
+            outputRaw.push_back(value);
```
cyrusknopf: Using reserve + push back does not seem to get vectorised. Whilst `std::fill` or `std::fill_n` commonly does get vectorised. We use `fill` in the current `CartesianProductProcessor`, I think we should do the same here.

[godbolt](https://godbolt.org/)

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -45,6 +46,47 @@ void gatherColumn(const Column* input,
...
+
+    typedOutput->clear();
+    typedOutput->reserve(typedInput->size() * factor);
+
+    auto& outputRaw = typedOutput->getRaw();
+    const auto& inputRaw = typedInput->getRaw();
+    for (size_t repeat = 0; repeat < factor; repeat++) {
+        for (const ElementType& value : inputRaw) {
+            outputRaw.push_back(value);
+        }
```
cyrusknopf: As with the above use of `fill`, we can use `std::copy` which is commonly lowered to memcpy instead of this push back loop

query/ir/interpreter/NLExecutor.cpp
rjb32: Thank you, done

query/ir/interpreter/NLExecutor.cpp
rjb32: Thanks, done

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -45,6 +47,49 @@ void gatherColumn(const Column* input,
...
+void blockRepeatColumn(const Column* input, size_t factor, Column* output) {
+    const ColumnVector<ElementType>* typedInput = static_cast<const ColumnVector<ElementType>*>(input);
+    ColumnVector<ElementType>* typedOutput = static_cast<ColumnVector<ElementType>*>(output);
+
+    const auto& inputRaw = typedInput->getRaw();
+    auto& outputRaw = typedOutput->getRaw();
+    outputRaw.resize(inputRaw.size() * factor);
+
+    // fill_n each input value into its `factor`-wide run; vectorises where the
+    // push_back loop would not.
```
cyrusknopf: "vectorises where the push_back loop would not" is just noise

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -45,6 +47,49 @@ void gatherColumn(const Column* input,
...
+void tileColumn(const Column* input, size_t factor, Column* output) {
+    const ColumnVector<ElementType>* typedInput = static_cast<const ColumnVector<ElementType>*>(input);
+    ColumnVector<ElementType>* typedOutput = static_cast<ColumnVector<ElementType>*>(output);
+
+    const auto& inputRaw = typedInput->getRaw();
+    auto& outputRaw = typedOutput->getRaw();
+    outputRaw.resize(inputRaw.size() * factor);
+
+    // Copy the whole input chunk into each `factor` repeat; copy commonly lowers
+    // to a memcpy whereas the push_back loop would not.
```
cyrusknopf: "copy commonly lowers to a memcpy whereas the push_back loop would not" is just noise

query/ir/interpreter/NLExecutor.cpp
rjb32: Done

query/ir/interpreter/NLExecutor.cpp
rjb32: Done

PR #695, opened 2026-06-23
MLIR: Implement the ADD operation

cyrusknopf:


query/ir/dialect/db/DBOps.td
```diff
@@ -137,4 +137,13 @@ def Output : TuringOp<"output"> {
   let assemblyFormat = "`(` $columns `)` attr-dict `:` qualified(type($columns))";
 }
 
+def AndOp : BinaryOp<"logical_and", [Pure]> {
```
rjb32: You can just name these "or" and "and"

query/ir/dialect/db/DBDialect.td
```diff
@@ -22,4 +22,9 @@ def DB : Dialect {
 class TuringOp<string mnemonic, list<Trait> traits = []>
     : Op<DB, mnemonic, traits>;
 
+// TODO: Not sure if this top-level, generic op is the best
```
rjb32, line 28: I think for now it's fine

rjb32:
So far it makes sense. Please add sample programs implementing:
* Expressions implemented at the db and nl level over constants
* Expressions mixing constants and columns obtained from get_node_properties

query/ir/lowering/DBLowering.cpp
```diff
@@ -105,6 +106,50 @@ mlir::Type aggregateResultElementType(mlir::OpBuilder& builder,
...
+    const bool anyFloat = mlir::isa<mlir::Float64Type>(lhs) || mlir::isa<mlir::Float64Type>(rhs);
+    if (anyFloat) {
+        return builder.getF64Type();
+    }
+
+    const auto lhsInteger = mlir::cast<mlir::IntegerType>(lhs);
+    const auto rhsInteger = mlir::cast<mlir::IntegerType>(rhs);
+    const bool anyUnsigned = lhsInteger.isUnsigned() || rhsInteger.isUnsigned();
+    if (anyUnsigned) {
+        return builder.getIntegerType(64, /*isSigned=*/false);
```
rjb32, line 148: Will this promotion work is in the case of -5+3? It will promote the result which is negative to unsigned which will be wrong?

cyrusknopf: We use the same promotion rules as C++. This is the case in the current pipeline, verified with:

~~~cpp
using C = ColumnCombination<Add, ColumnVector<types::Int64::Primitive>,
                            ColumnVector<types::UInt64::Primitive>>::ResultColumnType;
static_assert(std::is_same_v<C, ColumnVector<types::UInt64::Primitive>>); // true
~~~

rjb32: Ok I see. For now it's ok indeed. That's more the rules themselves that we may want to revisit in the future to avoid potential unintended promotions such that someUint64Property + (-5)

rjb32 (approved):

query/ir/lowering/DBLowering.cpp
```diff
@@ -1103,6 +1150,46 @@ void DBLowering::lowerConstant(mlir::db::ConstantOp constant) {
...
+        return firstBlock;
+    }
+
+    if (firstBlock == _entryBlock) {
+        return secondBlock;
+    } else if (secondBlock == _entryBlock) {
+        return firstBlock;
+    }
+
+    throw IRException("db.add operands must be bound in the same loop");
```
rjb32: This function is called deeperBlock in general and seems to be a generic utility. Still the exception message talks about db.add specifically

cyrusknopf: Fixed thanks

PR #696, opened 2026-06-24
MLIR: implement limit

rjb32:
Implement the db.limit op and its lowering into the nl dialect and execution.

## Example
```
MATCH (a) RETURN a LIMIT 3
``` 

db dialect:
```
func.func @main() {
  %a  = db.scan_nodes()
  %la = db.limit(%a) count 3
  db.output(%la)
  return
}
``` 

 nl dialect:
```
func.func @main() {
  %lim = nl.limit(3)                                          // counter, hoisted above the loop
  %scan = nl.scan_nodes()
  nl.for %chunk in %scan limit %lim {
    nl.limit_update %lim, %chunk
    nl.output(%chunk) limit %lim
  }
  return
}
``` 

In the NL dialect we avoid copying the columns just for the limit, when the limit feeds the output directly, by making the nl.output sink limit-aware.
- A limit handle is created above the loop with the limit budget
- nl.output sink is limit aware and receives the limit handle. An output sink is responsible for outputting colums only up to the remaining limit budget.
- nl.limit_update updates the remaining limit budget based on the size of the current chunk at a given iteration
- loops stop early by being aware of the limit state
- cross product is limit aware to avoid generating CxC where C is the chunk size for small-ish limits
- nl.limit_truncate truncates chunks according to the budgeted prefix so a mid-query limit (WITH ... LIMIT) can feed a downstream operation
- when nl.limit_truncate is immediately followed by nl.output, the limit is folded into nl.output directly to not make a copy if not necessary
- each db.limit gets its own handle, so multiple independent limits in one query don't interfere


query/ir/dialect/db/DBOps.td
```diff
@@ -216,4 +216,41 @@ def Output : TuringOp<"output"> {
...
+  let summary = "Take the first N rows of a column dataflow";
+  let description = [{
+    The declarative counterpart of a Cypher RETURN ... LIMIT N. Caps the result
+    to the first N rows of the columns flowing into db.output, passing them
+    through unchanged:
+
+      %a  = db.scan_nodes() : !db.column<"a">
+      %la = db.limit(%a) count 3 : (!db.column<"a">) -> !db.column<"a">
+      db.output(%la) : !db.column<"a">
+
```
cyrusknopf, line 232: Does this implementation require that the subsequent operation after the limit must be a db.output, as this comment suggests? I am asking because we might something arbitrary to come after the limit, so that we can support LIMITs in the middle of queries

query/ir/lowering/DBLowering.cpp
```diff
@@ -211,9 +235,13 @@ void DBLowering::lowerCrossProduct(mlir::db::CrossProduct product) {
...
     setInsertionInto(innerBody);
 
+    // _limitHandle is null when no db.limit is active (the product is built in
+    // full); when active, the cross carries the same handle every loop in the
+    // nest does, so it builds only the prefix the limit can emit this step.
     nl::CrossProduct cross = _builder.create<nl::CrossProduct>(_builder.getUnknownLoc(),
                                                                outerColumns,
-                                                               innerColumns);
+                                                               innerColumns,
+                                                               _limitHandle);
```
cyrusknopf: Is this to say that the limit effectively sets the chunk size of every instruction to that of the limit count?

query/ir/lowering/DBLowering.cpp
```diff
@@ -300,24 +328,70 @@ mlir::Type DBLowering::propertyValueChunkType(llvm::StringRef propertyName) {
...
+
 void DBLowering::lowerOutput(mlir::db::Output output) {
     llvm::SmallVector<mlir::Value, 4> columns;
+
+    // The limit governs this output only if it consumes a db.limit result. A
+    // db.limit passes its columns straight through, so a consumed result is a
+    // column whose defining op is the db.limit. A second db.output that does not
+    // consume it (a disconnected projection) stays unbounded and emits in full,
+    // rather than being clamped to an unrelated limit's count.
+    bool consumesLimit = false;
```
cyrusknopf: In what case can there be a second db.output? I thought that was the single sink

query/ir/dialect/db/DBOps.td
rjb32: Yes this was not supporting nested queries. Fixed in this new version where I introduce an explicit nl.limit_truncate operation

query/ir/lowering/DBLowering.cpp
rjb32: Only the cross product is limit-aware today.

query/ir/lowering/DBLowering.cpp
rjb32: This is hypothetical today because not explicitly forbidden in the db dialect.

query/ir/lowering/DBLowering.cpp
cyrusknopf: But does it set the chunk size to that of the limit?

query/ir/lowering/DBLowering.cpp
```diff
@@ -418,6 +424,65 @@ void DBLowering::assignProducerLoops(mlir::Value column, mlir::Value handle) {
...
+            if (!result.hasOneUse()) {
+                return;
+            }
+
+            nl::Output user = mlir::dyn_cast<nl::Output>(*result.user_begin());
+            if (!user || (output && output != user)) {
+                return;
+            }
+            output = user;
+        }
```
cyrusknopf: This is comment is very subjective to how I read this code, so you may not have a problem here, but:

1. The comment appears misleading to me, it seems to describe the direction of the implication as "column is truncated => has one user && that user is the same as every other truncated column". However, the actual direction of the implication is the reverse: "column has one user && all other columns share that user => columns can be truncated". In my opinion the comment should be rephrased

2. The loop iterates through all results columns to check for the above implication. For me it would be much easier to understand if it used an `all_of` pattern, which would declare what is actually being checked here. You would also be able to do away with the complicated if statement in the loop. Something like:

```cpp
auto fstUser  = results.front().users().front() // get the user of one result
bool canTruncate = std::ranges::all_of(results, [&](auto&& result) {
    if (!result.hasOneUse()) return false; // ensure single user
    if (auto user = dyn_cast<Output>; !user) return false; // ensure user is output
    if (user != fstUser) return false; // not all results have the same user
});
```

query/ir/lowering/DBLowering.cpp
```diff
@@ -418,6 +424,65 @@ void DBLowering::assignProducerLoops(mlir::Value column, mlir::Value handle) {
...
+        const mlir::OperandRange outputColumns = output.getColumns();
+        if (outputColumns.size() != results.size()) {
+            return;
+        }
+
+        for (size_t columnIndex = 0; columnIndex < results.size(); columnIndex++) {
+            if (outputColumns[columnIndex] != results[columnIndex]) {
+                return;
+            }
+        }
```
cyrusknopf: Could you give an example of a case where this check succeeds and a case where it fails? I can't understand what it is checking for, specifically why the order of the columns need be identical

query/ir/lowering/DBLowering.cpp
```diff
@@ -418,6 +424,65 @@ void DBLowering::assignProducerLoops(mlir::Value column, mlir::Value handle) {
     }
 }
 
+void DBLowering::foldTruncatesIntoOutputs(mlir::func::FuncOp nlFunction) {
```
cyrusknopf, line 427: Aside from my other comments, what motivates us doing this fold at lowering time rather than as an optimisation pass? I am thinking about the balance of how complex the initial lowering is vs splitting the complexity across multiple, isolated optimisation passes

query/ir/lowering/DBLowering.cpp
rjb32: It's because you can't just eliminate the copy of the limit operation if nl.output also consumes some other columns than the one being limited, for some reason. So for the conversion to be valid it's better to check that nl.output contains only the columns being limited

query/ir/lowering/DBLowering.cpp
rjb32: I agree it's a tradeoff between doing the folding in the lowering vs adding the optimisation infrastructure now

query/ir/lowering/DBLowering.cpp
rjb32: It fills it up to the limit remaining capacity

query/ir/lowering/DBLowering.cpp
rjb32: The comment has been updated

PR #700, opened 2026-06-26
Generate an MLIR DB program from a VariableDependencyGraph

cyrusknopf:
Uses a VariableDependencyGraph of a Cypher query to generate an MLIR program in the DB dialect for later lowering and execution.

- Supports queries with `MATCH` and `RETURN` clauses where
    - the variables being matched over form a single connected component (e.g. there is no need to insert a cross product)
    - there are no "merge" edges in the VDG (meaning no join is required)
    - there are no `WHERE` clauses


query/ir/codegen/DBProgramGenerator.h
```diff
@@ -0,0 +1,66 @@
...
+class OpBuilder;
+}
+
+namespace db {
+
+class CypherAST;
+class VariableDependency;
+
+class DBProgramGenerator {
+public:
```
rjb32, line 25: This class has non-trivial members, please add constructor and destructor defined in cpp file

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -0,0 +1,383 @@
...
+        const size_t resultIndex = GET_X_EDGES_RES_SIZE + i;
+        registerValue(carried[i], op.getResult(resultIndex));
+    }
+}
+
+void DBProgramGenerator::generate(const CypherAST* ast, mlir::ModuleOp* module) {
+    _mlirCtxt = module->getContext();
+    bioassert(_mlirCtxt, "Null context");
+
+    mlir::OpBuilder builder(module->getBodyRegion());
```
rjb32: Why not allocating the member at the class level instead of storing a pointer to something on the local stack? Or change the interface of the class so that ast and module are taken in the constructor?

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -0,0 +1,383 @@
...
+    std::vector<const VariableDependency*> carriedSet;
+
+    // TODO: Use nodes at ends of diameter
+    for (const VariableDependency& root : _vdg.vars()) {
+        if (defined.contains(&root)) {
+            continue;
+        }
+
+        const auto isEdgeTgtEdgeVar = [](const DependencyEdge* e) -> bool {
+            return e->data().type() == EdgeMetadata::EdgeType::GET_OUT_EDGES
```
rjb32: Is it kind of what producedEdgeVar already does?

query/ir/frontend/cypher/VariableDependencyGraph.cpp
```diff
@@ -8,48 +8,92 @@
...
+
+    const QueryCommand* q = ast->queries().front();
+    const auto* spq = dynamic_cast<const SinglePartQuery*>(q);
+    bioassert(spq, "Non-SinglePartQueries are not yet supported.");
+
+    const StmtContainer* stmtsContainer = spq->getReadStmts();
+    const StmtContainer::Stmts& stmts = stmtsContainer->stmts();
+
+    for (Stmt* s : stmts) {
+        const auto* rd = dynamic_cast<const MatchStmt*>(s);
```
rjb32: Use more explicit names that rd, ptn and ele

query/ir/codegen/DBProgramGenerator.h
cyrusknopf: Done

query/ir/codegen/DBProgramGenerator.cpp
cyrusknopf: You're right thanks it was a remnant of draft implementation. I've made the constructor take the module and made the class allocate the builder rather than store a pointer

query/ir/codegen/DBProgramGenerator.cpp
cyrusknopf: Ah yes good catch I will reuse that

query/ir/frontend/cypher/VariableDependencyGraph.cpp
cyrusknopf: Fixed

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -0,0 +1,386 @@
...
+                case EdgeMetadata::EdgeType::_SIZE:
+                    throw FatalException("Attempted to translate invalid edge.");
+                break;
+            }
+
+            defined.insert(src);
+            defined.insert(edge);
+            defined.insert(tgt);
+
+            carriedSet.push_back(src);
```
rjb32, line 346: src is added to the carried systematically. We may have too many things carried in the carried set

test/query/ir/EquivalenceTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+    expectEquivalent("MATCH (a)<--(b)<--(c) RETURN c");
+    expectEquivalent("MATCH (a)<--(b)<--(c) RETURN a, b, c");
+    expectEquivalent("MATCH (a)<--(b)<--(c) RETURN a, c");
+}
+
+TEST_F(EquivalenceTest, trees) {
+    expectEquivalent("MATCH (a)-->(b), (a)-->(c) RETURN a");
+    expectEquivalent("MATCH (a)-->(b), (a)-->(c)-->(d) RETURN a");
+    // Disabled due to suspected v2 bug
+    // expectEquivalent("MATCH (a)-->(b)-->(c), (b)-->(e) RETURN a");
```
rjb32, line 264: We should find another way to test that

cyrusknopf: Added CypherOutputTest for this and other queries

query/ir/frontend/cypher/VariableDependencyGraph.cpp
```diff
@@ -8,48 +8,92 @@
 #include <unordered_set>
 #include <utility>
 
-#include "DependencyEdge.h"
-#include "EdgeMetadata.h"
+#include <spdlog/spdlog.h>
```
rjb32, line 11: Better to include just the spdlog format header

query/ir/frontend/cypher/VariableDependencyGraph.cpp
```diff
@@ -8,48 +8,92 @@
...
         break;
     }
     throw FatalException("Invalid edge pattern direction");
 }
 
+static EdgeMetadata::EdgeType edgeTypeToNodeType(EdgeMetadata::EdgeType t) {
+    if (t == EdgeMetadata::EdgeType::GET_OUT_EDGES) {
+        return EdgeMetadata::EdgeType::GET_EDGE_TGT;
+    }
+    if (t == EdgeMetadata::EdgeType::GET_IN_EDGES) {
```
rjb32: Please use else if here to chain if statements that are cases of the same value

query/ir/frontend/cypher/VariableDependencyGraph.cpp
```diff
@@ -8,48 +8,92 @@
...
+    bioassert(ast->queries().size() == 1, "Single queries only.");
+
+    const QueryCommand* query = ast->queries().front();
+    const auto* spq = dynamic_cast<const SinglePartQuery*>(query);
+    bioassert(spq, "Non-SinglePartQueries are not yet supported.");
+
+    const StmtContainer* stmtsContainer = spq->getReadStmts();
+    const StmtContainer::Stmts& stmts = stmtsContainer->stmts();
+
+    for (Stmt* stmt : stmts) {
```
rjb32: Why stmt is not a const pointer here?

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -0,0 +1,412 @@
...
+        operands.push_back(column);
+        results.push_back(column.getType());
+    }
+
+    const auto loc = _opBuilder.getUnknownLoc();
+    auto op = _opBuilder.create<EdgeOp>(loc, results, operands);
+
+    const mlir::Value newSrcs = op.getResult(0);
+    const mlir::Value newEdges = op.getResult(1);
+    [[maybe_unused]] const mlir::Value newEdgeTypes = op.getResult(2);
```
rjb32: newEdgeTypes is actually unused?

query/ir/frontend/cypher/VariableDependencyGraph.cpp
cyrusknopf: I use `spdlog::warn` to print an error for the sample

query/ir/frontend/cypher/VariableDependencyGraph.cpp
cyrusknopf: Done

query/ir/frontend/cypher/VariableDependencyGraph.cpp
cyrusknopf: Fixed thanks

query/ir/codegen/DBProgramGenerator.cpp
cyrusknopf: Replaced with a TODO comment

rjb32 (approved):

PR #708, opened 2026-07-07
Parquet import: tests code review findings

rjb32:
Findings addressed:
1. String property values corrupted across data-page boundaries — capturePropertyByteArray keeps only the last page's values (test)
2. Node with an empty/null __labels list — **WRONG BY INVARIANT**: every node has at least one label, so an empty label list is not representable
3. Nullable __type on edges produces fewer EdgeTypeIDs than rows (test) **WRONG BY INVARIANT**
4. LIST-typed property column — **WRONG BY INVARIANT**: lists are query-language-only, never a stored property container
5. Wrong-typed required column trips an internal assertion instead of a clean error (test)


io/parquet/ParquetEdgeVisitor.cpp
```diff
@@ -14,6 +14,7 @@
 #include <parquet/types.h>
 
 #include "ID.h"
+#include "TuringException.h"
```
rjb32: Please put this include below

PR #710, opened 2026-07-07
MLIR: sum/min/max/avg aggregate ops for the db and nl dialects

rjb32:
Adds db.sum/db.min/db.max/db.avg and the parameterized nl.aggregate/aggregate_update/aggregate_result family they lower to, with the null-aware interpreter, dialect verifiers, and unit tests.

### DIFF
--- a/query/ir/lowering/DBLowering.cpp
+++ b/query/ir/lowering/DBLowering.cpp
@@ -74,28 +74,31 @@ mlir::Type aggregateResultElementType(mlir::OpBuilder& builder,
     const bool isString = mlir::isa<storage::StringType>(inputElement);
 
     switch (kind) {
-        case storage::AggregateKind::Sum:
+        case storage::AggregateKind::Sum: {
             if (!isNumeric) {
                 throw IRException("db.sum requires a numeric column");
             }
             return inputElement;
+        }
         break;
 
-        case storage::AggregateKind::Avg:
+        case storage::AggregateKind::Avg: {
             if (!isNumeric) {
                 throw IRException("db.avg requires a numeric column");
             }
             return builder.getF64Type();
+        }
         break;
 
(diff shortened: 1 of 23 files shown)


query/ir/lowering/DBLowering.cpp
```diff
@@ -58,6 +58,50 @@ mlir::Type valueTypeToElementType(mlir::OpBuilder& builder, ValueType valueType)
...
+    const bool isFloat = mlir::isa<mlir::Float64Type>(inputElement);
+    const auto integerType = mlir::dyn_cast<mlir::IntegerType>(inputElement);
+    const bool isBool = integerType && integerType.getWidth() == 1;
+    const bool isInteger = integerType && !isBool;
+    const bool isNumeric = isFloat || isInteger;
+    const bool isString = mlir::isa<storage::StringType>(inputElement);
+
+    switch (kind) {
+        case storage::AggregateKind::Sum:
+            if (!isNumeric) {
```
rjb32, line 78: Use braces if the case code is more than one line

test/query/ir/DBLoweringTest.cpp
```diff
@@ -152,6 +152,50 @@ class CollectingCountSink : public NLOutputSink {
...
+        const auto* values = dynamic_cast<const ColumnOptVector<int64_t>*>(chunks[0]);
+        ASSERT_NE(values, nullptr);
+
+        const auto& raw = values->getRaw();
+        for (size_t rowIndex = offset; rowIndex < offset + rowCount; rowIndex++) {
+            _values.push_back(raw[rowIndex]);
+        }
+    }
+
+    const std::vector<std::optional<int64_t>>& getValues() const { return _values; }
```
rjb32: This is a bit disgusting. Use a type alias

PR #711, opened 2026-07-07
Vector: use exact k-NN search instead of approximate LSH routing

rjb32:
I unplugged for now the approximate search implementation because it is way to coarse and gets very bad recall. The approximate implementation is kept to be revised soon.

### DIFF
--- a/vector/VectorStorageManager.cpp
+++ b/vector/VectorStorageManager.cpp
@@ -96,9 +96,9 @@ VectorResult<void> VectorStorageManager::createLibraryStorage(const VecLib& lib)
 }
 
 VectorResult<void> VectorStorageManager::persistShardRouter(const VecLib& lib) {
-    const auto* meta = lib.metadata();
-
     std::shared_lock lock(_mutex);
+
+    const auto* meta = lib.metadata();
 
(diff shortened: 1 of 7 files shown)


vector/VectorStorageManager.cpp
```diff
@@ -95,6 +95,23 @@ VectorResult<void> VectorStorageManager::createLibraryStorage(const VecLib& lib)
     return {};
 }
 
+VectorResult<void> VectorStorageManager::persistShardRouter(const VecLib& lib) {
+    const auto* meta = lib.metadata();
+
+    std::shared_lock lock(_mutex);
+
```
cyrusknopf, line 100: Can `metadata` be modified in a way by another thread between lines 99 to 101 (in between reading `meta` and acquiring the lock) that would make the below invalid?

rjb32: That is just acquiring the metadata pointer, not reading the metadata so that should be ok. But it's more proper if I move the statement after the lock.

rjb32: Moved `const auto* meta = lib.metadata();` below the `std::shared_lock` in 20693d893.

PR #712, opened 2026-07-07
Vector: yield score alongside ids from VECTOR SEARCH

rjb32:
Add a score to the VECTOR SEARCH query:
```
VECTOR SEARCH IN vdb FOR 3 (1.0, 0.0, 0.0, 0.0) YIELD ids, score RETURN ids, score
``` 
so that we can sort matches by similarity score

### DIFF
--- a/query/pipeline/processors/VectorSearchProcessor.cpp
+++ b/query/pipeline/processors/VectorSearchProcessor.cpp
@@ -86,11 +86,14 @@ void VectorSearchProcessor::execute() {
     const std::span<const int64_t> ids = result.ids();
     colIds->getRaw().assign(ids.begin(), ids.end());
 
+    // There is a single metric column: the score. The search result reports it as
+    // a distance (squared L2 for EUCLID, inner product for cosine), which is the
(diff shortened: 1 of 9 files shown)


query/pipeline/processors/VectorSearchProcessor.cpp
```diff
@@ -79,13 +79,19 @@ void VectorSearchProcessor::execute() {
...
     using ColumnInt = ColumnVector<types::Int64::Primitive>;
     ColumnInt* colIds = _outIds.getValues()->as<ColumnInt>();
 
     const std::span<const int64_t> ids = result.ids();
     colIds->getRaw().assign(ids.begin(), ids.end());
 
+    using ColumnDouble = ColumnVector<types::Double::Primitive>;
+    ColumnDouble* scoreColumn = _scoreColumn->as<ColumnDouble>();
+
+    const std::span<const float> distances = result.distances();
```
rjb32: Why is there both a score and a distance column here?

cyrusknopf (approved):

sulaimansuhas (approved):

PR #713, opened 2026-07-07
MLIR: add branched two-hop sample with carry sets

rjb32:
Adds a `db` and `nl` sample pair for 
```cypher
MATCH (a)->(b)->(c), (b)->(d) RETURN a
``` 
to pin down the carry-set handling behind wrong `a` counts on simpledb when compared to current pipeline.

* `branch.mlir` is the `db` form (scan + three `get_out_edges` hops) and `branch.nl.mlir` is its nested-loop lowering
* verified byte-for-byte against `mlir -dump-lowered branch.mlir`. 
* The branch hop takes `%src1` (the second hop's source = `b`), not `%tgt1` (=`c`), so it leaves from `b`, and it carries `a` through the carry set `{%a2, %tgt1}` — `RETURN a` must emit the branch-filtered `%a3`, not the pre-branch `%a2`.

Counts on simpledb: correct (`%a3`) = 42 rows; the bug (`%a2`) = 12. 42 = b=Remy (2 preds × 4 c × 4 d = 32) + b=Adam (1×3×3 = 9) + b=Ghosts (1×1×1 = 1). Returning `%a2` drops the branch's ×d multiplicity, which is the undercount to look for in the frontend's projection of the branch pattern.

### DIFF
--- a/samples/mlir/branch.mlir
+++ b/samples/mlir/branch.mlir
@@ -1,0 +1,35 @@
+module {
+  func.func @main() {
+    // MATCH (a)->(b)->(c), (b)->(d) RETURN a
+    //
+    // A nested two-hop chain a->b->c with a branch b->d hanging off `b`. The
+    // catch is the branch: it does NOT continue from the chain tip `c`, it
+    // reconnects to `b`. And because the query returns `a`, the `a` we emit must
+    // be the one filtered/replicated through *both* the chain and the branch -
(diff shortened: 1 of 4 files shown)


cyrusknopf:
Is this still necessary?

rjb32:
> Is this still necessary?

Yes it's a good idea to have it in the base of tests to keep the example

PR #717, opened 2026-07-08
Migrate non query endpoints to procedures.

sulaimansuhas:


procedures/GetEdgesProcedure.cpp
```diff
@@ -0,0 +1,141 @@
...
+    auto* typeCol = static_cast<EdgeTypeIDCol*>(data.getReturnColumn(3));
+    auto* propsCol = static_cast<StringColumn*>(data.getReturnColumn(4));
+
+    switch (proc->getStep()) {
+        case ProcedureState::Step::PREPARE:
+        case ProcedureState::Step::RESET:
+        break;
+
+        case ProcedureState::Step::EXECUTE: {
+            const GraphView& view = *ctxt->getGraphView();
```
rjb32: The execute would deserve its own function

procedures/GetEdgesProcedure.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+            if (typeCol) {
+                typeCol->clear();
+            }
+
+            if (propsCol) {
+                propsCol->clear();
+            }
+
+            for (const int64_t rawID : edgeIDs) {
```
rjb32: Why do we use int64_t here?

procedures/GetNodeEdgesProcedure.cpp
```diff
@@ -0,0 +1,236 @@
...
+using ListConst = ColumnConst<ListView>;
+using IntConst = ColumnConst<int64_t>;
+using ItemVariant = ListBuffer<>::ListItemVariant;
+using LimitMap = std::unordered_map<EdgeTypeID, size_t>;
+using CountMap = std::unordered_map<EdgeTypeID, size_t>;
+
+struct Data : public ProcedureData {};
+
+// Build an {edgeTypeID -> limit} map from parallel (types, values) list args.
+LimitMap buildLimitMap(const ListConst* typesArg, const ListConst* valuesArg) {
```
rjb32: Don't return maps by value. Should not happen as per CLAUDE.md

procedures/GetNodeEdgesProcedure.cpp
```diff
@@ -0,0 +1,236 @@
...
+    LimitMap limits;
+    const std::vector<int64_t> types = ProcUtils::readIntList(typesArg);
+    const std::vector<int64_t> values = ProcUtils::readIntList(valuesArg);
+    for (size_t i = 0; i < types.size() && i < values.size(); ++i) {
+        limits[EdgeTypeID {static_cast<EdgeTypeID::Type>(types[i])}] =
```
rjb32: Do the cast in a separate local variable first

procedures/GetNodeEdgesProcedure.cpp
```diff
@@ -0,0 +1,236 @@
...
+            static_cast<size_t>(values[i]);
+    }
+    return limits;
+}
+
+// JSON-encode an {edgeTypeID -> count} map as {"typeId": count, ...}.
+std::string countsToJson(const CountMap& counts) {
```
rjb32: Don't return the string by value. Pass by result-reference

procedures/GetNodeEdgesProcedure.cpp
```diff
@@ -0,0 +1,236 @@
...
+    auto* outCountsCol = static_cast<StringColumn*>(data.getReturnColumn(3));
+    auto* inCountsCol = static_cast<StringColumn*>(data.getReturnColumn(4));
+
+    switch (proc->getStep()) {
+        case ProcedureState::Step::PREPARE:
+        case ProcedureState::Step::RESET:
+        break;
+
+        case ProcedureState::Step::EXECUTE: {
+            const GraphView& view = *ctxt->getGraphView();
```
rjb32: All the executes could be their own function

procedures/GetNodeEdgesProcedure.cpp
```diff
@@ -0,0 +1,236 @@
...
+                    }
+                    outerItems.emplace_back(listBuffer->insert(edgeItems));
+                }
+                return listBuffer->insert(outerItems);
+            };
+
+            CountMap outCounts;
+            CountMap inCounts;
+
+            for (const int64_t rawID : nodeIDs) {
```
rjb32: Same thing for int64

procedures/HierarchicalLabelCountsProcedure.cpp
```diff
@@ -0,0 +1,130 @@
...
+namespace {
+
+using StringViewCol = ColumnVector<std::string_view>;
+using UInt64Col = ColumnVector<types::UInt64::Primitive>;
+
+struct Data : public ProcedureData {};
+
+// Read the `currentLabels` list argument (a `ColumnConst<ListView>` produced
+// from a list literal) into a LabelSet of the selected labels.
+LabelSet readSelectedLabels(const Column* rawArg, const LabelMap& labels) {
```
rjb32: Don't return POD data by value

procedures/ListNodesProcedure.cpp
```diff
@@ -0,0 +1,298 @@
...
+    dst.resize(src.size());
+    for (size_t i = 0; i < src.size(); ++i) {
+        dst[i] = static_cast<char>(std::tolower(static_cast<unsigned char>(src[i])));
+    }
+}
+
+// Read a list-literal argument (delivered as a `ColumnConst<ListView>`) into a
+// vector of its string elements; non-string / non-list inputs yield an empty
+// vector, which disables the corresponding filter.
+std::vector<std::string> readStringList(const ColumnConst<ListView>* rawArg,
```
rjb32: Don't return vector by value

procedures/ListNodesProcedure.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+            // Property filters: string properties only, queries lower-cased for
+            // case-insensitive substring matching.
+            std::vector<PropertyFilter> filters;
+            for (size_t i = 0; i < propKeys.size() && i < propValues.size(); ++i) {
+                const auto pType = propTypes.get(propKeys[i]);
+                if (!pType || pType.value()._valueType != ValueType::String) {
+                    continue;
+                }
+                std::string lowered;
```
rjb32: Hoist the string out of the loop

procedures/ProcUtils.cpp
```diff
@@ -0,0 +1,165 @@
...
+                    out += fmt::format("\\u{:04x}", static_cast<unsigned char>(c));
+                } else {
+                    out += c;
+                }
+        }
+    }
+    out += '"';
+}
+
+std::string ProcUtils::encodeProperties(const EntityPropertyView& props,
```
rjb32: Don't return strings or vectors by value

procedures/GetNodeEdgesProcedure.cpp
```diff
@@ -0,0 +1,236 @@
...
+            std::vector<ItemVariant> outerItems;
+
+            // Build the nested-list column value for one direction's edges,
+            // truncating per edge type; fills `counts` with the per-type totals
+            // (counting ALL edges, including those past the limit).
+            const auto buildEdges = [&](auto&& edgeRange, const LimitMap& limits,
+                                        bool isOut, CountMap& counts) -> ListView {
+                counts.clear();
+                outerItems.clear();
+                for (const EdgeRecord& e : edgeRange) {
```
sulaimansuhas: **Deleted (tombstoned) edges leak into results.** `hasTombstones` is set from `tombstones.hasNodes()` (line 128) and only node tombstones are checked (line 201). `buildEdges` iterates `reader.getOutEdges(&singleNodeID)` / `getInEdges(...)`, whose base `GetOutEdgesIterator`/`GetInEdgesIterator` do **not** filter edge tombstones — only the `…ChunkWriter` subclasses call `filterTombstones()`, and `GetOutEdgesRange::begin()` returns the plain base iterator. So a deleted edge is still emitted into `outgoingEdges`/`incomingEdges` and tallied in `outEdgeCounts`/`inEdgeCounts`.

`GetEdgesProcedure` handles this correctly (checks `tombstones.hasEdges()` + `containsEdge`). Here you should also fetch `hasEdges()`/`containsEdge` and skip tombstoned edges inside `buildEdges`.

*Failure:* delete an out-edge of node 0 and submit; `CALL db.getNodeEdges([0], …)` still reports the deleted edge, disagreeing with `CALL db.getEdges([...])`. The PR adds a deleted-*nodes* test but no deleted-*edges* test, which is why this is hidden.

test/query/queries/ShowProceduresTest.cpp
```diff
@@ -43,7 +43,7 @@ TEST_F(ShowProceduresTest, showProcedures) {
     const auto res = query("SHOW PROCEDURES", "default", [&](const Dataframe* df) -> void {
         ASSERT_TRUE(df != nullptr);
         ASSERT_EQ(df->cols().size(), 2);
-        ASSERT_EQ(df->getLogicalRowCount(), 8);
+        ASSERT_EQ(df->getLogicalRowCount(), 9);
```
sulaimansuhas: **This assertion is stale and will fail.** `ProcedureManager::init` now registers 12 procedures in namespace `db` (the 7 pre-existing + `hierarchicalLabelCounts` + `listNodes` + `getEdges` + `getNodes` + `getNodeEdges`), and `ShowProceduresProcessor` enumerates all of them plus `greeter.hello` → **13 rows**, not 9. `ASSERT_EQ(df->getLogicalRowCount(), 9)` fails, and `colName->at(8)` is `"db.listNodes"`, not `"greeter.hello"` (lines 64/79). The test was updated for `hierarchicalLabelCounts` but not for the four `get*/listNodes` procedures — add their name/signature rows and bump the count to 13.

storage/reader/GraphReader.cpp
```diff
@@ -69,18 +70,27 @@ const EdgeRecord* GraphReader::getEdge(EdgeID edgeID) const {
...
     size_t count = 0;
 
-    auto it = matchLabelSets(labelset);
-    for (; it.isValid(); it.next()) {
-        for (const auto& part : _view.dataparts()) {
-            const auto& nodes = part->nodes();
-            const LabelSetHandle& current = it.get();
-            if (!nodes.hasLabelSet(current)) {
-                continue;
+    auto filteredNodeCount = [](NodeRange range, const Tombstones* tombstones) {
```
sulaimansuhas, line 78: **Performance regression: the O(1) fast path is gone.** The previous implementation summed `part->nodes().getRange(current)._count` per matching labelset. `filteredNodeCount` now iterates **every** node in each matching range calling `tombstones->containsNode(id)` with no `hasNodes()` short-circuit, turning an O(matching-labelsets) counter into an O(matching-nodes) scan even when the graph has zero deletions.

This is not just the new `hierarchicalLabelCounts` (which calls it once per label → O(L·N)); it's also on the **planning hot path** via `CardinalityEstimation::estimateNodeCount` (and `DBServerProcessor`), so planning any `MATCH (n:Label)` now scans all matching nodes.

Add an early-out: `if (!tombstones->hasNodes()) { return range._count; }`. Also, `nodes.getRange(it.getKey())` (line 93) re-does two map lookups when `it.getValue()` already holds the `NodeRange`, and `&_view.tombstones()` is re-fetched every iteration — hoist it once.

query/parser/CypherLexer.l
```diff
@@ -114,6 +114,8 @@ using Token = db::YCypherParser::token;
 "EMBEDDING" { KEYWORD(EMBEDDING) }
 "FALSE" { KEYWORD(FALSE) }
 "GRAPH" { KEYWORD(GRAPH) }
+"GRAPHS" { KEYWORD(GRAPHS) }
```
sulaimansuhas, line 117: **Backward-incompatible parse regression.** With `%option caseless`, adding `GRAPHS` and `AVAILABLE` as keywords means `graphs`/`available` (any case) now tokenize as keywords everywhere. But neither is added to the `reservedWord` fallback in `CypherParser.y` (the `name` rule is `symbol | reservedWord`), so they can no longer be used as identifiers. Queries like `MATCH (n) WHERE n.available = true` or `RETURN x AS available` — previously valid, since these weren't keywords before this PR — now fail with a syntax error. `available` is an especially common property name. Add `GRAPHS` and `AVAILABLE` to the `reservedWord` rule to preserve their use as identifiers.

procedures/GetNodeEdgesProcedure.cpp
sulaimansuhas: **Non-deterministic JSON key order.** `countsToJson` iterates a `std::unordered_map<EdgeTypeID, size_t>` (`CountMap`, line 36), so key order in the `outEdgeCounts`/`inEdgeCounts` STRING column is implementation-defined. The fixtures hard-code a specific, non-sorted order (`call-get-node-edges.json` expects `{"1":3,"0":1}`), so these tests are green on the fixture-generating stdlib but can fail on a different one — and this repo runs both libstdc++ (Linux) and libc++ (macOS) CI. Serialize from an ordered `std::map` or sort the keys before emitting so the bytes are reproducible.

procedures/GetNodeEdgesProcedure.cpp
```diff
@@ -0,0 +1,236 @@
...
+        case ProcedureState::Step::EXECUTE: {
+            const GraphView& view = *ctxt->getGraphView();
+            const GraphReader reader(view);
+            ListBuffer<>* listBuffer = ctxt->getListBuffer();
+
+            const Tombstones& tombstones = view.tombstones();
+            const bool hasTombstones = tombstones.hasNodes();
+
+            const std::vector<int64_t> nodeIDs = ProcUtils::readIntList(inputNodeIDs);
+            const size_t defaultLimit = static_cast<size_t>(ProcUtils::readInt(inputLimit, 10));
```
sulaimansuhas: **Negative limit wraps to unbounded.** `static_cast<size_t>(ProcUtils::readInt(inputLimit, 10))` isn't guarded against negatives; `static_cast<size_t>(-1)` is `SIZE_MAX`, so `seen > limit` never fires and every edge is emitted — the opposite of a cap, and a potential huge response for a high-degree node. `buildLimitMap` has the same unguarded `static_cast<size_t>(values[i])` (line 47). `ListNodesProcedure` already does the right thing (`arg > 0 ? static_cast<size_t>(arg) : 0`); apply the same clamp here.

procedures/ProcUtils.cpp
```diff
@@ -0,0 +1,165 @@
...
+    out += '}';
+    return out;
+}
+
+std::vector<int64_t> ProcUtils::readIntList(const ColumnConst<ListView>* arg) {
+    std::vector<int64_t> out;
+    if (!arg) {
+        return out;
+    }
+    for (const ListElementView& el : arg->getRaw()) {
```
sulaimansuhas: **Unchecked column-kind assumption.** `readIntList` calls `arg->getRaw()` and every LIST-arg call site does an unchecked `static_cast<const ColumnConst<ListView>*>(getInputColumn(i))` (e.g. `GetEdgesProcedure.cpp:64`, `GetNodeEdgesProcedure.cpp:104-109`, `ListNodesProcedure`'s `readStringList`). If a LIST argument is ever delivered as anything other than a `ColumnConst<ListView>` (e.g. a per-row `ColumnVector<ListView>` from a bound symbol), the cast + `getRaw()` is undefined behavior rather than a graceful empty result. Note the scalar readers `readInt`/`readBool` and `HierarchicalLabelCountsProcedure::readSelectedLabels` all defensively check `getKind()` first — the guard was known to be needed; apply it consistently (ideally check `getKind()` inside `readIntList` and a shared `readStringList`).

procedures/ProcUtils.cpp
```diff
@@ -0,0 +1,165 @@
...
+                for (const float f : *ptr) {
+                    if (!firstElem) {
+                        out += ',';
+                    }
+                    firstElem = false;
+                    out += fmt::format("{}", f);
+                }
+                out += ']';
+            } else {
+                out += fmt::format("{}", *ptr);
```
sulaimansuhas, line 61: **NaN/Inf produce invalid JSON.** `fmt::format("{}", *ptr)` (and the float-span branch at line 42) emit `nan`/`inf`/`-inf` for non-finite `Double`/float values, which are not valid JSON tokens — the visualiser's `JSON.parse` on the properties/embedding STRING column then throws. The data model permits `Double` properties, so a stored NaN/Inf yields e.g. `{"score":nan}`. Emit `null` (or a quoted sentinel) for non-finite values.

procedures/ProcUtils.cpp
```diff
@@ -0,0 +1,165 @@
...
+    }
+    return out;
+}
+
+int64_t ProcUtils::readInt(const Column* arg, int64_t deflt) {
+    if (!arg) {
+        return deflt;
+    }
+    using IntColumn = ColumnVector<types::Int64::Primitive>;
+    if (arg->getKind() == ColumnConst<types::Int64::Primitive>::staticKind()) {
```
sulaimansuhas: **`readInt`/`readBool` silently ignore UInt64 columns.** `readInt` only matches `ColumnConst<Int64>`/`ColumnVector<Int64>` and otherwise returns `deflt`; `readBool` is analogous. But `readIntList` (lines 118-123) handles *both* `Int` and `UInt` element tags, so the handling is asymmetric. A scalar integer argument delivered as UInt64 (e.g. a `UInt64`-typed value threaded from another procedure's `outEdgeCount`) is not recognized and silently falls back to the default limit/skip — a wrong result with no error. Handle the `UInt64` const/vector kinds too, mirroring `readIntList`.

procedures/GetNodesProcedure.cpp
```diff
@@ -0,0 +1,159 @@
...
+            if (propsCol) {
+                propsCol->clear();
+            }
+
+            std::vector<LabelID> labelIDs;
+            std::vector<ListBuffer<4096>::ListItemVariant> items;
+
+            for (const int64_t rawID : nodeIDs) {
+                const NodeID nodeID {static_cast<NodeID::Type>(rawID)};
+                if (hasTombstones && tombstones.containsNode(nodeID)) {
```
sulaimansuhas: **Duplicated tombstone predicate — reuse `GraphReader`.** `GraphReader` already exposes `nodeIsDeleted(NodeID)` / `edgeIsDeleted(EdgeID)` (GraphReader.cpp:293-299). Instead, each of the four fetch procedures re-fetches `view.tombstones()`, caches `hasNodes()/hasEdges()`, and inlines the `containsNode/containsEdge` test (here at lines 90-91 & 116; also `GetNodeEdgesProcedure` 127-128/201, `GetEdgesProcedure` 82-83/109, `ListNodesProcedure` 144-145/239-240). Four copies must stay in sync with `GraphReader`; call `reader.nodeIsDeleted(id)` / `reader.edgeIsDeleted(id)` instead.

procedures/ProcUtils.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include "views/PropertyView.h"
+#include "metadata/PropertyTypeMap.h"
+#include "metadata/PropertyType.h"
+#include "columns/ColumnConst.h"
+#include "columns/ColumnVector.h"
+#include "list/ListView.h"
+#include "list/ListElementView.h"
+#include "list/ListBufferTypeTag.h"
+
+namespace db {
```
sulaimansuhas: CLAUDE.md: *"In `.cpp` files, open with `using namespace db;` after the includes — don't wrap the body in `namespace db { ... }`."* This file wraps the whole body in `namespace db {` (closed at line 165). All five sibling procedure `.cpp` files added in this PR use `using namespace db;`; make this one consistent.

procedures/ProcUtils.h
```diff
@@ -0,0 +1,46 @@
...
+    // Encode an entity's properties as a JSON object of typed values, keyed by
+    // property NAME (properties with no name are skipped). Used for both node and
+    // edge properties so the visualiser sees one consistent shape.
+    static std::string encodeProperties(const EntityPropertyView& props,
+                                        const PropertyTypeMap& propTypes);
+
+    // Read a LIST-literal argument (delivered as a `ColumnConst<ListView>`) into
+    // a vector of its integer elements. Non-integer / non-list inputs yield an
+    // empty vector.
+    static std::vector<int64_t> readIntList(const ColumnConst<ListView>* arg);
```
sulaimansuhas: CLAUDE.md: *"Don't return STL containers or strings; use output references."* `readIntList` returns `std::vector<int64_t>` by value and `encodeProperties` (line 29) returns `std::string` by value; the anonymous-namespace `buildLimitMap`/`countsToJson` (GetNodeEdges) do too. Also `ListNodesProcedure::readStringList` (ListNodesProcedure.cpp:59) is declared to *both* fill an out-param and return it by value — and all three call sites ignore the return (lines 152-154), so every call deep-copies the vector only to discard it. Convert these to the out-reference pattern (`void readIntList(arg, std::vector<int64_t>& out)`), and make `readStringList` return `void`.

test/query/queries/CallProcedureTest.cpp
```diff
@@ -9,87 +9,102 @@
...
-    auto query(std::string_view q, std::string_view graphName, auto callback) {
-        db::QueryCallbacks callbacks;
-        callbacks.setOnOutputData(callback);
-        const db::QueryState state(graphName, &_env->getMem(), &_queryConfig, &callbacks);
-        return _db->query(q, state);
+    // Total number of result rows a query yields across all chunks.
+    size_t rowCount(const std::string& q) {
+        size_t n = 0;
+        const auto res = query(q, [&](const Dataframe* df) {
+            if (df) n += df->getLogicalRowCount();
```
sulaimansuhas: CLAUDE.md: *"All control structures … must have braces, even for single-statement bodies. Never use the brace-less form, including not on the same line as the condition."*
```suggestion
            if (df) {
                n += df->getLogicalRowCount();
            }
```

procedures/ProcUtils.h
```diff
@@ -0,0 +1,46 @@
...
+
+    // Read a scalar integer argument (delivered as a const or single-row vector
+    // column), falling back to `deflt` when absent or of an unexpected shape.
+    static int64_t readInt(const Column* arg, int64_t deflt);
```
sulaimansuhas: CLAUDE.md: *"No abbreviations in identifiers — spell them out … Applies to locals, parameters, members."* The `deflt` parameter of `readInt`/`readBool` (here and lines 43, plus ProcUtils.cpp:131/148) should be `defaultValue`. Same rule is broken in `GetNodeEdgesProcedure.cpp`: `outsLV`/`insLV` (lines 210/212, `LV` = `ListView`) and `limitIt` (line 166, `It` = `iterator`; also should be `const auto`).

procedures/HierarchicalLabelCountsProcedure.cpp
```diff
@@ -0,0 +1,130 @@
...
+    proc->addArgument("currentLabels", ProcedureType::LIST);
+    proc->addReturnValue("label", ProcedureType::STRING_VIEW);
+    proc->addReturnValue("nodeCount", ProcedureType::UINT_64);
+    ns->addProcedure(proc);
+}
+
+void HierarchicalLabelCountsProcedure::execute(ProcedureState* proc) {
+    const ProcedureContext* ctxt = proc->getContext();
+
+    auto* namesCol = static_cast<StringViewCol*>(proc->data<Data>().getReturnColumn(0));
```
sulaimansuhas: Two CLAUDE.md style issues in `execute`:
- *"Never call the same getter repeatedly when the result can be reused. Capture it once in a named local."* `proc->data<Data>()` is called three times (lines 75, 76, 89). Every other procedure in this PR opens with `Data& data = proc->data<Data>();` — do the same here.
- *"Use `const` extensively — all local variables should be `const` unless they need mutation."* `size_t nodeCount` (line 111) is never mutated (only read at 113/122) → `const size_t nodeCount`.

procedures/GetEdgesProcedure.cpp
sulaimansuhas: We get a list of int64s here as that is how cypher parses numbers.

rjb32 (approved):

PR #718, opened 2026-07-08
MLIR: implement scan_nodes_by_label at the db and nl level

rjb32:
Implements scan_by_label operation

### DIFF
--- a/query/ir/dialect/nl/NLOps.cpp
+++ b/query/ir/dialect/nl/NLOps.cpp
@@ -80,6 +80,17 @@ LogicalResult ScanNodes::inferReturnTypes(MLIRContext* context,
     return success();
 }
 
+// A label scan filters the rows a plain scan would produce, never their shape,
+// so it produces the same single chunk of node IDs per step - the label list is
+// a filter, not a result type.
+LogicalResult ScanNodesByLabel::inferReturnTypes(MLIRContext* context,
+                                                 std::optional<Location> location,
+                                                 ScanNodesByLabel::Adaptor adaptor,
+                                                 SmallVectorImpl<Type>& inferredReturnTypes) {
+    inferredReturnTypes.push_back(IteratorType::get(context, {getNodeIDChunkType(context)}));
+    return success();
+}
+
 // An out-edges fetch produces one row of edge chunks per step, then one
 // filtered chunk per carried column, mirroring db.get_out_edges
 LogicalResult GetOutEdges::inferReturnTypes(MLIRContext* context,
--- a/query/ir/interpreter/NLExecutor.cpp
+++ b/query/ir/interpreter/NLExecutor.cpp
@@ -566,7 +566,10 @@ void NLExecutor::runScanNodesByLabelLoop(NLExecutionContext* context, NLFunction
 
     // The LabelSetHandle borrows the loop data's owned LabelSet, which lives for
     // the whole program, so the handle stays valid for every fill below.
-    ScanNodesByLabelChunkWriter chunkWriter(*context->getView(), LabelSetHandle(loopData->getLabelSet()));
+    const GraphView& view = *context->getView();
+    const LabelSetHandle labelset(loopData->getLabelSet());
+
+    ScanNodesByLabelChunkWriter chunkWriter(view, labelset);
     chunkWriter.setNodeIDs(nodeIDs);
 
     const auto runIteration = [&]() {
--- a/query/ir/interpreter/NLExecutor.cpp
+++ b/query/ir/interpreter/NLExecutor.cpp
@@ -9,6 +9,7 @@
 #include "iterators/GetOutEdgesIterator.h"
 #include "iterators/GetPropertiesWithNullIterator.h"
 #include "iterators/ScanNodesIterator.h"
+#include "iterators/ScanNodesByLabelIterator.h"
 #include "columns/ColumnOptVector.h"
 #include "metadata/PropertyType.h"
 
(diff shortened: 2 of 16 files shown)


query/ir/interpreter/NLExecutor.cpp
```diff
@@ -547,6 +548,48 @@ void NLExecutor::runScanNodesLoop(NLExecutionContext* context, NLFunctionData* d
...
+    const NLStmtContainer* loopBody = loopData->getStmts();
+    ColumnNodeIDs* nodeIDs = loopData->getNodeIDs();
+    const size_t chunkSize = context->getChunkSize();
+
+    // A null limit leaves the loop unbounded, exactly as in runScanNodesLoop.
+    const NLLimitState* limit = loopData->getLimit();
+
+    // The LabelSetHandle borrows the loop data's owned LabelSet, which lives for
+    // the whole program, so the handle stays valid for every fill below.
+    ScanNodesByLabelChunkWriter chunkWriter(*context->getView(), LabelSetHandle(loopData->getLabelSet()));
```
rjb32: Pull the view and the handle outside in local variables

rjb32: Done in ab5869113 — pulled the view and the LabelSetHandle into named locals before constructing the chunk writer.

query/ir/dialect/nl/NLOps.cpp
```diff
@@ -80,6 +80,17 @@ LogicalResult ScanNodes::inferReturnTypes(MLIRContext* context,
     return success();
 }
 
+// A label scan filters the rows a plain scan would produce, never their shape,
+// so it produces the same single chunk of node IDs per step - the label list is
+// a filter, not a result type.
+LogicalResult ScanNodesByLabel::inferReturnTypes(MLIRContext* context,
+                                                 std::optional<Location> location,
+                                                 ScanNodesByLabel::Adaptor adaptor,
+                                                 SmallVectorImpl<Type>& inferredReturnTypes) {
```
cyrusknopf, line 89: What's the difference between using `SmallVector` and `SmallVectorImpl`?

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -547,6 +548,51 @@ void NLExecutor::runScanNodesLoop(NLExecutionContext* context, NLFunctionData* d
...
 }
 
+void NLExecutor::runScanNodesByLabelLoop(NLExecutionContext* context, NLFunctionData* data) {
+    NLScanByLabelLoopData* loopData = static_cast<NLScanByLabelLoopData*>(data);
+
+    // A requested label was absent from the schema, so no node carries the full
+    // conjunction: the scan matches nothing and the loop body never runs.
+    if (!loopData->isMatchable()) {
+        return;
+    }
```
cyrusknopf, line 558: If this means that a label was provided that doesn't exist, then I think that is already caught upstream in the analyzer

query/ir/interpreter/NLTranslator.cpp
```diff
@@ -140,6 +140,12 @@ void NLTranslator::translateBlock(mlir::Block& block, NLStmtContainer* body) {
     for (mlir::Operation& operation : block) {
         if (nl::ScanNodes scanNodes = mlir::dyn_cast<nl::ScanNodes>(operation)) {
             _iteratorConfigs[scanNodes.getResult()] = IteratorConfig {IteratorKind::ScanNodes, {}, {}};
+        } else if (nl::ScanNodesByLabel scanNodesByLabel = mlir::dyn_cast<nl::ScanNodesByLabel>(operation)) {
+            IteratorConfig config {IteratorKind::ScanNodesByLabel, {}, {}};
+            for (const mlir::Attribute label : scanNodesByLabel.getLabels()) {
+                config._labels.push_back(mlir::cast<mlir::StringAttr>(label).str());
```
cyrusknopf: What does the `.str()` method do?

query/ir/interpreter/NLTranslator.h
```diff
@@ -40,6 +41,10 @@ class NLTranslator {
 
         // The accumulator a Sort iterator drains; null for the other kinds.
         NLSortState* _sortState {nullptr};
+
+        // The label names a ScanNodesByLabel iterator filters by; empty for the
+        // other kinds. Resolved to a LabelSet when the loop is translated.
+        std::vector<std::string> _labels;
```
cyrusknopf: Can we change this to be string views over the query string, once the entire CYPHER -> MLIR is wired up?

query/ir/dialect/nl/NLOps.cpp
rjb32: SmallVectorImpl is the base of SmallVector where the size is typed-erased instead of being a template parameter. This is the interface mandated by InferTypeOpInterface it seems.

query/ir/interpreter/NLExecutor.cpp
rjb32: Yes, but this is good to have at as well at the execution level as a guard

query/ir/interpreter/NLTranslator.cpp
rjb32: it converts the specific type that MLIR uses for string attributes into an std::string

query/ir/interpreter/NLTranslator.cpp
```diff
@@ -252,6 +260,40 @@ void NLTranslator::translateScanLoop(mlir::Block& loopBody, NLLimitState* limit,
     translateBlock(loopBody, loopData->getStmts());
 }
 
+void NLTranslator::translateScanByLabelLoop(const IteratorConfig& config,
+                                            mlir::Block& loopBody,
+                                            NLLimitState* limit,
+                                            NLStmtContainer* body) {
+    // A label scan binds the same single node ID chunk as a plain scan.
+    ColumnNodeIDs* nodeIDs = static_cast<ColumnNodeIDs*>(allocColumn(loopBody.getArgument(0)));
```
rjb32: Extract loopBody.getArgument into a local variable before

query/ir/interpreter/NLTranslator.h
rjb32: Probably not if MLIR APIs do not let us, because the label names are extracted from the way the label strings are represented as operation attributes by MLIR.

test/query/ir/DBLoweringTest.cpp
```diff
@@ -1134,6 +1201,24 @@ class DBLoweringTest : public TuringTest {
         }
     }
 
+    // Runs a single-column node scan program and returns its node IDs sorted.
+    // The storage assigns node IDs grouped by label set, not by insertion order,
+    // so the label-scan tests assert the semantics (superset match, conjunction =
+    // intersection) against the plain scan rather than hardcoding an ID scheme.
+    std::vector<uint64_t> collectScan(const char* program, const GraphView& view) {
```
rjb32: Do not return vector as value

query/ir/interpreter/NLTranslator.cpp
cyrusknopf: Then this line should use emplace_back

rjb32: Thanks, done

query/ir/interpreter/NLTranslator.h
rjb32: I actually use llvm StringRef now, which is a view instead of allocating new strings

cyrusknopf: Perfect

PR #720, opened 2026-07-09
MLIR: implement get out/in edges by type

rjb32:
Add db/nl get_out_edges_by_type and get_in_edges_by_type with chunk writers that filter edges by type on the spot, without materializing an unfiltered chunk first.


cyrusknopf:
Could we add some unit tests for the new iterator/chunk writer?

rjb32:
> Could we add some unit tests for the new iterator/chunk writer?

Yes done

cyrusknopf (approved):

PR #721, opened 2026-07-09
MLIR: Implement constants

cyrusknopf:
Constant operations to produce a `ColumnConst` with a specific value. Lowered from DB, to NL, to execution. Tests for hardcoded MLIR programs.

### DIFF
--- a/query/ir/interpreter/NLTranslator.cpp
+++ b/query/ir/interpreter/NLTranslator.cpp
@@ -41,6 +41,17 @@ llvm::StringRef edgeTypeName(mlir::Value handle) {
     return handleOp.getName();
 }
 
+static bool isDefinedInOuterScope(mlir::Operation* definingOp, mlir::Block* outputBlock) {
+    mlir::Block* current = outputBlock;
+    while (mlir::Operation* parentOp = current->getParentOp()) {
+        current = parentOp->getBlock();
+        if (current == definingOp->getBlock()) {
+            return true;
+        }
+    }
+    return false;
+}
+
 // The with-null fetch handler for a property's value type, on the node side
 // when isNode is true and the edge side otherwise. Selecting it here keeps the
 // value-type dispatch with the rest of translation; the handler bodies live in
@@ -511,9 +522,11 @@ void NLTranslator::translateOutput(nl::Output output, NLStmtContainer* body) {
         // produced in this same loop body; it is equally available to output.
         mlir::Operation* definingOp = column.getDefiningOp();
         const bool isProducedInThisBlock = definingOp && definingOp->getBlock() == outputBlock;
+        const bool isFromOuterScope = definingOp && isDefinedInOuterScope(definingOp, outputBlock);
 
-        if (!isInnermostLoopVariable && !isProducedInThisBlock) {
-            throw IRException("nl.output columns must belong to the innermost enclosing nl.for body");
+        if (!isInnermostLoopVariable && !isProducedInThisBlock && !isFromOuterScope) {
+            throw IRException("nl.output columns must belong to the innermost enclosing "
+                              "nl.for body or an outer scope");
         }
 
         outputData->addOutputColumn(getColumn(column));
--- a/query/ir/interpreter/NLTranslator.cpp
+++ b/query/ir/interpreter/NLTranslator.cpp
@@ -5,6 +5,7 @@
 #include <spdlog/fmt/bundled/format.h>
 
 #include "mlir/IR/Block.h"
+#include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Verifier.h"
 
@@ -83,6 +84,21 @@ ValueType valueTypeFromElementType(mlir::Type elementType) {
     }
 
     throw IRException("Unsupported nullable value chunk element type");
+}
+
+template <SupportedType T>
+typename T::Primitive constantValueAs(mlir::TypedAttr value) {
+    if constexpr (std::same_as<T, types::Int64>) {
+        return mlir::cast<mlir::IntegerAttr>(value).getInt();
+    } else if constexpr (std::same_as<T, types::UInt64>) {
+        return mlir::cast<mlir::IntegerAttr>(value).getValue().getZExtValue();
+    } else if constexpr (std::same_as<T, types::Double>) {
+        return mlir::cast<mlir::FloatAttr>(value).getValueAsDouble();
+    } else if constexpr (std::same_as<T, types::Bool>) {
+        return CustomBool(mlir::cast<mlir::IntegerAttr>(value).getInt() != 0);
+    } else { // string and embedding not yet supported
+        throw IRException("Unsupported constant value type");
+    }
 }
 
 // The runtime reduction the interpreter dispatches on, from the MLIR one the op
@@ -187,6 +203,10 @@ void NLTranslator::translateBlock(mlir::Block& block, NLStmtContainer* body) {
             translateFor(forLoop, body);
         } else if (mlir::isa<nl::GetPropertyType, nl::GetEdgeType>(operation)) {
             // The handle carries only a name; a fetch/hop resolves it on consumption
+        } else if (mlir::isa<nl::GetPropertyType>(operation)) {
+            // The handle carries only a name; a fetch resolves it on consumption
+        } else if (nl::Constant constant = mlir::dyn_cast<nl::Constant>(operation)) {
+            translateConstant(constant);
         } else if (nl::GetNodeProperties getNodeProperties = mlir::dyn_cast<nl::GetNodeProperties>(operation)) {
             translatePropertyFetch(getNodeProperties.getInputNodes(),
                                    getNodeProperties.getPropertyType(),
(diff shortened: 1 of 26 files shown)


test/query/ir/DBLoweringTest.cpp
```diff
@@ -316,6 +317,112 @@ class CollectingEdgeIntPropSink : public NLOutputSink {
...
 
+// Collects constant result rows: each output column is a ColumnConst<T> - the
+// single broadcast value a db.constant lowers to - so a constant program emits
+// exactly one row, one value per column. Every column must be a ColumnConst<T>.
+template <typename T>
+class CollectingConstSink : public NLOutputSink {
+public:
+    void appendChunks(std::span<const Column* const> chunks, size_t offset, size_t rowCount) override {
+        for (size_t rowIndex = offset; rowIndex < offset + rowCount; rowIndex++) {
+            std::vector<T> row;
```
rjb32, line 328: Please hoist outside of the loop

rjb32:
So you use a chunk type at the nl level such as nl.chunk<i64> for constants and dynamically use ColumnConst or ColumnVector. Can you elaborate on how comfortable are we to not make ColumnVector or ColumnConst reflected in the nl type system? Can it fully be treated as a matter of representation optimisation?

samples/mlir/constant_types.mlir
```diff
@@ -0,0 +1,10 @@
+module {
+  func.func @main() {
+    %i = db.constant(30 : i64)
+    %u = db.constant(7 : ui64)
+    %f = db.constant(2.5 : f64)
+    %b = db.constant(true)
+    db.output(%i, %u, %f, %b) : !db.column<i64>, !db.column<ui64>, !db.column<f64>, !db.column<i1>
```
rjb32, line 7: Please add a sample with an output being fed constants and chunks coming from get_node_properties

cyrusknopf:
> So you use a chunk type at the nl level such as nl.chunk for constants and dynamically use ColumnConst or ColumnVector. Can you elaborate on how comfortable are we to not make ColumnVector or ColumnConst reflected in the nl type system? Can it fully be treated as a matter of representation optimisation?

Yes it is as you describe. I view `ColumnConst` as a storage optimisation. I initially used `ColumnVector` for constants until I realise the column operations require `ColumnConst`s to behave correctly for operations with constants. Unless we can think of a case where a `ColumnConst` is not purely a storage optimisation, I think this is the best approach - removing complexity from `nl`.

An `nl.chunk` is still an abstraction over "chunked data", the choice of how we represent that in memory (`Vector` vs `Const`) should be independent of the `nl` dialect IMO.

rjb32:
> > So you use a chunk type at the nl level such as nl.chunk for constants and dynamically use ColumnConst or ColumnVector. Can you elaborate on how comfortable are we to not make ColumnVector or ColumnConst reflected in the nl type system? Can it fully be treated as a matter of representation optimisation?
> 
> Yes it is as you describe. I view `ColumnConst` as a storage optimisation. I initially used `ColumnVector` for constants until I realise the column operations require `ColumnConst`s to behave correctly for operations with constants. Unless we can think of a case where a `ColumnConst` is not purely a storage optimisation, I think this is the best approach - removing complexity from `nl`.
> 
> An `nl.chunk` is still an abstraction over "chunked data", the choice of how we represent that in memory (`Vector` vs `Const`) should be independent of the `nl` dialect IMO.

Yes I agree this is the most elegant solution from the nl perspective. We can just see it as just a chunk of T in the abstract sense.

rjb32:
Please add a sample with db.constant, db.scan_nodes() and db.output of both the constant and the chunk

query/ir/interpreter/NLTranslator.cpp
```diff
@@ -388,6 +407,23 @@ void NLTranslator::translatePropertyFetch(mlir::Value inputValue,
...
+    Column* column = nullptr;
+    const auto materialize = [&]<SupportedType T>() {
+        auto* typed = _memory->alloc<ColumnConst<typename T::Primitive>>();
+        typed->set(constantValueAs<T>(value));
+        column = typed;
+    };
+    ValueTypeDispatcher(valueType).execute(materialize);
+    bioassert(column, "Failed to allocate column.");
+
+    _valueSlots[constant.getResult()] = column;
```
rjb32: constant.getResult() has already been called line 412, please extract into a local variable

cyrusknopf: Fixed

cyrusknopf:
@rjb32 Fixed the issue with `MATCH (n) RETURN 5` by adding a `cardinality` argument to `nl.output` which determines the cardinality of the output (in lieu of a skip or limit)

rjb32:
> @rjb32 Fixed the issue with `MATCH (n) RETURN 5` by adding a `cardinality` argument to `nl.output` which determines the cardinality of the output (in lieu of a skip or limit)

Please add sample db and nl programs showing how the cardinality is used and manifests itself, for example in a program that implements `MATCH (n) RETURN 5`

cyrusknopf:
> > @rjb32 Fixed the issue with `MATCH (n) RETURN 5` by adding a `cardinality` argument to `nl.output` which determines the cardinality of the output (in lieu of a skip or limit)
> 
> Please add sample db and nl programs showing how the cardinality is used and manifests itself, for example in a program that implements `MATCH (n) RETURN 5`

NLExecutorTest.cpp:
~~~mlir
constexpr const char* constantOverMatchProgram = R"mlir(
func.func @main() {
  %c = nl.constant(5 : i64)
  %nodes = nl.scan_nodes()
  nl.for %a in %nodes : !nl.iter<!nl.chunk<!storage.node_id>> {
    nl.output(%c) cardinality %a : !nl.chunk<i64>
  }
  func.return
}
)mlir";
~~~

DBLoweringTest.cpp:
~~~mlir
// MATCH (a), (b) RETURN 5
constexpr const char* constantOverCrossProductProgram = R"mlir(
func.func @main() {
  %0:2 = db.cross_product factor {
    %a = db.scan_nodes() : !db.column<!storage.node_id>
    db.yield %a : !db.column<!storage.node_id>
  } factor {
    %b = db.scan_nodes() : !db.column<!storage.node_id>
    db.yield %b : !db.column<!storage.node_id>
  }
  %c = db.constant(5 : i64)
  db.output(%c) : !db.column<i64>
  return
}
)mlir";
~~~

I've added them both into `samples/mlir` as well

samples/mlir/xprod_return_constant.mlir
```diff
@@ -0,0 +1,12 @@
+func.func @main() {
+  %0:2 = db.cross_product factor {
+    %a = db.scan_nodes() : !db.column<!storage.node_id>
+    db.yield %a : !db.column<!storage.node_id>
+  } factor {
+    %b = db.scan_nodes() : !db.column<!storage.node_id>
+    db.yield %b : !db.column<!storage.node_id>
+  }
+  %c = db.constant(5 : i64)
+  db.output(%c) : !db.column<i64>
```
rjb32, line 10: Why is there no cardinality given here?

samples/mlir/scan_return_constant.nl.mlir
```diff
@@ -0,0 +1,8 @@
+func.func @main() {
+  %c = nl.constant(5 : i64)
+  %nodes = nl.scan_nodes()
+  nl.for %a in %nodes : !nl.iter<!nl.chunk<!storage.node_id>> {
+    nl.output(%c) cardinality %a : !nl.chunk<i64>
```
rjb32, line 5: What happens for nested loops? Is it product cardinality? The innermost?

samples/mlir/xprod_return_constant.mlir
cyrusknopf: The cardinality argument is only in the `nl` dialect, not `db`, because it depends on the loops which otherwise bound the output

rjb32: Can you add the nl dialect form of this program then please?

samples/mlir/scan_return_constant.nl.mlir
cyrusknopf: The cardinality is determined by `_innermostCardinality` which is updated when translating loops and cross products from `db` to `nl` in `DBLowering`

samples/mlir/xprod_return_constant.mlir
cyrusknopf:
~~~mlir
func.func @main() {
  %0 = nl.constant(5 : i64)
  %1 = nl.scan_nodes()
  nl.for %arg0 in %1 : !nl.iter<!nl.chunk<!storage.node_id>> {
    %2 = nl.scan_nodes()
    nl.for %arg1 in %2 : !nl.iter<!nl.chunk<!storage.node_id>> {
      %3:2 = nl.cross_product{%arg0} {%arg1} : {!nl.chunk<!storage.node_id>} {!nl.chunk<!storage.node_id>}
      nl.output(%0) cardinality %3#0 : !nl.chunk<i64>
    }
  }
  return
}
~~~

added under `samples/mlir/xprod_return_constant.nl.mlir`

rjb32 (approved):

PR #723, opened 2026-07-10
MLIR: Implement EQ

cyrusknopf:


query/ir/lowering/DBLowering.cpp
```diff
@@ -992,6 +1053,79 @@ void DBLowering::foldSkipTruncatesIntoOutputs(mlir::func::FuncOp nlFunction) {
...
+        return firstBlock;
+    }
+
+    if (firstBlock == _entryBlock) {
+        return secondBlock;
+    } else if (secondBlock == _entryBlock) {
+        return firstBlock;
+    }
+
+    throw IRException("row-wise operands must be bound in the same loop");
```
rjb32: This is the right version of deeperBlock that should be used in the earlier add PR

rjb32 (approved):

PR #725, opened 2026-07-10
MLIR: implement grouped aggregation

rjb32:
Add a db.group_aggregate operator that reduces rows to one row per grouping-key tuple


query/ir/dialect/db/DBOps.td
```diff
@@ -681,4 +681,66 @@ def AddOp : BinaryOp<"add", [Pure]> {
...
+    aggregate inputs, one per aggregate). `aggregates` names the reduction of each
+    aggregate input, as GroupAggregateKind integers (0 = count, 1 = sum, 2 = min,
+    3 = max, 4 = avg), so its length is the number of aggregate inputs.
+
+      %a     = db.scan_nodes() : !db.column<!storage.node_id>
+      %city  = db.get_node_properties(%a, "city")  : (!db.column<!storage.node_id>) -> !db.column<none>
+      %score = db.get_node_properties(%a, "score") : (!db.column<!storage.node_id>) -> !db.column<none>
+      %gcity, %total = db.group_aggregate(%city, %score) keys 1 aggregates [1]
+                         : (!db.column<none>, !db.column<none>) -> (!db.column<none>, !db.column<none>)
+      db.output(%gcity, %total) : !db.column<none>, !db.column<none>
```
cyrusknopf: We are using a dense integer array to represent the aggregate functions which are present. Can we use an enum instead? or a bitmask? the first is more readable (`[1, 0, 3]` denoting sum, count and min is not obvious), whilst the second is more efficient (one bit per aggregate, however this would not preserve order, if that is important)

rjb32: I agree this is far from ideal to just use a list of enum numbers in aggregates to specify the aggregate functions

query/ir/lowering/DBLowering.cpp
```diff
@@ -982,6 +1014,98 @@ void DBLowering::lowerAggregate(mlir::Value input, mlir::Value result, storage::
...
+    for (size_t aggregateIndex = 0; aggregateIndex < kinds.size(); aggregateIndex++) {
+        const auto kind = static_cast<storage::GroupAggregateKind>(kinds[aggregateIndex]);
+        const mlir::Value inputChunk = chunks[keyCount + aggregateIndex];
+
+        if (kind == storage::GroupAggregateKind::Count) {
+            // count tallies rows regardless of value, so its input may be an ID
+            // column (count(*)) or a property value column (count(x)); either way the
+            // result is a single non-null unsigned i64 per group.
+            chunkTypes.push_back(countChunkType);
+        } else {
```
cyrusknopf: There are a few places where its `if (count) else ...` where I think it would be better to use a switch with all the current `else` cases falling into the same case, because otherwise adding an extra aggregate type here will silently fall into the else branch, possibly where it shouldn't

query/ir/dialect/db/DBOps.td
rjb32: Done, see samples

query/ir/lowering/DBLowering.cpp
rjb32: Done

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -232,6 +234,23 @@ void distinctAppendValueBytes(std::string& key, const Primitive& value) {
     key.append(reinterpret_cast<const char*>(&value), sizeof(Primitive));
 }
 
+// A double is normalized before its bytes are appended so keys compare by Cypher
+// value equality, not bit pattern: -0.0 collapses to +0.0 (0.0 == -0.0 in Cypher)
+// and every NaN payload maps to one canonical NaN, so two distinct bit patterns of
+// the same value never split into separate groups (or survive DISTINCT as two rows).
+void distinctAppendValueBytes(std::string& key, double value) {
```
rjb32, line 343: @cyrusknopf That's the bit I was mentioning about -0.0 == +0.0 in cypher

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -490,6 +509,254 @@ NLAggregateUpdateFunction selectMinMaxUpdate(ValueType inputType) {
...
+            std::optional<double>& running = raw[group];
+            running = running.value() + static_cast<double>(*value);
+            counts[group]++;
+        }
+    }
+}
+
+// Tally every row into its group (count(*) over a never-null column): each row
+// charges its group regardless of value, so the input values are never read.
+void groupFoldCountAll(Column* accumulator, std::vector<uint64_t>& counts, const Column* input, const std::vector<size_t>& groups) {
```
rjb32: Put each argument on its own line one below each other

query/ir/interpreter/NLProgram.h
```diff
@@ -1069,6 +1070,193 @@ class NLAggregateResultData : public NLFunctionData {
...
+
+    void addKeyColumn(const KeyColumn& key) { _keyColumns.push_back(key); }
+    void addAggregate(const Aggregate& aggregate) { _aggregates.push_back(aggregate); }
+
+    // Mutable so the emit-loop translation can fill in each column's _output after
+    // the update translation set up the rest.
+    std::vector<KeyColumn>& keyColumns() { return _keyColumns; }
+    std::vector<Aggregate>& aggregates() { return _aggregates; }
+
+    std::unordered_map<std::string, size_t>& groups() { return _groups; }
```
rjb32: Do not return an std::unordered_map by reference directly. Let's make it its own class to reify the data structure

query/ir/interpreter/NLExecutor.cpp
cyrusknopf: Nice, the code below looked a bit suspect to me, but I checked in godbolt (https://godbolt.org/z/d6YG5xY7W) and it should be correct. Might be worth adding a comment though because it looks like a noop

PR #730, opened 2026-07-13
MLIR: Implement Filters (Dialect and frontend)

cyrusknopf:
`DBProgramGenerator` now generates the new `db.filter` operation for `WHERE` expressions

### DIFF
--- a/query/ir/codegen/DBProgramGenerator.cpp
+++ b/query/ir/codegen/DBProgramGenerator.cpp
@@ -5,6 +5,7 @@
 #include <string_view>
 #include <type_traits>
 
+#include "expr/Operators.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/Block.h"
 #include "mlir/IR/Builders.h"
@@ -34,8 +35,10 @@
 #include "SinglePartQuery.h"
 #include "WhereClause.h"
 #include "decl/VarDecl.h"
+#include "Literal.h"
 #include "expr/BinaryExpr.h"
 #include "expr/Expr.h"
+#include "expr/LiteralExpr.h"
 #include "expr/PropertyExpr.h"
 #include "expr/UnaryExpr.h"
 #include "stmt/MatchStmt.h"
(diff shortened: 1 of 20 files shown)


query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -524,31 +536,315 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
         finalIdentities[varName] = finalValue;
     }
 
-    const auto lookup = [&](auto&& item) -> mlir::Value {
+    const auto fetch = [&](auto&& item) -> mlir::Value {
```
rjb32: We should have a more explicit name than fetch

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -524,31 +536,315 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
...
+        const Pattern* pattern = matchStmt->getPattern();
+        const WhereClause* where = pattern->getWhere();
+        if (!where) {
+            continue;
+        }
+
+        const Expr* predicateExpr = where->getExpr();
+        translateExpr(predicateExpr);
+
+        if (!_exprMap.contains(predicateExpr)) {
```
rjb32: What does it mean to do a translate which puts into the _exprMap and then do a contains check afterwards?

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -524,31 +536,315 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
...
+            continue;
+        }
+
+        const mlir::Value predicate = _exprMap.at(predicateExpr);
+        const mlir::Location loc = _opBuilder.getUnknownLoc();
+
+        // Filter all columns that are defined here
+        llvm::SmallVector<mlir::Value> columnsToFilter;
```
rjb32: Hoist the small vectors outside of the loop

query/ir/interpreter/NLTranslator.cpp
```diff
@@ -602,6 +630,54 @@ void NLTranslator::translateOr(nl::Or orOp, NLStmtContainer* body) {
...
+}
+
+void NLTranslator::translateFilter(nl::Filter filter, NLStmtContainer* body) {
+    const Column* mask = getColumn(filter.getMask());
+
+    const auto maskChunk = mlir::cast<nl::ChunkType>(filter.getMask().getType());
+    const bool maskNullable = mlir::isa<storage::NullableType>(maskChunk.getElementType());
+    const NLMaskSurvivorFunction filterFunction =
+        NLExecutor::selectMaskSurvivorFunction(maskNullable);
+
```
rjb32, line 653: Too many blank lines

query/ir/codegen/DBProgramGenerator.cpp
cyrusknopf: Done

query/ir/codegen/DBProgramGenerator.cpp
cyrusknopf: Good point, changed to a single lookup and an assertion

query/ir/codegen/DBProgramGenerator.cpp
cyrusknopf: Changed to `getVarForItem`

query/ir/interpreter/NLTranslator.cpp
cyrusknopf: Fixed thanks

rjb32 (approved):

PR #732, opened 2026-07-14
Cap data parts per graph, rejecting COMMIT/CHANGE SUBMIT past the limit

rjb32:
Rejects COMMIT and CHANGE SUBMIT once a graph's main-head plus in-progress-change data parts reach a configurable limit (TURING_MAX_DATAPARTS, default 32)


query/pipeline/processors/DataPartLimit.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "PipelineException.h"
+#include "versioning/Transaction.h"
+#include "versioning/ChangeAccessor.h"
+#include "versioning/CommitBuilder.h"
+
+#include "BioAssert.h"
+
+using namespace db;
+
+void db::throwIfTooManyDataParts(const ChangeAccessor& access, const TuringConfig* config) {
```
cyrusknopf:
```suggestion
void throwIfTooManyDataParts(const ChangeAccessor& access, const TuringConfig* config) {
```

sulaimansuhas:
I'm not sure how this will work given the current state of the merge dataparts? If a branch has too many commits, there is no way for the user to merge all the commits into 1? So if we do this I think we basically break the whole version control write system. It's also tricky to do this for a branch  because then we lose the history - unless we are rebasing merge commit + branch commit history into main. There is also the issue of the memory usage just increasing without a restart.

PR #737, opened 2026-07-15
MLIR: Translate JOINS

cyrusknopf:


query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -291,6 +294,56 @@ void DBProgramGenerator::generateTraversal(const CypherAST* ast) {
...
+    const VariableDependency* fstMergeSource = nullptr;
+    const VariableDependency* sndMergeSource = nullptr;
+    for (const DependencyEdge* inEdge : mergeVar->incoming()) {
+        if (!inEdge->isMetaEdge()) {
+            continue;
+        }
+        if (!fstMergeSource) {
+            fstMergeSource = inEdge->src();
+        } else {
+            sndMergeSource = inEdge->src();
```
rjb32, line 308: Shouldn't we break when we found both?

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -569,6 +626,110 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
...
+            continue;
+        }
+
+        const MatchStmt* matchStmt = static_cast<const MatchStmt*>(stmt);
+        const Pattern* pattern = matchStmt->getPattern();
+
+        // Collect all property constraint expressions from every entity pattern
+        // in this MATCH. Each EntityPropertyConstraint._expr is a synthesized
+        // equality BinaryExpr (e.g. n.name = "Alice") produced by the analyzer.
+        std::vector<const Expr*> constraintExprs;
```
rjb32: Hoist vector out of loop

query/ir/frontend/cypher/VariableDependencyGraph.cpp
```diff
@@ -364,11 +364,29 @@ void VariableDependencyGraph::getNextAnonymisation(VariableDependency* v, std::s
 }
 
 void VariableDependencyGraph::canonicaliseCycle(Cycle& cyc) {
+    const auto isEdgeVariable = [](const VariableDependency* v) {
```
rjb32, line 367: Didn't we see something similar to that lambda already?

test/query/ir/EquivalenceTest.cpp
```diff
@@ -320,6 +320,77 @@ TEST_F(EquivalenceTest, eqFilters) {
...
+    // expectEquivalent("MATCH (a {isFrench: true})-->(b {name: 'Bio'}) RETURN a");
+}
+
+// FIXME: Enable commented tests below once string literals are supported
+TEST_F(EquivalenceTest, inlineEdgePropertyConstraints) {
+    // Integer edge property constraint.
+    expectEquivalent("MATCH (a)-[e {duration: 20}]->(b) RETURN a, b");
+
+    // String edge property constraint — only expert edges.
+    // expectEquivalent("MATCH (a)-[e {proficiency: 'expert'}]->(b) RETURN a, b");
```
rjb32: Why these are uncommented?

rjb32:
Claude code review to be assessed:

1. addMergeFilter builds its FilterOp over the global _varMap, not the component-local set — DBProgramGenerator.cpp:324

The merge FilterOp is emitted into the current component's isolated scratch region, but for (auto& [mapVar, values] : _varMap) pulls in columns from previously-translated connected components that live in separate mlir::Regions.

Failure: `MATCH (x), (a)-[e1]->(b)-[e2]->(d), (a)-[e3]->(c)-[e4]->(d) RETURN x` — x's component is registered in _varMap before the diamond. When the diamond's addMergeFilter runs, it passes x's SSA value (defined in another region) as a FilterOp operand → malformed MLIR (operand doesn't dominate its use / cross-sibling-region reference), so verification/lowering fails and the query errors. addEdgeTraversal gets this right — it carries only the component-local carriedSet. That's the fix pattern to mirror.

2. Cascaded multi-cycle merge trips the "No edge producer" assertion — DBProgramGenerator.cpp:409 · CONFIRMED

When a node is the convergence point of 3+ merge paths, VariableDependencyGraph::cascadeMerges inserts an intermediate parent merge var. The parent is reached via a meta edge (processed with haveTriple=false), and at line 403 it pushes the final target v with predPred = the parent's meta pred.

Failure: When v is popped, haveTriple = pred && predPred is now true, so the if (!haveTriple) block (408–411) is skipped — addMergeFilter(v) never runs. Execution falls through to line 420 where both producesEdgeVar(pred) and producesNodeVar(pred) are false for a MERGE edge, and bioassert(producesEdgeVar(...), "No edge producer") at line 422 throws. A valid multi-cycle query fails with an internal error, and no merge filter is applied.

A single diamond gives the sink exactly 2 merge edges (canonicaliseCycle picks the sink as pivot, subdivideWithMerge runs twice into it), so it stays under the threshold. You need a third converging path. The minimal query is a "theta" pattern — three internally-disjoint paths sharing the same two endpoints:

```
MATCH (a)-->(b)-->(d),
      (a)-->(c)-->(d),
      (a)-->(e)-->(d)
RETURN a, d
```

Why it fires:
- d has in-degree 3, so it's chosen as the cycle pivot for both basis cycles → it ends up with 3 incoming MERGE edges.
- cascadeMerges sees count_if(..., isMeta) > 2, folds two of them into a new parent node: src1 → parent, src2 → parent, and parent → d (all MERGE edges). Now d has 2 meta edges (one original mid, one from parent), and parent has 2 incoming meta edges.
- In translateComponent's DFS, parent is popped via a meta edge with predPred == null → haveTriple is false, so it takes the else branch at line 419 and pushes d as (d, parent→d, pred=meta-into-parent).
- When d is popped, pred (parent→d) and predPred (meta-into-parent) are both non-null → haveTriple == true, so the if (!haveTriple) block at 408–411 is skipped and addMergeFilter(d) never runs.
- Execution falls through to line 420. Both pred and predPred are MERGE edges, so producesEdgeVar is false for both → bioassert(producesEdgeVar(edgeVarProd), "No edge producer") at line 422 throws.

So that valid query dies with an internal "No edge producer" error, and no merge filter is applied.

4. Redundant second filter pass — DBProgramGenerator.cpp:200 

generatePropertyConstraints emits a full db::FilterOp over every live column, then generateFilters emits a second db::FilterOp over the same columns for WHERE. For a query with both inline constraints and WHERE (e.g. MATCH (a{age:32})-->(b) WHERE a.hasPhD), two back-to-back FilterOps materialize and filter every column at runtime. Folding the synthesized constraints into the same AND predicate as the WHERE clause would filter once.

5. Filter-emission block copy-pasted three times — DBProgramGenerator.cpp:707

The "iterate _varMap → collect columns → build resultTypes → create db::FilterOp → push results back" block appears in generateFilters (779–802), generatePropertyConstraints (707–729), and addMergeFilter (322–343) — and they've already diverged (two use _varMap.at(var).push_back, addMergeFilter uses registerValue). A shared emitFilterOverLiveColumns(...) helper collapses all three. (Note: this is the same call site as #1 — factoring it through the component-local set fixes the divergence and the bug.)

7. isEdgeVariable lambda re-encodes producesEdgeVar — VariableDependencyGraph.cpp:367 

The "edge produced by GET_OUT_EDGES/GET_IN_EDGES" classification now exists in two TUs (here and DBProgramGenerator.cpp:59). Add a new edge-producing EdgeType and one site can be missed, making canonicaliseCycle's pivot ranking disagree with DBProgramGenerator's triple orientation. A shared predicate on EdgeMetadata/DependencyEdge keeps them in sync.

query/ir/codegen/DBProgramGenerator.cpp
cyrusknopf: Good point thanks

query/ir/codegen/DBProgramGenerator.cpp
cyrusknopf: Fixed

query/ir/frontend/cypher/VariableDependencyGraph.cpp
cyrusknopf: I've added the methods to DependencyEdge, and use that logic in VariableDependencyGraph and DBProgramGenerator

test/query/ir/EquivalenceTest.cpp
cyrusknopf: We didn't have string literal support when I wrote them, I've uncommented them now we do

cyrusknopf:
> Claude code review to be assessed:
> 
> 1. addMergeFilter builds its FilterOp over the global _varMap, not the component-local set — DBProgramGenerator.cpp:324
> 
> The merge FilterOp is emitted into the current component's isolated scratch region, but for (auto& [mapVar, values] : _varMap) pulls in columns from previously-translated connected components that live in separate mlir::Regions.
> 
> Failure: `MATCH (x), (a)-[e1]->(b)-[e2]->(d), (a)-[e3]->(c)-[e4]->(d) RETURN x` — x's component is registered in _varMap before the diamond. When the diamond's addMergeFilter runs, it passes x's SSA value (defined in another region) as a FilterOp operand → malformed MLIR (operand doesn't dominate its use / cross-sibling-region reference), so verification/lowering fails and the query errors. addEdgeTraversal gets this right — it carries only the component-local carriedSet. That's the fix pattern to mirror.
> 
> 2. Cascaded multi-cycle merge trips the "No edge producer" assertion — DBProgramGenerator.cpp:409 · CONFIRMED
> 
> When a node is the convergence point of 3+ merge paths, VariableDependencyGraph::cascadeMerges inserts an intermediate parent merge var. The parent is reached via a meta edge (processed with haveTriple=false), and at line 403 it pushes the final target v with predPred = the parent's meta pred.
> 
> Failure: When v is popped, haveTriple = pred && predPred is now true, so the if (!haveTriple) block (408–411) is skipped — addMergeFilter(v) never runs. Execution falls through to line 420 where both producesEdgeVar(pred) and producesNodeVar(pred) are false for a MERGE edge, and bioassert(producesEdgeVar(...), "No edge producer") at line 422 throws. A valid multi-cycle query fails with an internal error, and no merge filter is applied.
> 
> A single diamond gives the sink exactly 2 merge edges (canonicaliseCycle picks the sink as pivot, subdivideWithMerge runs twice into it), so it stays under the threshold. You need a third converging path. The minimal query is a "theta" pattern — three internally-disjoint paths sharing the same two endpoints:
> 
> ```
> MATCH (a)-->(b)-->(d),
>       (a)-->(c)-->(d),
>       (a)-->(e)-->(d)
> RETURN a, d
> ```
> 
> Why it fires:
> 
> * d has in-degree 3, so it's chosen as the cycle pivot for both basis cycles → it ends up with 3 incoming MERGE edges.
> * cascadeMerges sees count_if(..., isMeta) > 2, folds two of them into a new parent node: src1 → parent, src2 → parent, and parent → d (all MERGE edges). Now d has 2 meta edges (one original mid, one from parent), and parent has 2 incoming meta edges.
> * In translateComponent's DFS, parent is popped via a meta edge with predPred == null → haveTriple is false, so it takes the else branch at line 419 and pushes d as (d, parent→d, pred=meta-into-parent).
> * When d is popped, pred (parent→d) and predPred (meta-into-parent) are both non-null → haveTriple == true, so the if (!haveTriple) block at 408–411 is skipped and addMergeFilter(d) never runs.
> * Execution falls through to line 420. Both pred and predPred are MERGE edges, so producesEdgeVar is false for both → bioassert(producesEdgeVar(edgeVarProd), "No edge producer") at line 422 throws.
> 
> So that valid query dies with an internal "No edge producer" error, and no merge filter is applied.
> 
> 4. Redundant second filter pass — DBProgramGenerator.cpp:200
> 
> generatePropertyConstraints emits a full db::FilterOp over every live column, then generateFilters emits a second db::FilterOp over the same columns for WHERE. For a query with both inline constraints and WHERE (e.g. MATCH (a{age:32})-->(b) WHERE a.hasPhD), two back-to-back FilterOps materialize and filter every column at runtime. Folding the synthesized constraints into the same AND predicate as the WHERE clause would filter once.
> 
> 5. Filter-emission block copy-pasted three times — DBProgramGenerator.cpp:707
> 
> The "iterate _varMap → collect columns → build resultTypes → create db::FilterOp → push results back" block appears in generateFilters (779–802), generatePropertyConstraints (707–729), and addMergeFilter (322–343) — and they've already diverged (two use _varMap.at(var).push_back, addMergeFilter uses registerValue). A shared emitFilterOverLiveColumns(...) helper collapses all three. (Note: this is the same call site as #1 — factoring it through the component-local set fixes the divergence and the bug.)
> 
> 7. isEdgeVariable lambda re-encodes producesEdgeVar — VariableDependencyGraph.cpp:367
> 
> The "edge produced by GET_OUT_EDGES/GET_IN_EDGES" classification now exists in two TUs (here and DBProgramGenerator.cpp:59). Add a new edge-producing EdgeType and one site can be missed, making canonicaliseCycle's pivot ranking disagree with DBProgramGenerator's triple orientation. A shared predicate on EdgeMetadata/DependencyEdge keeps them in sync.

1 and 2 fixed with tests; 4 is left as we discussed yesterday we will revisit the filter passes; 7 is fixed, 5. is no longer valid as the code to add a filter has changed

rjb32 (approved):

PR #740, opened 2026-07-15
MLIR: implement const_scan_nodes in db and nl dialects

rjb32:
Add a const_scan_nodes op to the db and nl dialects that injects an explicitly-listed set of node IDs into a variable. Drops the NodeIDs that don't exist in the graph.

### DIFF
--- a/query/ir/dialect/db/DBOps.cpp
+++ b/query/ir/dialect/db/DBOps.cpp
@@ -222,17 +222,6 @@ LogicalResult ScanNodesByLabel::verify() {
     return success();
 }
 
-// A const scan must name at least one node ID. An empty set opens no dataflow and
-// yields no row, so - like an empty db.scan_nodes_by_label - it is malformed IR
-// here rather than a scan that silently produces nothing.
-LogicalResult ConstScanNodes::verify() {
-    if (getNodeIDs().empty()) {
-        return emitOpError("requires at least one node ID");
-    }
-
-    return success();
-}
-
 // db.limit passes its columns straight through, so the results must be exactly
 // the input columns - same count, same types and in the same order.
 LogicalResult Limit::verify() {
--- a/query/ir/dialect/db/DBOps.cpp
+++ b/query/ir/dialect/db/DBOps.cpp
@@ -222,6 +222,17 @@ LogicalResult ScanNodesByLabel::verify() {
     return success();
 }
 
+// A const scan must name at least one node ID. An empty set opens no dataflow and
+// yields no row, so - like an empty db.scan_nodes_by_label - it is malformed IR
+// here rather than a scan that silently produces nothing.
+LogicalResult ConstScanNodes::verify() {
+    if (getNodeIDs().empty()) {
+        return emitOpError("requires at least one node ID");
+    }
+
+    return success();
+}
+
 // db.limit passes its columns straight through, so the results must be exactly
 // the input columns - same count, same types and in the same order.
 LogicalResult Limit::verify() {
--- a/query/ir/interpreter/NLExecutor.cpp
+++ b/query/ir/interpreter/NLExecutor.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <cmath>
 #include <limits>
+#include <span>
 #include <string>
 #include <string_view>
 #include <type_traits>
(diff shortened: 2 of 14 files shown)


query/ir/interpreter/NLExecutor.cpp
```diff
@@ -1041,6 +738,47 @@ void NLExecutor::runScanNodesByLabelLoop(NLExecutionContext* context, NLFunction
...
+    // local to this call, so a const scan nested in a cross product restarts from
+    // the first ID on every outer step, the same way runScanNodesLoop opens a
+    // fresh chunk writer each call.
+    size_t cursor = 0;
+
+    const auto runIteration = [&]() {
+        const size_t remaining = totalCount - cursor;
+        const size_t rows = std::min(chunkSize, remaining);
+
+        std::vector<NodeID>& raw = nodeIDs->getRaw();
```
rjb32, line 1067: We need to check if these NodeIDs are valid

rjb32: Done in e6718fc15. I filter invalid IDs (`GraphReader::graphHasNode`) in `translateConstScanLoop` rather than here in the executor: the graph view is fixed for the whole run, so validating once at translation matches how the label scan resolves its LabelSet and how the pipeline's ConstScanProcessor filters in prepare() — vs. re-checking every loop invocation (each outer step of a cross product) if done in the executor. Unknown IDs are dropped silently (Cypher `id(n) IN [...]` set-membership), surviving IDs keep listed order; covered by `constScanNodesDropsIDsWithNoNode`.

query/ir/interpreter/NLTranslator.cpp
```diff
@@ -425,6 +390,41 @@ void NLTranslator::translateScanByLabelLoop(const IteratorConfig& config,
...
+    std::vector<NodeID> constNodeIDs;
+    constNodeIDs.reserve(config._nodeIDs.size());
+    for (const int64_t nodeID : config._nodeIDs) {
+        const NodeID candidate(static_cast<uint64_t>(nodeID));
+        if (reader.graphHasNode(candidate)) {
+            constNodeIDs.emplace_back(candidate);
+        }
+    }
+
+    NLConstScanLoopData* loopData = _program->allocFunctionData<NLConstScanLoopData>(nodeIDs, constNodeIDs);
```
rjb32: Would be better to alloc the data before the loop and add nodeIDs directly in the vector in there

rjb32: Done in b8311945f. `NLConstScanLoopData` is now allocated first and the valid IDs are appended straight into its owned vector via `reserveNodeIDs`/`addNodeID` — no staging vector and no copy through `allocFunctionData`.

query/ir/interpreter/NLTranslator.cpp
```diff
@@ -425,6 +390,40 @@ void NLTranslator::translateScanByLabelLoop(const IteratorConfig& config,
...
+                                          NLLimitState* limit,
+                                          NLStmtContainer* body) {
+    // A const scan binds the same single node ID chunk as a plain scan.
+    const mlir::Value nodeChunk = loopBody.getArgument(0);
+    ColumnNodeIDs* nodeIDs = static_cast<ColumnNodeIDs*>(allocColumn(nodeChunk));
+
+    NLConstScanLoopData* loopData = _program->allocFunctionData<NLConstScanLoopData>(nodeIDs);
+    loopData->setLimit(limit);
+
+    // Resolve the constant i64 list into the NodeIDs the loop emits, appending each
```
rjb32: Remove this comment block

rjb32: Done in 84f9883d9.

query/ir/dialect/db/DBOps.cpp
```diff
@@ -222,6 +217,17 @@ LogicalResult ScanNodesByLabel::verify() {
...
+// A const scan must name at least one node ID. An empty set opens no dataflow and
+// yields no row, so - like an empty db.scan_nodes_by_label - it is malformed IR
+// here rather than a scan that silently produces nothing.
+LogicalResult ConstScanNodes::verify() {
+    if (getNodeIDs().empty()) {
+        return emitOpError("requires at least one node ID");
+    }
+
+    return success();
+}
```
cyrusknopf: Why is it invalid for the scan to be empty? Further, what if I provide a column of NodeIDs which are all invalid? Currently the invalid NodeIDs get removed from the scan, but is there a subsequent check that the post-filter is not empty?

rjb32: Yes indeed thanks. I removed the emptyness check

PR #744, opened 2026-07-17
MLIR: implement collect and unwind (collect buffer, per-group + per-element drains)

rjb32:
Implement collect and unwind for a buffer collected by collect


query/ir/dialect/db/DBOps.td
```diff
@@ -864,4 +864,60 @@ def GroupAggregate : TuringOp<"group_aggregate", [Pure]> {
...
 }
 
+/**
+* Collect
+*/
+def Collect : TuringOp<"collect", [Pure]> {
+  let summary = "Group rows by a key tuple and collect one column per group into a list";
+  let description = [{
+    Cypher `WITH k..., collect(x) AS xs`. Emits one row per distinct grouping-key
+    tuple: the key values then a list of the group's non-null collected values.
```
cyrusknopf, line 934: How does this work given that we don't support `WITH`?

cyrusknopf: Actually I think I am confusing the `COLLECT` keyword with the `collect()` function.

PR #745, opened 2026-07-23
MLIR: Implement node label constraints

cyrusknopf:
Allows queries which impose a label constraint on a node, e.g.
~~~
MATCH (n:Person) RETURN n
~~~

Currently this emits a scan nodes + filter by label; future work would involve writing an MLIR pass which would squash a scan + label filter into a scan nodes by label.

### DIFF
--- a/query/ir/codegen/DBProgramGenerator.cpp
+++ b/query/ir/codegen/DBProgramGenerator.cpp
@@ -720,6 +720,8 @@ void DBProgramGenerator::generatePropertyConstraints(const CypherAST* ast) {
 
     // Reused across iterations
     std::vector<const Expr*> constraintExprs;
+    using NodeDataPair = std::pair<const NodePattern*, const NodePatternData*>;
+    std::vector<NodeDataPair> labelConstrainedNodes;
     llvm::SmallVector<mlir::Value> columnsToFilter;
     llvm::SmallVector<const VariableDependency*> orderedVars;
     llvm::SmallVector<mlir::Type> resultTypes;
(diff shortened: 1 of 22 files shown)


query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -765,25 +754,17 @@ void DBProgramGenerator::generatePropertyConstraints(const CypherAST* ast) {
                     for (const EntityPropertyConstraint& constraint : nodeData->exprConstraints()) {
                         constraintExprs.push_back(constraint._expr);
                     }
-
```
rjb32: Does it mean that we do all the label filtering at the end now?

cyrusknopf: I think I made changes since your comment, but the current order is:
~~~
generateTraversal(ast);
resolveEdgeIdentities();
generatePropertyConstraints(ast);
generateLabelConstraints(ast);
generateFilters(ast);
generateOutput(ast);
~~~

so that maps to: inline property filters -> label filters -> where clause filters

rjb32: Please add samples of a db and nl program with label constraints so that we can see how practical it will be to push down the filters. I am especially worried for label filters if they are not inline

cyrusknopf: Added `samples/mlir/label_constraint*[.nl].mlir`

rjb32 (approved):

PR #752, opened 2026-07-28
Add v3 prefix to TuringShell for MLIR execution

cyrusknopf:
Enables execution of Cypher queries in TuringShell via the `#v3` prefix

### DIFF
--- a/tools/turingdb/TuringShell.cpp
+++ b/tools/turingdb/TuringShell.cpp
@@ -23,6 +23,25 @@
 #include "LineNoiseHandle.h"
 #include "LocalMemory.h"
 
+#include "mlir/Dialect/Func/IR/FuncOps.h"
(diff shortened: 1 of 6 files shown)


tools/turingdb/TuringShell.cpp
```diff
@@ -681,6 +700,131 @@ void queryCallback(size_t execCount, const Dataframe* df, tabulate::Table& table
...
+
+private:
+    tabulate::Table _table;
+    size_t _execCount {0};
+    size_t _rowCount {0};
+};
+
+}
+
+void TuringShell::runMLIRQuery(std::string_view query) {
```
rjb32, line 735: Would be worth factoring this into some QueryInterpreterV3 class

rjb32 (approved):

PR #755, opened 2026-07-29
MLIR: generate `ORDER BY` in frontend

rjb32:
Implements the ORDER BY clause in DBProgramGenerator.

Generates a `db.sort` from the projection's ORDER BY, ahead of SKIP and LIMIT, and stops the limit's producer walk at pipeline breakers so it no longer bounds the loop that fills their accumulator.

### DIFF
--- a/query/ir/codegen/DBProgramGenerator.cpp
+++ b/query/ir/codegen/DBProgramGenerator.cpp
@@ -1,6 +1,7 @@
 #include "DBProgramGenerator.h"
 
 #include <algorithm>
+#include <iterator>
 #include <memory>
 #include <string_view>
 #include <type_traits>
@@ -46,6 +47,8 @@
 #include "expr/UnaryExpr.h"
 #include "stmt/Limit.h"
 #include "stmt/MatchStmt.h"
+#include "stmt/OrderBy.h"
+#include "stmt/OrderByItem.h"
 #include "stmt/ReturnStmt.h"
 #include "stmt/Skip.h"
 #include "stmt/StmtContainer.h"
@@ -696,7 +699,7 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
 
     const Projection::Items& returned = proj->items();
 
-    std::unordered_map<std::string_view, mlir::Value> finalIdentities;
+    FinalIdentityMap finalIdentities;
     for (auto& [cypherVar, mlirCol] : _varMap) {
         const std::string_view varName = cypherVar->getName();
 
@@ -722,17 +725,7 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
             bioassert(findIt != end(finalIdentities), "Return variable '{}' not found", name);
             return findIt->second;
         } else {
-            const VarDecl* var = item->getExprVarDecl();
-            if (var) {
-                const std::string_view name = var->getName();
-                const auto findIt = finalIdentities.find(name);
-                if (findIt != end(finalIdentities)) {
-                    return findIt->second;
-                }
-            }
-
-            translateExpr(item);
-            return _exprMap.at(item);
+            return resolveExprColumn(finalIdentities, item);
         }
     };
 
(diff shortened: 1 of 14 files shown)


query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -777,6 +776,71 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
...
 
+void DBProgramGenerator::translateOrderBy(const OrderBy* orderBy,
+                                          const FinalIdentityMap& identities,
+                                          llvm::SmallVectorImpl<mlir::Value>& projected) {
+    const OrderBy::ItemVector& items = orderBy->getItems();
+    bioassert(!items.empty(), "ORDER BY without a key");
+
+    // Every projected column is sorted, so the projection stays row-aligned. A key the
+    // projection does not carry - the a.age of RETURN a ORDER BY a.age - is sorted as an
+    // extra column: it moves with its row but is never output
```
cyrusknopf: Not sure I understand what this comment is saying

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -777,6 +776,71 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
...
+
+        // A key the sorted columns already carry is sorted in place; any other key is
+        // appended, so its index is one past the columns collected so far
+        const auto findIt = std::ranges::find(sorted, keyColumn);
+        const size_t keyIndex = static_cast<size_t>(std::distance(sorted.begin(), findIt));
+        if (findIt == sorted.end()) {
+            sorted.push_back(keyColumn);
+        }
+
+        keyColumns.push_back(static_cast<int64_t>(keyIndex));
```
cyrusknopf: What is the keyIndex used for? and any key which is not in sorted has the same index, is that fine?

cyrusknopf (commented):
Could we add some ORDER BY query tests in EquivalenceTest.cpp?

query/ir/codegen/DBProgramGenerator.cpp
rjb32: We should not use a signed type for an index here

rjb32:
> Could we add some ORDER BY query tests in EquivalenceTest.cpp?

Done. I had to make another helper expectEquivalentOrdered to be able to test the sorting itself

query/ir/codegen/DBProgramGenerator.cpp
rjb32: Yes the comment has been fixed. It means that basically ORDER BY sorts all the columns in the projection by the ordering key, here a.age and that a.age is sorted but never output in this case because it is not returned, it is just the key used for sorting.

query/ir/codegen/DBProgramGenerator.cpp
rjb32: Unfortunately MLIR uses signed integers for indexes

rjb32: keyIndex is the positions of the columns against which the sort will be done, sort by c1,c2 for example, among the columns given to db.sort.

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -91,6 +95,27 @@ EdgeMetadata::EdgeType reverseEdge(EdgeMetadata::EdgeType type) {
...
 }
 
+// The index of the projected column @param key reads, or the item count when the
+// projection does not carry it. The projection is one column per return item, so the
+// index is the column's index too.
+//
+// A key and the item it reads are separate AST nodes - the n.name of
+// RETURN n.name ORDER BY n.name is written twice, so it is parsed twice - which is why
+// the two are matched by structure rather than by identity
+size_t findProjectedColumn(const Projection::Items& items, const Expr* key) {
```
rjb32, line 105: Return std::optional size_t instead

PR #758, opened 2026-07-30
Add support for optional procedure arguments

cyrusknopf:


procedures/ProcedureTypeVector.h
```diff
@@ -63,7 +64,23 @@ class ProcedureTypeVector {
...
+        constexpr bool optional = false;
+        _values.emplace_back(name, type, optional);
+    }
+
+    void addOptional(std::string_view name, ProcedureType type) {
+        constexpr bool optional = true;
+        _values.emplace_back(name, type, optional);
+    }
+
+    size_t requiredCount() const {
```
rjb32: Maintain a count of optional or required arguments so far instead of doing a loop here

cyrusknopf: Done

rjb32 (approved):

PR #754, opened 2026-07-29
Implement a neighbourhood random sample `CALL` procedure

cyrusknopf:

PR #759, opened 2026-07-30
MLIR: support CALL of registered procedures end to end

rjb32:
Runs a Cypher CALL of a registered procedure through the MLIR engine, from the frontend down to the interpreter.

A procedure that declares a row-aligned argument now has to report the input row of the rows it emits: registration refuses one that does not, and an argument read once per call rather than per row is declared as constant.

Works today:

```cypher
// a call is the whole query
CALL db.labels() YIELD id, label RETURN id, label

// no YIELD: every return value the procedure declares is emitted
CALL db.labels()

// a procedure declaring no return value, driven for its effect alone
CALL test.sideEffect()

// YIELD renames what the query reads, and may take a subset of the return values
CALL db.history() YIELD commit AS commitHash, nodeCount, edgeCount, partCount
RETURN commitHash, nodeCount, edgeCount, partCount

// YIELD ... WHERE filters the rows the call emitted
CALL db.labels() YIELD label WHERE label = 'Manager' RETURN label

// correlated: the matched node is the argument, and rides through the call
MATCH (n) CALL test.nodeScore(n) YIELD score RETURN n, score

// the matched node rides through the call even where the projection drops it: every
// column in flight is carried, which the procedure's input-row report is what allows
MATCH (n) CALL test.doubleNodeID(n) YIELD doubled RETURN doubled

// uncorrelated after a MATCH: the call is crossed with it
MATCH (n) CALL db.labels() YIELD label RETURN n, label

// several yields off one call, here two lists
MATCH (n) CALL test.nodeTags(n) YIELD tags, coords RETURN n, tags, coords

// chained calls, the second's WHERE reading what the first yielded
MATCH (n) CALL test.expandNodeID(n) YIELD copy
          CALL test.fanOutNodeID(n) YIELD value WHERE copy > 100 RETURN value

// one call per end of a traversal, projecting an expression over both yields
MATCH (n)-->(m) CALL test.expandNodeID(n) YIELD copy AS x
                CALL test.fanOutNodeID(m) YIELD value AS y RETURN x + y

// chained calls and no MATCH: the first opens the dataflow, and its yield is both the
// next call's argument and carried on past it
CALL test.twoNodes() YIELD id
CALL test.expandNodeID(id) YIELD copy
CALL test.fanOutNodeID(id) YIELD value
RETURN id, copy, value

// three calls reading nothing of each other, so all three are crossed
CALL test.twoNodes() YIELD id AS a
CALL test.twoNodes() YIELD id AS b
CALL test.twoNodes() YIELD id AS c
RETURN a, b, c
```

Added a test for GNN-like sampling:
```cypher
MATCH (n:Seed)
CALL gnn.neighbourhoodSample(n, 3, 42)     YIELD tgt AS hop1
CALL gnn.neighbourhoodSample(hop1, 2, 42)  YIELD tgt AS hop2
CALL gnn.neighbourhoodSample(hop2, 2, 42)  YIELD tgt AS hop3
RETURN n, hop1, hop2, hop3
``` 

Added a test for a call crossing and a call joining a hop chain:
```cypher
CALL test.firstNodes(2) YIELD person
MATCH (person)-->(m)
MATCH (m)-->(z)
CALL test.offsetPair(17) YIELD offset
RETURN person, z, offset

CALL test.firstNodes(2) YIELD person AS a
MATCH (a)-->(m)
CALL test.firstNodes(3) YIELD person AS z
MATCH (m)-->(z)
RETURN a, m, z
```


query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -332,62 +348,103 @@ void DBProgramGenerator::generateTraversal(const CypherAST* ast) {
...
-    for (auto& [var, values] : _varMap) {
-        const mlir::Value column = values.back();
+    const auto isLive = [&](const mlir::Value column) {
         mlir::Operation* const definingOp = column.getDefiningOp();
         mlir::Block* const definingBlock = definingOp
             ? definingOp->getBlock()
             : mlir::cast<mlir::BlockArgument>(column).getOwner();
-        if (definingBlock != insertionBlock) {
+
+        return definingBlock == insertionBlock;
```
rjb32: Why is this true?

procedures/HistoryProcedure.cpp
```diff
@@ -102,20 +117,11 @@ void HistoryProcedure::execute(ProcedureState* proc) {
...
-            } else if (tx->writingPendingCommit()) {
-                data._commit = tx->get<PendingCommitWriteTx>().commitBuilder()->commit();
-            }
-            bioassert(data._commit, "headCommitHash not found");
-        }
-        break;
-
+        case ProcedureState::Step::PREPARE:
         case ProcedureState::Step::RESET:
+            // Walking the history consumes the cursor, so a rewind seeks back to the
```
rjb32: Why can't we just store the commit hash of where we started?

procedures/Procedure.h
```diff
@@ -49,6 +49,10 @@ class Procedure {
     void addArgument(std::string_view name, ProcedureType type);
     void addOptionalArgument(std::string_view name, ProcedureType type);
 
+    // True if a procedure supports indices to report input rows corresponding to its output rows
```
rjb32: Rename into setIndices/getIndices

procedures/ProcedureData.h
```diff
@@ -39,9 +42,19 @@ class ProcedureData {
         _returnColumns[i] = col;
     }
 
+    // The indices column for procedures reporting indices corresponding to their input.
+    ColumnVector<size_t>* getInputRowIndices() const {
```
rjb32: Add a type alias in the class to define ColumnIndices. Check if ColumnIndices is not defined in a header somewhere

query/ir/codegen/DBProgramGenerator.h
```diff
@@ -58,6 +62,22 @@ class DBProgramGenerator {
...
 
+    // The column a CALL produced for each of its yielded return values, named as the
+    // query knows it - its alias when the YIELD renamed it. A yielded variable appears in
+    // no pattern, so it is not a VDG variable and cannot live in _varMap; the projection
+    // and symbol translation consult both. Kept in yield order, because a standalone CALL
+    // has no projection of its own and emits its columns in that order.
+    using YieldedColumn = std::pair<std::string_view, mlir::Value>;
+    std::vector<YieldedColumn> _yieldedColumns;
+
+    // What a CALL left behind: dropped from its carry set because nothing read it again, so
```
rjb32: Rephrase this comment more concisely

query/ir/codegen/DBProgramGenerator.cpp
rjb32: Answered in a comment on the lambda (ec9269543).

The rule is about row alignment, not visibility: a chunk holds the rows of the loop whose body binds it, so only a column bound in the insertion block is row-aligned with the rows flowing past that point. A column bound in an enclosing block — or in a loop the dataflow has already left — holds a different row set, and an op that consumes the whole row set (a filter, a call's carry set) would pair its rows with unrelated ones.

Worth noting the check is deliberately stricter than dominance: an outer block's value *does* dominate the insertion point, it is just the wrong rows. That is why the test is `==` on the block rather than a dominance query.

procedures/HistoryProcedure.cpp
rjb32: Done in 19be364a8 — no reason, that is simply better. `Data` now keeps `_headCommit`, resolved once at PREPARE, and `_commit` is just the walk cursor; RESET restores the cursor from it.

That drops the re-derivation (and the branch over the three transaction kinds) from every rewind. I stored the `const Commit*` rather than the hash, so the frozen-commit path also skips a `getCommitSafe` lookup on rewind.

`MLIRCallProcedureTest.cypherCrossedCallRewindsItsCursor` covers the RESET path and still passes.

procedures/Procedure.h
rjb32: Renamed in c5d5c5558: `reportsInputRows()`/`setReportsInputRows()` → `getIndices()`/`setIndices()`, member `_reportsInputRows` → `_indices`. Updated the two lowering/translation checks, the NLOps.td doc reference, and the three test call sites.

procedures/ProcedureData.h
rjb32: Done in b7e8dfb9e.

`ColumnIndices` does already exist — `storage/columns/ColumnIndices.h:7`, `using ColumnIndices = ColumnVector<size_t>`. I did not include it here though: `ProcedureData.h` deliberately forward-declares `ColumnVector`, and it is pulled in via `Procedure.h` by a lot of translation units, so including it would push `ColumnVector.h` into all of them for one pointer member.

So this is a class-local alias to the same type, which is what `storage/iterators/TombstoneFilter.h:37` does for the same reason. The comment points at the global one. Happy to switch to the include if you would rather have a single definition.

query/ir/codegen/DBProgramGenerator.h
rjb32: Shortened in 5a4b1a7dd, five lines to two:

```cpp
// Columns a CALL dropped from its carry set because nothing read them again. Their last
// value belongs to a loop the dataflow has left, so collectLiveColumns must skip them.
```

procedures/Procedure.cpp
```diff
@@ -26,6 +26,10 @@ void Procedure::setDeallocCallback(DeallocCallback cb) {
     _deallocCallback = cb;
 }
 
+void Procedure::setIndices(bool indices) {
+    _indices = indices;
+}
+
```
cyrusknopf: I do this slightly differently in #754 , we should compare and see which pattern we want

query/ir/codegen/DBProgramGenerator.h
```diff
@@ -91,6 +122,52 @@ class DBProgramGenerator {
...
+    void dropUnusedLiveColumns(const VariableNames& usedAfterCall, LiveColumns& live);
+
+    void generateCrossedCall(std::string_view procedureName,
+                             llvm::ArrayRef<mlir::Attribute> yieldedNames,
+                             llvm::ArrayRef<std::string_view> yieldedVariables,
+                             const LiveColumns& live);
+
+    // Collect those columns. One bound in another block is skipped: an op here can only
+    // take what this block binds.
+    void collectLiveColumns(LiveColumns& live);
```
cyrusknopf: What are "those" columns?

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -363,62 +377,104 @@ void DBProgramGenerator::generateTraversal(const CypherAST* ast) {
...
-    llvm::SmallVector<const VariableDependency*> orderedVars;
-    for (auto& [var, values] : _varMap) {
-        const mlir::Value column = values.back();
+    // A chunk holds the rows of the loop whose body binds it, so only a column bound in this
+    // block is row-aligned with the rows flowing past this point. One bound in an enclosing
+    // block - or in a loop the dataflow has already left - holds a different row set, and an
+    // op that consumes the whole row set (a filter, a call's carry set) would pair its rows
+    // with unrelated ones. Note this is stricter than dominance on purpose: an outer block's
+    // value does dominate here, it is just the wrong rows.
+    const auto isLive = [&](const mlir::Value column) {
```
cyrusknopf: If I understand correctly this comment can be simplified to: "a column is live if it is defined in the current block"

test/query/ir/MLIRCallProcedureTest.cpp
```diff
@@ -0,0 +1,2207 @@
...
+    constexpr const char* nlProgram = R"mlir(
+func.func @main() {
+  %call = nl.procedure("db.labels") yields ["id", "label"]
+  %rows = nl.procedure_init(%call, (), {}) : (!nl.procedure_state) -> !nl.iter<!nl.chunk<!storage.label_id>, !nl.chunk<!storage.string>>
+  nl.for %ids, %labels in %rows : !nl.iter<!nl.chunk<!storage.label_id>, !nl.chunk<!storage.string>> {
+    nl.output(%ids, %labels) : !nl.chunk<!storage.label_id>, !nl.chunk<!storage.string>
+  }
+  return
+}
+)mlir";
```
cyrusknopf, line 2339: Can we have some example NL and DB programs as samples in `samples/mlir`?

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -705,6 +761,484 @@ void DBProgramGenerator::moveComponentToFactor(TranslatedComponent& component,
...
+            // declaration read above.
+        break;
+
+        case Expr::Kind::_SIZE:
+            throw FatalException("Invalid expression kind.");
+        break;
+    }
+}
+
+void DBProgramGenerator::dropUnusedLiveColumns(const VariableNames& usedAfterCall, LiveColumns& live) {
```
cyrusknopf: Doesn't this feel like something to be done on an optimisation pass? It feels like DCE

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -705,6 +761,484 @@ void DBProgramGenerator::moveComponentToFactor(TranslatedComponent& component,
...
+    LiveColumns live;
+    collectLiveColumns(live);
+
+    // A call reading none of those rows produces the same rows for every one of them,
+    // which is their cartesian product rather than anything a carry set can express. The
+    // choice reads the unfiltered set: rows in flight have to be paired with the call's
+    // even when the projection returns none of their columns, or the query would lose
+    // their cardinality.
+    if (!live._columns.empty() && inputs.empty()) {
+        generateCrossedCall(procedureName, yieldedNames, yieldedVariables, live);
```
cyrusknopf: Do we want separate logic for cross product of a MATCH and a CALL? Ideally I would want centralised cross product logic to join islands regardless of whether they are from a MATCH or a CALL. I think it would involve maintaining CALLs in the connected components (we already tracked connected components in `generateTraversal`

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -851,6 +1385,19 @@ void DBProgramGenerator::generateOutput(const CypherAST* ast) {
 
     const ReturnStmt* rtn = sglPart->getReturnStmt();
     if (!rtn) {
+        // A standalone CALL has no projection of its own: what it emits is the columns it
+        // yielded, in the order the procedure declares them. A call yielding none - a
+        // procedure declaring no return value - emits nothing at all, so the query is the
+        // drive and no output op is generated.
```
cyrusknopf: Verbose

query/ir/lowering/DBLowering.cpp
```diff
@@ -1401,6 +1479,125 @@ void DBLowering::lowerUnwindCollect(mlir::db::UnwindCollect unwindCollect) {
...
+    // column, with at least one yield, so the ranges below line up. Each yielded name
+    // resolves to one of the procedure's declared return values, whose type is the
+    // chunk the result is read as; an unknown name throws here.
+    const mlir::ArrayAttr yields = call.getYields();
+    llvm::SmallVector<mlir::Type, 4> chunkTypes;
+    for (const mlir::Attribute yield : yields) {
+        const llvm::StringRef name = mlir::cast<mlir::StringAttr>(yield).getValue();
+        const size_t returnIndex = procedure->getReturnValueIndex(std::string_view(name.data(), name.size()));
+
+        chunkTypes.push_back(procedureChunkType(_builder, procedure->getReturnValueType(returnIndex)));
```
cyrusknopf: I think this could benefit from being split onto a few lines

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -705,6 +761,484 @@ void DBProgramGenerator::moveComponentToFactor(TranslatedComponent& component,
...
+}
+
+void DBProgramGenerator::collectNamesUsedAfterCall(const CallStmt* call,
+                                                   llvm::ArrayRef<const CallStmt*> laterCalls,
+                                                   const ReturnStmt* returnStmt,
+                                                   VariableNames& used) const {
+    // This call's own YIELD ... WHERE filters the rows it emitted, so it runs after the
+    // call and its variables have to survive it. Its arguments do not: they are operands of
+    // the call, read where it sits rather than past it.
+    collectYieldWhereNames(call, used);
```
cyrusknopf: Why is this true? In v2 arguments do live past a CALL

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -705,6 +761,484 @@ void DBProgramGenerator::moveComponentToFactor(TranslatedComponent& component,
...
+            continue;
+        }
+
+        for (const Expr* argument : *arguments) {
+            collectExprNames(argument, used);
+        }
+    }
+
+    // With no projection the query emits the columns the calls yielded, in yield order,
+    // and nothing besides - so every yielded variable is read and no pattern variable is.
```
cyrusknopf: What does "no pattern variable is [read]"?

query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -705,6 +761,484 @@ void DBProgramGenerator::moveComponentToFactor(TranslatedComponent& component,
...
+    // RETURN * names no variable of its own, so anything in scope may be projected and
+    // none of it can be called dead.
+    if (projection->isReturningAll()) {
+        for (const auto& [variable, columns] : _varMap) {
+            used.insert(variable->getName());
+        }
+
+        for (const YieldedColumn& yieldedColumn : _yieldedColumns) {
+            used.insert(yieldedColumn.first);
+        }
```
cyrusknopf: If I'm not mistaken the `RETURN *` should already expand the returned items in the analyser? So it shouldn't be necessary (potentially: wrong) to iterate over `_varMap` to determine what should be returned

query/ir/codegen/DBProgramGenerator.h
```diff
@@ -64,6 +68,19 @@ class DBProgramGenerator {
     // Maps each WHERE clause expression to the MLIR value it produces
     ExprValueMap _exprMap;
 
+    // The column a CALL produced for each of its yielded return values, named as the
```
rjb32, line 87: Remove comment

query/ir/codegen/DBProgramGenerator.h
```diff
@@ -64,6 +68,19 @@ class DBProgramGenerator {
...
+    // query knows it - its alias when the YIELD renamed it. A yielded variable appears in
+    // no pattern, so it is not a VDG variable and cannot live in _varMap; the projection
+    // and symbol translation consult both. Kept in yield order, because a standalone CALL
+    // has no projection of its own and emits its columns in that order.
+    using YieldedColumn = std::pair<std::string_view, mlir::Value>;
+    std::vector<YieldedColumn> _yieldedColumns;
+
+    // Columns a CALL dropped from its carry set because nothing read them again. Their last
```
rjb32: Replace column by saying that _deadVariables is the set of variables not read after the call so there is no point to put them in the carried set of the call.

query/ir/codegen/DBProgramGenerator.h
```diff
@@ -72,6 +89,20 @@ class DBProgramGenerator {
         llvm::SmallVector<mlir::Value> _columns;
     };
 
+    // The columns live at the current insertion point, and what each belongs to so the
```
rjb32: We need this to put the live columns after the CALL into the carried set of the CALL.
For example in:
MATCH (n)-->(m)
CALL proc(n) YIELD res
RETURN res, m

query/ir/codegen/DBProgramGenerator.h
```diff
@@ -94,6 +125,52 @@ class DBProgramGenerator {
...
+    void generateCrossedCall(std::string_view procedureName,
+                             llvm::ArrayRef<mlir::Attribute> yieldedNames,
+                             llvm::ArrayRef<std::string_view> yieldedVariables,
+                             const LiveColumns& live);
+
+    // Collect those columns. One bound in another block is skipped: an op here can only
+    // take what this block binds.
+    void collectLiveColumns(LiveColumns& live);
+
+    // Rebind them to an op's results, so later ops read what it produced. firstResult is
```
rjb32, line 183: Change comment to say that this is to rebind live columns after a call that has a carried set of live columns

query/ir/dialect/db/DBOps.cpp
```diff
@@ -358,6 +358,48 @@ LogicalResult RemoveDuplicates::verify() {
     return verifyPassThrough(getOperation(), getColumns(), getResults());
 }
 
+// A call produces one column per yielded return value then one per carried column, so
```
rjb32, line 407: Remove comment

procedures/ProcedureState.h
```diff
@@ -35,6 +35,14 @@ class ProcedureState {
 
     void finish() { _finished = true; }
 
+    void setData(ProcedureData* data) { _data = data; }
+    void setProcedure(const Procedure* procedure) { _procedure = procedure; }
```
rjb32: Why setProcedure and setContext need to be put in ProcedureState?

procedures/Procedure.cpp
rjb32: I rebased against main and adopted your approach now

query/ir/codegen/DBProgramGenerator.h
rjb32: Removed

query/ir/codegen/DBProgramGenerator.cpp
rjb32: Done

query/ir/codegen/DBProgramGenerator.cpp
rjb32: Done

query/ir/lowering/DBLowering.cpp
rjb32: Done

test/query/ir/MLIRCallProcedureTest.cpp
rjb32: Done

query/ir/codegen/DBProgramGenerator.cpp
rjb32: Yes it would be a better shape to have one unified procedure for the cross product.
But today in this PR, the cross product is decided before CALL statements are processed, in generateTraversal. So this is where it would require integration of the CALLs with the VDG to use the same logic based on components identification in the VDG. For now we do codegen of calls afterwards to minimise changes to the VDG until we can come with a better approach.

procedures/ProcedureState.h
rjb32: `setProcedure` is gone. `_procedure` was write-only: `ProcedureState` had no `getProcedure()`, and the only two accesses in the tree were the two writes. Removed the member and both writes.

`setContext` stays for now. The context is read by every procedure through `proc->getContext()` — 17 sites across 14 procedures — and the only channel a procedure gets is the `ProcedureState*`, since `ExecuteCallback` is `void(*)(ProcedureState*)`.

Agreed it does not belong in the state of ONE procedure: `ProcedureContext` is ambient (graph, view, transaction, procedure manager, chunk size, list buffer), one per query. Fixing it means widening the callback to `void(*)(ProcedureState*, const ProcedureContext*)`, which touches all 14 procedures — left out of this PR.

rjb32:
FYI @cyrusknopf

PR #762, opened 2026-08-03
MLIR: Implement SET

cyrusknopf:


query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -836,6 +839,81 @@ void DBProgramGenerator::generateCreate(const CypherAST* ast) {
     }
 }
 
+mlir::Value DBProgramGenerator::resolveEntityColumn(std::string_view varName) {
```
rjb32, line 850: Are we sure that we don't already have something like this?

cyrusknopf: There isn't something exactly like this. But we should relook at all the similar functions at some point soon to reconcile the approach we want to take

rjb32 (approved):

PR #769, opened 2026-08-03
MLIR: various fixes for collect and aggregates

rjb32:
- `nl.collect_update` and `nl.group_aggregate_update` now reject input columns that are not row-aligned with the grouping keys, instead of folding rows into the wrong group (shorter column) or reading past the group assignments (longer column).
- `nl.collect` / `nl.unwind_collect` now reject an `nl.collect_buffer` that is already drained, instead of letting the second drain rebind the first's output columns to its own incompatible types.
- `db.collect` / `db.unwind_collect` now reject a trailing result whose type does not match the collected column, instead of letting a mismatched list element type or unwound value type verify and then be silently discarded by the lowering.
- `QueryInterpreterV3` now carries the `catch (...)` arm `QueryInterpreterV2` has at every stage, so a throw not derived from `std::exception` is reported as an error status instead of unwinding out of the shell.

### DIFF
--- a/query/ir/dialect/db/DBOps.cpp
+++ b/query/ir/dialect/db/DBOps.cpp
@@ -487,6 +487,18 @@ LogicalResult Collect::verify() {
         return emitOpError("collected result must be a list column");
     }
 
+    // The list holds the values of the collected column, so its element is that
+    // column's type. DBLowering builds the emit chunk from the collected column and
+    // rebinds this result to it, so an element type that disagrees is discarded rather
+    // than honoured - the verifier is the only place it can be caught.
+    const ColumnType valueColumn = llvm::cast<ColumnType>(columns[keyCount].getType());
+    const storage::ListType collectedList = llvm::cast<storage::ListType>(listColumn.getType());
+    if (collectedList.getElementType() != valueColumn.getType()) {
+        return emitOpError("collected result must be a list of ") << valueColumn.getType()
+                                                                  << ", the collected column's type, but collects "
+                                                                  << collectedList.getElementType();
+    }
+
     return success();
 }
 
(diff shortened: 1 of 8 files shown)


query/ir/dialect/db/DBOps.cpp
```diff
@@ -519,5 +531,13 @@ LogicalResult UnwindCollect::verify() {
...
     }
 
+    // The unwound value is one element of the collected column, so the trailing result
+    // keeps that column's type. DBLowering builds the emit chunk from the collected
+    // column and rebinds this result to it, so a type that disagrees is discarded rather
+    // than honoured - the verifier is the only place it can be caught.
+    if (columns[keyCount].getType() != results[keyCount].getType()) {
+        return emitOpError("unwound value result must have the same type as the collected column");
+    }
+
```
cyrusknopf, line 541: Not sure a comment here is needed at all, it is quite clear what the code does

query/ir/dialect/db/DBOps.cpp
```diff
@@ -487,6 +487,18 @@ LogicalResult Collect::verify() {
...
+    // column's type. DBLowering builds the emit chunk from the collected column and
+    // rebinds this result to it, so an element type that disagrees is discarded rather
+    // than honoured - the verifier is the only place it can be caught.
+    const ColumnType valueColumn = llvm::cast<ColumnType>(columns[keyCount].getType());
+    const storage::ListType collectedList = llvm::cast<storage::ListType>(listColumn.getType());
+    if (collectedList.getElementType() != valueColumn.getType()) {
+        return emitOpError("collected result must be a list of ") << valueColumn.getType()
+                                                                  << ", the collected column's type, but collects "
+                                                                  << collectedList.getElementType();
+    }
```
cyrusknopf, line 500: Same comment on probably unnecessary comment

query/ir/interpreter/NLTranslator.cpp
```diff
@@ -216,6 +216,19 @@ bool isConstantLike(mlir::Value value) {
...
+// second drain would rebind the first's outputs to its own, incompatible column types
+// (a per-group list cell against an unwound value), leaving the first to emit through
+// the other's columns. Generated IR names each accumulator from a single drain, so a
+// second one means malformed IR - the accumulate side rejects a second
+// nl.collect_update the same way.
+void throwIfAlreadyDrained(const NLCollectState* state) {
+    if (state->getValueOutput()) {
+        throw IRException("an nl.collect_buffer must be drained by a single nl.collect or nl.unwind_collect");
+    }
+}
```
cyrusknopf, line 230: Why is this an `IRException` but the errors in `NLExecutor` are `bioassert`? are both of these not due to malformed IR (internal error rather than user error)

rjb32: Agree, will do another PR for converting these internal errors to bioassert. Its is a bit like the difference between PlanException vs bioassert in v2. The difference is thin.

PR #783, opened 2026-08-13
MLIR: Split filters to be maximally granular

cyrusknopf:
Splitting each predicate (separated by `AND` clauses) into its own filter operation.

Motivations:

- short circuiting (cardinality reduction reduces row count for subsequent filters)
- more malleable code for predicate pushdown/peephole optimisations


rjb32 (approved):

rjb32:
Please add an mlir example that shows the splitting

PR #788, opened 2026-08-14
MLIR: Refactor operator dispatch to reduce code repetition

cyrusknopf:
The `lowerX` or `translateX` functions had a lot of code duplication. Refactor to use templated methods instead, to reduce the bloat


rjb32 (approved):
Good one

PR #792, opened 2026-08-17
MLIR: fix SKIP & LIMIT over a projection of constants

rjb32:
Fixes for SKIP and LIMIT over a constant.
```cypher
MATCH (n) RETURN n LIMIT 0

MATCH (n) RETURN n SKIP 1 LIMIT 1

MATCH (n) RETURN n SKIP 1 LIMIT 0

RETURN 42 LIMIT 0

RETURN 42 LIMIT 1

RETURN 42 SKIP 1 LIMIT 1

### DIFF
--- a/query/ir/interpreter/NLExecutor.cpp
+++ b/query/ir/interpreter/NLExecutor.cpp
@@ -344,6 +344,13 @@ void blockRepeatColumn(const Column* input, size_t factor, size_t outputRowCount
     }
 }
 
+// Block-repeat for a constant chunk. A constant already stands for every row of
+// the step, so there is nothing to repeat: all the cut leaves in question is
+// whether a row survived, and assignFromLine empties the output when none did.
+void blockRepeatConstColumn(const Column* input, size_t factor, size_t outputRowCount, Column* output) {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: query/ir/interpreter/NLExecutor.cpp [hunk 1]
```diff
@@ -386,6 +386,14 @@ void copyRangeColumn(const Column* input, size_t inputOffset, size_t rowCount, C
     std::copy(first, first + rowCount, outputRaw.begin());
 }
 
+// Range copy for a constant chunk. A constant holds one value standing for every
```
rjb32: remove comment

rjb32: Removed. Dropped the same comment on the blockRepeatConstColumn twin too.

cyrusknopf:
What was the bug here?

rjb32:
> What was the bug here?

There was no specific treatment of a column that is constant like in the way copy and output functions were chosen for addTruncateColumn and addSkipColumn. It was triggering IRException("Unsupported chunk element type")

PR #794, opened 2026-08-17
MLIR: size a constant projection by the filtered relation

rjb32:
Fixing wrong row count post filtering when returning a constant projection.

A projection of constants alone was sized by the raw loop chunk rather than the relation the filter left standing, so MATCH (n) WHERE n.name = 'Remy' RETURN 5 emitted one row per node in the graph, as did count/sum over a constant under a WHERE.

```
MATCH (n) WHERE n.name = 'Remy' RETURN 5                     18 rows -> 1 row
MATCH (n) WHERE n.name = 'nobody' RETURN 5                   18 rows -> 0 rows
MATCH (n) WHERE n.name = 'Remy' RETURN count(42)             18      -> 1
MATCH (n) WHERE n.name = 'Remy' RETURN 1 AS x, sum(x)        18      -> 1
```

### DIFF
--- a/query/ir/lowering/DBLowering.cpp
+++ b/query/ir/lowering/DBLowering.cpp
@@ -1927,6 +1927,18 @@ void DBLowering::lowerFilter(mlir::db::FilterOp filter) {
         const mlir::Value outputChunk = nlFilter.getResult(columnIndex);
         _valueMap[filteredCol] = outputChunk;
     }
+
+    // The cardinality driver stands for the relation a projection of constants alone is a
+    // projection of, and this filter is what narrows that relation: leaving the driver on
+    // the unfiltered chunk would size the emission by the rows the filter dropped.
+    for (size_t columnIndex = 0; columnIndex < columnChunks.size(); columnIndex++) {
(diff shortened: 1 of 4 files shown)


query/ir/lowering/DBLowering.cpp
```diff
@@ -1927,6 +1927,18 @@ void DBLowering::lowerFilter(mlir::db::FilterOp filter) {
...
+
+    // The cardinality driver stands for the relation a projection of constants alone is a
+    // projection of, and this filter is what narrows that relation: leaving the driver on
+    // the unfiltered chunk would size the emission by the rows the filter dropped.
+    for (size_t columnIndex = 0; columnIndex < columnChunks.size(); columnIndex++) {
+        if (columnChunks[columnIndex] != _innermostCardinality) {
+            continue;
+        }
+
+        _innermostCardinality = nlFilter.getResult(columnIndex);
```
rjb32, line 1939: _innermostCardinality is the SSA value of the chunk with the innermost cardinality in the nl dialect program. We repoint it to the SSA value of the chunk after filtering

rjb32:
FYI @cyrusknopf

PR #798, opened 2026-08-18
MLIR: Fix generation for `MATCH ... CREATE`

cyrusknopf:
`MATCH ... CREATE` queries resulted in a single `CREATE` rather than 1-per `MATCH` row in cases where there are no properties.

This PR adds a "cardinality" argument to the `CREATE` MLIR ops which informs how many nodes/edges are to be created.

### DIFF
--- a/query/ir/codegen/DBProgramGenerator.cpp
+++ b/query/ir/codegen/DBProgramGenerator.cpp
@@ -922,6 +922,11 @@ void DBProgramGenerator::generateCreate(const CypherAST* ast) {
         }
     }
 
+    mlir::Value matchCardinality;
+    if (!knownVars.empty()) {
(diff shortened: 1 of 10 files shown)


rjb32:
Add a cypher unit test for the following query shape (match filtered + create):
```cypher
match (n) where n = 0 
create (n)-[:LIKES]->(c:Interest{name:"Cake"})
```

cyrusknopf:
> Add a cypher unit test for the following query shape (match filtered + create):
> 
> ```cypher
> match (n) where n = 0 
> create (n)-[:LIKES]->(c:Interest{name:"Cake"})
> ```

Done

rjb32 (approved):

PR #800, opened 2026-08-18
Make conversion functions comply with OpenCypher

cyrusknopf:
The OpenCypher spec requires conversion functions such as `toInteger` to return `NULL` if the input cannot be converted to the output type.

Before this PR, these functions would throw an exception if the conversion failed.

This PR makes the functions comply with OpenCypher by returning `NULL`, instead of throwing.


rjb32:
Please add cypher-level unit tests

cyrusknopf:
> Please add cypher-level unit tests

Done

rjb32 (approved):

PR #801, opened 2026-08-18
MLIR: implement list literals

rjb32:
Implements list literals from cypher to db and nl dialect. Reuses the db.constant operation.

    RETURN [1, 2, 3]           %0 = db.constant([1, 2, 3])            !db.column<!storage.list<i64>>
    RETURN ['one', 'two']      %0 = db.constant(["one", "two"])       !db.column<!storage.list<!storage.string>>
    RETURN [1, 'Hello', [1]]   %0 = db.constant([1, "Hello", [1]])    !db.column<!storage.list<!storage.list_element>>
    RETURN [1, null]           %0 = db.constant([1, unit])            !db.column<!storage.list<!storage.list_element>>
    RETURN []                  %0 = db.constant([])                   !db.column<!storage.list<!storage.list_element>>

Lists and embeddings are distinguished by the attribute type: lists use an ArrayAttr and embeddings use a dense array:

    db.constant(dense<[1.0, 2.0, 3.0]> : tensor<3xf32>)   !db.column<!storage.embedding>
    db.constant([1.0, 2.0, 3.0])                          !db.column<!storage.list<f64>>

Examples in a match statement:

    MATCH (n) RETURN [1, 2]              one row per matched node, each holding the list
    MATCH (n) RETURN n, [1, 2] SKIP 2    the list laid out across the surviving rows
    MATCH (n) RETURN [1, 2], count(n)    the list beside a tally of the whole scan
    MATCH (n) RETURN count([1, 2])       a list is never null, so this counts every row


cyrusknopf:
Why do we need a separate op for list literals?

rjb32:
> Why do we need a separate op for list literals?

Because there is too much type collision for type inference: db.constant already maps [1.0, 2.0]  to storage.embedding

cyrusknopf:
> > Why do we need a separate op for list literals?
> 
> Because there is too much type collision for type inference: db.constant already maps [1.0, 2.0] to storage.embedding

Is the solution then to not use type inference rather than to create a whole new operation? Since it is a constant we know the type and "compile" time so we do not strictly need inference

rjb32:
> > > Why do we need a separate op for list literals?
> > 
> > 
> > Because there is too much type collision for type inference: db.constant already maps [1.0, 2.0] to storage.embedding
> 
> Is the solution then to not use type inference rather than to create a whole new operation? Since it is a constant we know the type and "compile" time so we do not strictly need inference

Actually we can use db.constant for everything, `db.constant([1.5, 2.5])`can be inferred as `db.column<!storage.list<f64>>` because [1.5, 2.5] is ArrayAttr and ` db.constant(dense<[1.0,2.0]>)`can be inferred as embedding because it uses a DenseElementsAttr

rjb32:
We now use db.constant for every type of constant with no ambiguity between lists and embeddings.

cyrusknopf:
> We now use db.constant for every type of constant with no ambiguity between lists and embeddings.

Great

PR #817, opened 2026-08-24
MLIR: match several patterns on one node pair

rjb32:
MATCH (a)-[:KNOWS_WELL]->(b), (a)-[e1]->(b), (a)-[e2]->(b) RETURN count(*) failed the producesEdgeVar assert in codegen because the traversal read a cascaded merge edge as half of a (source, edge, target) triple.


cyrusknopf (approved):
Can we add DB and NL MLIR sample code for this query?

rjb32:
> Can we add DB and NL MLIR sample code for this query?

Yes done

PR #820, opened 2026-08-25
MLIR: reject an edge variable repeated in one pattern

rjb32:
Reject an edge variable repeated in one pattern.

```
MATCH (n:Person)-[e]->(m:Person)-[e]->(m)
RETURN n;

before, v3: PLAN_ERROR: Unexpected exception: Internal Error: The assertion
            '_varMap.contains(var)' failed at
            query/ir/codegen/DBProgramGenerator.cpp:776
            Component var e'0 not registered

after, v3:  PLAN_ERROR: Re-using the same edge variable in a single pattern is
            not supported
```

### DIFF
--- a/query/ir/frontend/cypher/VariableDependencyGraph.cpp
+++ b/query/ir/frontend/cypher/VariableDependencyGraph.cpp
@@ -30,6 +30,7 @@
 
 #include "BioAssert.h"
 #include "FatalException.h"
+#include "TuringException.h"
 
 using namespace db;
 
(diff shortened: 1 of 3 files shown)


query/ir/frontend/cypher/VariableDependencyGraph.cpp
```diff
@@ -130,8 +131,23 @@ void VariableDependencyGraph::registerPatternElement(const PatternElement* ptn)
 
     const auto& chain = ptn->getElementChain();
 
+    // Two patterns sharing an edge variable are joined on identity; one element naming it
+    // twice is rejected instead, as no edge is two hops of one element.
+    std::vector<const VarDecl*> edgesInElement;
```
rjb32, line 136: I don't really like this but the other alternative is std::unordered_set which may be overkill for most queries

cyrusknopf: Does this TU link to LLVM? If so why not SmallSet or similar

rjb32: No it does not link LLVM

PR #825, opened 2026-08-26
MLIR: Implement shortest path

cyrusknopf:


query/ir/interpreter/NLExecutor.cpp
```diff
@@ -2234,6 +2238,152 @@ void runProcedureDrive(NLExecutionContext* context,
...
+        return left.distance > right.distance;
+    }
+};
+
+// Weighted Dijkstra over out-edges from the buffered source set to the buffered target
+// set, summing the resolved weight property. Emits one row - the total weight and the
+// back-traced path (target-first, alternating node and edge IDs) - or no row when no
+// target is reachable. Ported from the V2 ShortestPathProcessor.
+template <SupportedType T>
+void shortestPathSearch(NLExecutionContext* context, NLShortestPathLoopData* loopData) {
```
rjb32, line 2663: Can not we reuse the shortest path utils class that we have in v2?

test/query/ir/ShortestPathLoweringTest.cpp
```diff
@@ -0,0 +1,302 @@
...
+                                PropertyTypeID weightID,
+                                NodeID source,
+                                NodeID target,
+                                int64_t weight) {
+        const EdgeRecord& edge = builder.addEdge(0, source, target);
+        builder.addEdgeProperty<types::Int64>(edge, weightID, weight);
+    }
+};
+
+TEST_F(ShortestPathLoweringTest, findsCheapestTwoHopPath) {
```
rjb32, line 222: These are only MLIR-level tests. Please add cypher level tests

query/ir/interpreter/NLExecutor.cpp
cyrusknopf: The shortest path logic in v2 lives solely in the ShortestPathProcessor from what I can see

test/query/ir/ShortestPathLoweringTest.cpp
cyrusknopf: Done

test/query/ir/CypherShortestPathTest.cpp
```diff
@@ -0,0 +1,213 @@
...
+                  "MATCH (a {name: 'A'}), (b {name: 'E'}) "
+                  "SHORTESTPATH(a, b, weight, d, p) RETURN d, p"),
+              expected);
+}
+
+TEST_F(CypherShortestPathTest, multiSourceMatchPicksCheapest) {
+    const std::vector<StringRowSink::Row> expected {{"3", "3, 4, 2, 2, 1"}};
+    EXPECT_EQ(run(_graphName,
+                  "MATCH (a), (b {name: 'D'}) WHERE a.name = 'A' OR a.name = 'B' "
+                  "SHORTESTPATH(a, b, weight, d, p) RETURN d, p"),
```
rjb32, line 186: Where the weight are coming from? Must it be a weight property?

rjb32:
@sulaimansuhas FYI if you can comment

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -2234,6 +2238,152 @@ void runProcedureDrive(NLExecutionContext* context,
...
+template <typename T>
+struct ShortestPathHeapNode {
+    NodeID id;
+    NodeID previous;
+    EdgeID edge;
+    T distance {0};
+};
+
+template <typename T>
+struct ShortestPathBest {
```
sulaimansuhas: I find this naming a bit weird. It has been changed from `HeapMapValues` which describes what it literally represents the values in the map used to check staleness of heap values.

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -2234,6 +2238,152 @@ void runProcedureDrive(NLExecutionContext* context,
     }
 }
 
+template <typename T>
+struct ShortestPathHeapNode {
```
sulaimansuhas: Also here I think it would be a bit clearer to stick to the old djistraka-style naming. `DjistrakaNode`.

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -2234,6 +2238,152 @@ void runProcedureDrive(NLExecutionContext* context,
...
+
+template <typename T>
+struct ShortestPathBest {
+    NodeID previous;
+    EdgeID edge;
+    T distance {0};
+};
+
+template <typename T>
+struct ShortestPathHeapOrder {
```
sulaimansuhas: Same here: `DjistrakaNodeComparator`

query/ir/interpreter/NLExecutor.cpp
cyrusknopf: I have reverted all names and comments to the original style

cyrusknopf:
@rjb32 @sulaimansuhas I kept as faithful as possible to the original implementation:
- kept existing comments
- kept existing naming scheme (`DijkstraX`)
- fill set and priority queues directly in each iteration of the MLIR NL loops, rather than buffer an intermediate column

rjb32:
Please add a unit test for a query also returning the matched variables, such as:
```
MATCH (a),(b)
SHORTESTPATH(a,b,weight,d,p)
RETURN a,d,p
```

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -3788,6 +3929,48 @@ void NLExecutor::runCollectLoop(NLExecutionContext* context, NLFunctionData* dat
...
+
+void NLExecutor::runShortestPathUpdate(NLExecutionContext* context, NLFunctionData* data) {
+    NLShortestPathUpdateData* update = static_cast<NLShortestPathUpdateData*>(data);
+    NLShortestPathState* state = update->getState();
+
+    const std::vector<NodeID>& nodes = update->getNodes()->getRaw();
+
+    // A target only needs to be a member; a source seeds the frontier at distance zero, so it
+    // goes through the typed accumulator.
+    if (update->isTarget()) {
```
rjb32, line 4683: Only targets get deduplicated using a set

test/query/ir/CypherShortestPathTest.cpp
cyrusknopf: Yes it must be a property, the third argument is the edge property that you want to use as weights, in this case, it is called `weight`

query/ir/interpreter/NLExecutor.cpp
cyrusknopf: This is consistent with the V2 implementation

cyrusknopf:
> Please add a unit test for a query also returning the matched variables, such as:
> 
> ```
> MATCH (a),(b)
> SHORTESTPATH(a,b,weight,d,p)
> RETURN a,d,p
> ```

This is not allowed due to the implementation of the shortest path. Added a test to ensure an error is thrown

query/ir/interpreter/NLExecutor.cpp
sulaimansuhas: This is a good point actually - we need to add some sort of deduplication for the source set here in the future.

rjb32 (approved):

PR #831, opened 2026-09-01
Implement `STARTS WITH`, `ENDS WITH`, `CONTAINS`

cyrusknopf:
Implements the Cypher `STARTS WITH`, `ENDS WITH`, and `CONTAINS` string operations for the MLIR V3 interpreter. Remains unsupported in v2.

Uses C++ standard library algorithms

### DIFF
--- a/test/query/ir/StringPredicateTest.cpp
+++ b/test/query/ir/StringPredicateTest.cpp
@@ -122,6 +122,104 @@ TEST_F(StringPredicateTest, combinedWithPropertyPredicates) {
                {{"Adam"}});
 }
 
+TEST_F(StringPredicateTest, countOverOrOfStringPredicates) {
+    expectRows("MATCH (n) WHERE n.name STARTS WITH 'C' OR n.name STARTS WITH 'A' RETURN count(n)",
+               {{"5"}});
+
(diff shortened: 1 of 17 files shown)


test/query/ir/StringPredicateTest.cpp
```diff
@@ -0,0 +1,127 @@
...
+        EXPECT_EQ(rows, expected) << "query: " << query;
+    }
+
+    const std::string _graphName = "simpledb";
+    std::unique_ptr<TuringTestEnv> _env;
+    std::unique_ptr<QueryInterpreterV3> _interpreter;
+};
+
+TEST_F(StringPredicateTest, startsWithLiteralPrefix) {
+    expectRows("MATCH (n) WHERE n.name STARTS WITH 'C' RETURN n.name",
```
rjb32, line 66: Add tests where these string expressions are mixed with another expression on properties with OR/AND

rjb32, line 66: Also add a test where the result is fed to COUNT

cyrusknopf: I have some at the bottom of this file, anything other tests you want to see specifically?

cyrusknopf: done

rjb32 (approved):

PR #833, opened 2026-09-01
MLIR: type labels and edgeType as owned strings

rjb32:
Fix the ORDER BY labels(n) segfault.


cyrusknopf:
#832 is a duplicate

rjb32:
> #832 is a duplicate

Fair enough thanks

PR #837, opened 2026-09-02
MLIR: fuse node-ID disjunction filters into const_scan_nodes

rjb32:
Add an MLIR pass turning a node scan filtered by a disjunction of node IDs into a const_scan_nodes.

### DIFF
--- a/query/ir/dialect/db/DBPasses.cpp
+++ b/query/ir/dialect/db/DBPasses.cpp
@@ -398,8 +398,11 @@ struct NodeIDScanChain {
 };
 
 bool matchNodeIDScanChain(FilterOp filter, NodeIDScanChain& chain) {
+    // The filter is replaced by a source, which yields the listed nodes and nothing else.
+    // Any other column it carries (a property read before the WHERE, say) has no filtered
+    // counterpart in that source to be rewired to, so such a filter has to stay.
     const Operation::operand_range columns = filter.getColumnsToFilter();
-    if (columns.empty()) {
+    if (columns.size() != 1) {
         return false;
     }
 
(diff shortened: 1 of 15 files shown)


query/ir/dialect/db/DBPasses.cpp
```diff
@@ -347,6 +349,146 @@ struct PushDownFilters : public impl::PushDownFiltersBase<PushDownFilters> {
...
+    }
+
+    // A carried column other than the scan is row-aligned with it and would be left
+    // behind by a source that reads nothing.
+    const bool carriesScanOnly = llvm::all_of(columns, [&](const Value column) {
+        return column == scanColumn;
+    });
+    if (!carriesScanOnly) {
+        return false;
+    }
```
cyrusknopf: What is this early return for? I am not sure I understand from the comment

query/ir/dialect/db/DBPasses.cpp
```diff
@@ -413,8 +413,9 @@ bool matchNodeIDScanChain(FilterOp filter, NodeIDScanChain& chain) {
...
     }
 
-    // A carried column other than the scan is row-aligned with it and would be left
-    // behind by a source that reads nothing.
+    // The filter is replaced by a source, which yields the listed nodes and nothing else.
+    // Any other column it carries (a property read before the WHERE, say) has no filtered
+    // counterpart in that source to be rewired to, so such a filter has to stay.
     const bool carriesScanOnly = llvm::all_of(columns, [&](const Value column) {
         return column == scanColumn;
     });
```
cyrusknopf: Okay now the comment is more clear, but should the condition be something like
`columns.size() == 1 && columns.front() == scanColumn`?

rjb32: Yes agree, fixed

PR #842, opened 2026-09-03
Fix: Write to what the same query created

rjb32:
DELETE and SET mistook the write-buffer offset a CREATE returns in its result column for a committed ID, thus hitting an unrelated committed entity instead.

```
CREATE (n:Tag {name: 'ephemeral'}) DELETE n
CREATE (a:Tag {name: 'kept'}), (b:Tag {name: 'dropped'}) DELETE b
CREATE (a:Tag {name: 'a'})-[e:LINK]->(b:Tag {name: 'b'}) DELETE e
CREATE (a:Tag {name: 'a'})-[:LINK]->(b:Tag {name: 'b'}) DETACH DELETE a
CREATE (a:Tag {name: 'a'})-[e:LINK]->(b:Tag {name: 'b'}) DELETE e, a

CREATE (n:Tag {name: 'x'}) SET n.age = 1
CREATE (n:Tag {name: 'x'}) SET n.name = 'y'
CREATE (n:Tag {name: 'x'}) SET n.name = 'y', n.age = 2
CREATE (a:Tag {name: 'a'})-[e:LINK]->(b:Tag {name: 'b'}) SET e.name = 'a -> b'
MATCH (p:Person {name: 'Remy'}) CREATE (t:Tag {name: 'x'}) SET p.age = 99
```

### DIFF
--- a/query/analyzer/WriteStmtAnalyzer.cpp
+++ b/query/analyzer/WriteStmtAnalyzer.cpp
@@ -63,16 +63,12 @@ void WriteStmtAnalyzer::analyze(const Stmt* stmt) {
 }
 
 void WriteStmtAnalyzer::analyze(const CreateStmt* createStmt) {
-    _hasCreate = true;
     if (createStmt->getPattern()) {
         analyze(createStmt->getPattern());
     }
 }
 
 void WriteStmtAnalyzer::analyze(const SetStmt* setStmt) {
-    if (_hasCreate) {
-        throwError("CREATE ... SET is not yet supported. Please use brace initialisation.", setStmt);
-    }
     for (SetItem* item : setStmt->getItems()) {
         analyze(item);
     }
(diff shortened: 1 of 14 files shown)


cyrusknopf:
I am already working on a fix for this. I think we should wait for my PR to be ready and compare the two fixes to see what we prefer.

The problem is more systemic than what this PR addresses: `CREATE ... RETURN` returns incorrect values

rjb32:
> I am already working on a fix for this. I think we should wait for my PR to be ready and compare the two fixes to see what we prefer.
> 
> The problem is more systemic than what this PR addresses: `CREATE ... RETURN` returns incorrect values

Ok let's wait for yours

cyrusknopf:
> > I am already working on a fix for this. I think we should wait for my PR to be ready and compare the two fixes to see what we prefer.
> > The problem is more systemic than what this PR addresses: `CREATE ... RETURN` returns incorrect values
> 
> Ok let's wait for yours

#844 is a start to fix the problem I mention above. A separate PR is needed to try and fix any issues with `CREATE ... DELETE`

PR #844, opened 2026-09-03
MLIR: Fix `CREATE ... RETURN`

cyrusknopf:
`CREATE (n:N) RETURN n` would return `0` (number of pending nodes in the `CommitWriteBuffer`) regardless of how many nodes already exist in the graph.

The correct behaviour is to return `total nodes allocated + number of nodes in CommitWriteBuffer`

### DIFF
--- a/query/ir/interpreter/NLExecutor.cpp
+++ b/query/ir/interpreter/NLExecutor.cpp
@@ -37,6 +37,7 @@
 #include "metadata/PropertyNull.h"
 #include "metadata/PropertyType.h"
 
+#include "reader/GraphReader.h"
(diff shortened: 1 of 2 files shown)


rjb32:
What if this number is used to create an edge from it?
CREATE (n:Person)
WITH n as new
CREATE (new)-[:KNOWS]->(p:Person)

cyrusknopf:
> What if this number is used to create an edge from it? CREATE (n:Person) WITH n as new CREATE (new)-[:KNOWS]->(p:Person)

Added a test for this case, and subsequent match to ensure CREATE succeeded

rjb32 (approved):

PR #848, opened 2026-09-03
Abstract Turing Proto Write Sink

sulaimansuhas:


net/decoders/TuringProtoDecoder.h
```diff
@@ -0,0 +1,388 @@
...
+#include "TuringException.h"
+
+#include "BioAssert.h"
+
+namespace net::proto {
+
+// A templated class containing the protocol decoding logic. It expects to be templated
+// with a sink class that provides the underlying storage to write the decoded bytes into
+// (the sink)
+template <typename Sink>
```
rjb32: There is no concept or abstract class that sets the requirements for the sink?

net/turing_proto_client/TuringClient.cpp
```diff
@@ -456,9 +458,11 @@ db::QueryStatus TuringClient::sendQuery(const std::string& query,
     _listBuffer.clear();
     db::DataframeManager dfMan;
     db::QueryStatus res;
-    std::vector<DecodedColumnSchema> colSchemas;
+    std::vector<DecodedColumnSchema> columnSchemas;
     db::Dataframe df;
-    TuringProtoDecoder decoder(_localMem, &dfMan, &_inBuf, &_embeddingBuffer, &_stringBuffer, &_listBuffer, colSchemas);
```
rjb32, line 461: The impact on the client is not massive

rjb32:
The principle looks great to me. I think we just need to be cautious of 2 things: centralise the types supported in containers and the types of containers, as much as we can, and constrain the interface of an acceptable sink class as much as possible to avoid drift.

net/decoders/TuringProtoDecoder.h
sulaimansuhas: I actually have it in a local commit and forgot to move it. I'll do it now.

net/turing_proto_client/TuringClient.cpp
sulaimansuhas: Yes it's just a bit of repackaging of existing components.

net/decoders/DecodedColumnSchema.h
```diff
@@ -0,0 +1,74 @@
...
+#include <stddef.h>
+#include <string>
+
+#include "Bitmask.h"
+#include "TuringProtoHeaders.h"
+
+namespace net::proto {
+
+// Per-column decode state carried across chunk and buffer boundaries: the null
+// bitmask and how many rows the column contains.
```
cyrusknopf: The first clause is fine, but after the ":" it just states the members of the class which is unnecessary IMO

net/decoders/DecodedColumnSchema.h
```diff
@@ -0,0 +1,74 @@
...
+public:
+    ProtoColumnState();
+    ProtoColumnState(const ProtoColumnState&);
+    ProtoColumnState(ProtoColumnState&&);
+    ProtoColumnState& operator=(const ProtoColumnState&);
+    ProtoColumnState& operator=(ProtoColumnState&&);
+    ~ProtoColumnState();
+
+    DynamicLargeBitMask<uint64_t>& getBitMask() { return _bitMask; }
+    const DynamicLargeBitMask<uint64_t>& getBitMask() const { return _bitMask; }
```
cyrusknopf: I would typedef this

net/decoders/DecodedColumnSchema.h
```diff
@@ -0,0 +1,74 @@
...
+    // across multiple chunks if needed.
+    DynamicLargeBitMask<uint64_t> _bitMask {0};
+    // The total number of rows that prefixes every column's wire bytes
+    WireSize _numRows {0};
+    // If an optional const column has a value - useful if the hasValue flag is split
+    // across packet boundaries
+    bool _constHasValue {false};
+};
+
+// One decoded column's schema: its wire header, its name, and its decode state.
```
cyrusknopf: Again the comment just spells out the struct members

net/decoders/TuringSink.h
```diff
@@ -0,0 +1,89 @@
...
+    template <typename T>
+    using ColumnOptVector = db::ColumnVector<std::optional<T>>;
+
+    template <typename T>
+    using ColumnConst = db::ColumnConst<T>;
+
+    template <typename T>
+    using ColumnOptConst = db::ColumnConst<std::optional<T>>;
+
+    TuringSink() = delete;
```
cyrusknopf: You don't need to delete this

net/decoders/Decoders.h
```diff
@@ -0,0 +1,679 @@
...
+
+    return true;
+}
+
+// The dispatch layer always supplies T explicitly (it comes from the wire type code);
+// Sink is deduced from the sink argument. Each decode family is a single function
+// selecting the wire format with if constexpr on T; the trailing else handles the
+// trivial fixed-width types.
+template <typename T, typename Sink>
+inline bool decodeVector(DecodeContext* context,
```
cyrusknopf: this should definitely be specialised on T instead of so many if constexpr blocks

net/decoders/Decoders.h
```diff
@@ -0,0 +1,679 @@
...
+            return false;
+        }
+
+        context->_inBuf->readData(typedColumn->data() + context->_rowIndex, rowsRemaining * sizeof(T));
+        return true;
+    }
+}
+
+template <typename T, typename Sink>
+inline bool decodeOptVector(DecodeContext* context,
```
cyrusknopf: Same comment for specialisation vs if constexpr for this and const decoder

net/decoders/TuringProtoDecoder.h
cyrusknopf: Do you have those changes?

net/decoders/TuringProtoDecoderConcepts.h
```diff
@@ -44,15 +43,15 @@ concept SupportedColumnVectorTypes = std::is_same_v<T, db::types::UInt64::Primit
 || std::is_same_v<T, db::EntityList>
 || std::is_same_v<T, db::ListElementView>
 || std::is_same_v<T, db::ListView>
-|| std::is_same_v<T, std::string>;
+|| std::is_same_v<T, db::types::String::Primitive>;
```
cyrusknopf, line 45: Very nice

net/decoders/TuringSinkColumnContainer.h
```diff
@@ -0,0 +1,34 @@
...
+}
+
+namespace net::proto {
+
+// Column-container view over a db::Dataframe for the decode sink. The turing proto decoder
+// works against containers that reference columns directly, so this adapter hides the
+// NamedColumn indirection we use in the Dataframe class.
+class TuringSinkColumnContainer {
+public:
+    TuringSinkColumnContainer() = delete;
```
cyrusknopf: This is implictly deleted

net/decoders/TuringProtoDecoder.h
```diff
@@ -0,0 +1,388 @@
...
+class TuringProtoDecoder {
+public:
```
cyrusknopf, line 28: I think it would be better to put all the implementations in this file that can be in a .cpp and explicitly instantiate the template for Sink=TuringSink since there is a known closed set of Sink parameters

net/decoders/TuringProtoDecoder.h
```diff
@@ -0,0 +1,388 @@
...
+    // Decode an incoming data chunk - we first check if we have some unfinished state
+    // from the previous chunk and fill that up first
+    void decodeIncomingChunk(typename Sink::ColumnContainer* container);
+    // Decode new data from the incoming chunk
+    void decodeIncomingData(typename Sink::ColumnContainer* container);
+
+    void reset();
+
+    template <typename T>
+    bool decodeColumn(typename Sink::template ColumnVector<T>* column, ProtoColumnState* columnState);
```
cyrusknopf: Since its the template ragument you should be able to typedef all of the Sink-specific types you use in the class to make the signatures less ugly.

Also, all sinks need to agree on the name of the column types? e.g. a WASM sink would need a type called `ColumnVector<T>` as well, even if it is not representing a TuringDB column vector? do we want that naming scheme or make the sink template names more generic?

net/decoders/TuringProtoDecoder.h
```diff
@@ -0,0 +1,388 @@
...
+        case ColumnInternalKind::LABEL_ID:
+            return fn.template operator()<db::LabelID>(encoding);
+        case ColumnInternalKind::LABEL_SET_ID:
+            return fn.template operator()<db::LabelSetID>(encoding);
+        case ColumnInternalKind::CHANGE_ID:
+            return fn.template operator()<db::ChangeID>(encoding);
+        case ColumnInternalKind::PROPERTY_NULL:
+            return fn.template operator()<db::PropertyNull>(encoding);
+        default:
+            break;
```
cyrusknopf: I think we should favour removing the default case so that when adding new values it requires explicit handling

net/decoders/TuringProtoDecoder.h
sulaimansuhas: They should be in now.

net/decoders/TuringProtoDecoder.h
sulaimansuhas: good points.

net/decoders/TuringSinkColumnContainer.h
sulaimansuhas: why would it be?

cyrusknopf: default constructor is implicitly deleted if there is a user-defined constructor

net/decoders/TuringProtoDecoder.h
sulaimansuhas: This gets a bit complex with the the library structuring and basically becomes a infeasible with the wasm build. I think it's better that we leave it as this for now.

net/decoders/TuringProtoDecoder.h
sulaimansuhas: Very good point, I caught that we aren't decoding CommitHashID types haha.

net/decoders/TuringProtoDecoder.h
sulaimansuhas: I've left the column renaming for now - can pick it up in later refactors.

rjb32 (approved):

PR #850, opened 2026-09-03
Migrate `labels()` to return a column of `std::string_view`

cyrusknopf:


cyrusknopf:
Closed; since #910 `labels()` no longer returns a string-type at all

PR #873, opened 2026-09-07
Storage: compare a mask like any other boolean column

rjb32:
Allow a mask operand in equality.

### DIFF
--- a/storage/columns/AllowedKinds.h
+++ b/storage/columns/AllowedKinds.h
@@ -187,8 +187,7 @@ struct PairRestrictions<Op> {
     >;
 
     using Excluded = ExcludedContainers<
-        ContainerKind::code<ColumnSet>(),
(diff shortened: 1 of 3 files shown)


storage/columns/AllowedKinds.h
```diff
@@ -186,9 +186,10 @@ struct PairRestrictions<Op> {
         MixedKind<ColumnMask, std::optional<CustomBool>>
     >;
 
+    // A predicate over two non-nullable columns answers in a mask, so a mask is one of the
+    // shapes a boolean operand comes in and equality has to take it like the others.
```
cyrusknopf: I think the comment is unnecessary

rjb32: Yes I just removed it

cyrusknopf (approved):

PR #874, opened 2026-09-08
MLIR: Fix reads (e.g. `labels(n)`) on a `CREATE`d variable

cyrusknopf:
Populates a created entity in the MLIR executor to store the labels and edge types (as strings) such that they can be returned by the `labels()` and `edgeType()` functions, respectively.


rjb32:
Please test with a WITH barrier CREATE (n:Tag) WITH n RETURN labels(n)

rjb32 (approved):

PR #886, opened 2026-09-09
Feature/enable v3 for bin proto

sulaimansuhas:


net/decoders/TuringSinkColumnContainer.h
```diff
@@ -0,0 +1,33 @@
...
+class Dataframe;
+class DataframeManager;
+}
+
+namespace net::proto {
+
+// Column-container view over a db::Dataframe for the decode sink. The turing proto decoder
+// works against containers that reference columns directly, so this adapter hides the
+// NamedColumn indirection we use in the Dataframe class.
+class TuringSinkColumnContainer {
```
rjb32, line 17: We should probably have a v3/v2 version of that because maybe we could have huge simplication if we are just dealing with columns coming from LocalMemory instead of having to deal with dataframes, DataframeManager and NamedColumn

net/encoders/TuringProtoEncoder.h
```diff
@@ -231,9 +210,14 @@ struct ListElementWriteVisitor {
 
 class DataWriter {
 public:
-    explicit DataWriter(net::proto::TuringProtoOutBuf* outBuf, std::stack<NestedListStackElement>& stack)
+    DataWriter(net::proto::TuringProtoOutBuf* outBuf,
+               std::stack<NestedListStackElement>& stack,
+               size_t offset,
```
rjb32, line 237: Why do we pass offset and rowCount here and how does it relates to columns being written?

net/encoders/TuringProtoEncoder.h
```diff
@@ -231,9 +210,14 @@ struct ListElementWriteVisitor {
 
 class DataWriter {
 public:
-    explicit DataWriter(net::proto::TuringProtoOutBuf* outBuf, std::stack<NestedListStackElement>& stack)
+    DataWriter(net::proto::TuringProtoOutBuf* outBuf,
+               std::stack<NestedListStackElement>& stack,
```
rjb32, line 236: Why the stack must be passed as argument and what is it not an internal state?

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -3589,11 +3238,14 @@ NLExecutor::~NLExecutor() {
...
+        NLOutputData* const outputData = _prog->getOutputData();
+        std::span<const Column* const> outputs;
+        if (outputData) {
+            outputs = outputData->outputs();
+        }
 
-    const bool hasNamesToPublish = sink && !columnNames.empty();
-    if (hasNamesToPublish) {
-        sink->setColumnNames(columnNames);
+        sink->declareOutput(_prog->columnNames(), outputs);
```
rjb32, line 3620: What if columnNames is empty? Could you test for queries that have no results and for queries that return a constant such as RETURN 42

net/encoders/TuringProtoEncoder.h
sulaimansuhas: `SKIP` and `LIMIT` are implemented with these offset and rowCount values. So the DataWriter object which is instantiated per chunk is constructed with these as they are re-used to get the correct view of the columns we have outputted.

net/encoders/TuringProtoEncoder.h
sulaimansuhas: The `DataWriter` class is created once per chunk, and is the wrapper class for the functions responsible for writing the turingdb types to network bytes. So it isn't supposed to keep any persistent internal state. Also the stack can be encoding a list view that spans multiple chunks so it can't be local to the DataWriter class.

net/encoders/TuringProtoEncoder.cpp
```diff
@@ -14,50 +16,86 @@ TuringProtoEncoder::TuringProtoEncoder(net::proto::TuringProtoOutBuf* outBuf)
...
+    size_t chunkHeaderLen = sizeof(WireSize);
+    for (const std::string_view name : names) {
+        chunkHeaderLen += net::proto::ColumnWireHeader::wireSize();
+        chunkHeaderLen += name.size();
+    }
+    bioassert(chunkHeaderLen <= _outBuf->capacity(), "Column schema exceeds maximum chunk size");
+
+    writeColumnCount(cols.size());
+
+    for (const auto [name, column] : std::views::zip(names, cols)) {
```
rjb32: This will fail to compile on ubuntu 22.04

net/encoders/TuringProtoEncoder.cpp
```diff
@@ -14,50 +16,86 @@ TuringProtoEncoder::TuringProtoEncoder(net::proto::TuringProtoOutBuf* outBuf)
...
+void TuringProtoEncoder::writeColumnHeaders(std::span<const std::string_view> names,
+                                            std::span<const db::Column* const> cols) {
+    size_t chunkHeaderLen = sizeof(WireSize);
+    for (const std::string_view name : names) {
+        chunkHeaderLen += net::proto::ColumnWireHeader::wireSize();
+        chunkHeaderLen += name.size();
+    }
+    bioassert(chunkHeaderLen <= _outBuf->capacity(), "Column schema exceeds maximum chunk size");
+
+    writeColumnCount(cols.size());
```
rjb32: column count is sized by cols.size but the zip will depend on the names size

net/encoders/TuringProtoEncoder.h
rjb32, line 387: Discards the row count for a ColumnConst so will not work in v3 for MATCH (n) RETURN 5

sulaimansuhas: what is the expected result here? That 5 will be outputted multiple times?

rjb32: Yes. v3 uses the row count in ColumnConst to give the number of times the constant must be replicated because in the case of MATCH (n) RETURN 5 the column for n that drives the row count may not be outputted at all.

sulaimansuhas: I see, interestingly this is pre-existing behaviour in the protocol. We do not have a test that catches this case.

query/ir/interpreter/NLExecutor.cpp
sulaimansuhas:
```
        const bool hasExplicitAlias = !item->getName().empty();

        if (hasExplicitAlias) {
            name = item->getName();
        } else if (item->getKind() == Expr::Kind::SYMBOL) {
            const SymbolExpr* symbolExpr = static_cast<const SymbolExpr*>(item);
            const Symbol* symbol = symbolExpr->getSymbol();
            name = symbol->getName();
        } else {
            const std::string_view generatedName =
                srcMan->getStringRepr(std::bit_cast<std::uintptr_t>(item));
            if (generatedName.empty()) [[unlikely]] {
                throwError("Failed to generate name for projection item", item);
            }

            item->setName(generatedName);
            name = generatedName;
        }
```
It looks like in the query path all columns are given names. So we shouldn't run into this on the cypher path.

net/encoders/TuringProtoEncoder.cpp
sulaimansuhas: there is a bioassert ensuring they are the same size in this code path.

rjb32 (approved):

PR #902, opened 2026-09-12
Cypher: implement CREATE followed by WITH

rjb32:
Implement CREATE followed by WITH.

```
CREATE (n:Person {name: 'Zoe'}) WITH n RETURN n.name

MATCH (p:Person) CREATE (n:Person {name: p.name}) WITH n ORDER BY n.name LIMIT 2 RETURN n.name

CREATE (a:Person {name: 'Ana'}) WITH a CREATE (a)-[:KNOWS]->(b:Person {name: 'Bo'})

CREATE (a:Person {name: 'Eva'})-[e:KNOWS {since: 1999}]->(b:Person {name: 'Fay'}) RETURN e.since

CREATE (a:Person {name: 'Ana'})-[:MENTORS]->(b:Brand {name: 'Ora'}) WITH a MATCH (a)-[e]->(m) WITH e, m RETURN type(e), m.name

CREATE (n:Person {name: 'Rex'}) WITH n OPTIONAL MATCH (p:Person {name: 'Remy'})-[:KNOWS_WELL]->(f) RETURN n.name, f.name

CREATE (n:Person {name: 'Zoe'}) WITH n MATCH (p:Person) ORDER BY p.name LIMIT 1 MATCH (i:Interest {name: 'Cooking'}) RETURN n.name, labels(n), p.name
```

* Parser: singlePartQuery is now one flat list and remove the readingStatements preceding updateStatements because CREATE WITH is a read statement following an update statement
* Analyzer: a read after an update must be separated by a WITH, it delimits a query part. This is why we can't express CREATE (n) MATCH (n) but we must use CREATE.. WITH n ..
* Analyzer: a part that ends on a read needs a RETURN, so CREATE (n) WITH n MATCH (m) is rejected
* An OPTIONAL MATCH whose pattern names what a CREATE wrote is rejected; one that only carries it past the join is not
* A MATCH's own ORDER BY / SKIP / LIMIT closes the part when a MATCH or a literal UNWIND follows it.
* Add an all_pending flag to operations such as get_node_properties to say that this input is to be answered from the CommitWriteBuffer


cyrusknopf:
Would be good to have a high level explanation as to how this is implemented

rjb32:
> Would be good to have a high level explanation as to how this is implemented

Indeed, done

cyrusknopf:
> > Would be good to have a high level explanation as to how this is implemented
> 
> Indeed, done

Can you try

`CREATE (n:New) WITH n MATCH (n)-->(m) RETURN n, m`

and similarly

`CREATE (n:New)-[:NewEdge]->(:New) WITH n MATCH (n)-->(m) RETURN n, m`

rjb32:
> > > Would be good to have a high level explanation as to how this is implemented
> > 
> > 
> > Indeed, done
> 
> Can you try
> 
> `CREATE (n:New) WITH n MATCH (n)-->(m) RETURN n, m`
> 
> and similarly
> 
> `CREATE (n:New)-[:NewEdge]->(:New) WITH n MATCH (n)-->(m) RETURN n, m`

Tests added and fixed

PR #912, opened 2026-09-15
MLIR: fix the five Cypher bugs pinned by QA on 2026-09-12

rjb32:
Fix the five Cypher queries v3 answers wrong.


query/analyzer/ExprAnalyzer.cpp
```diff
@@ -381,6 +405,11 @@ void ExprAnalyzer::analyzeBinaryExpr(BinaryExpr* expr) {
                 break;
             }
 
+            if (_isV3 && computesOverListItem(pair)) {
```
rjb32, line 408: Why if computesOverListItem it means the type should be Double

query/ir/interpreter/NLExecutor.cpp
```diff
@@ -935,6 +935,22 @@ void broadcastNullableConstantColumn(const Column* value, size_t rowCount, Colum
     std::fill_n(outputRaw.begin(), rowCount, typedValue->getRaw());
 }
 
+// A chunk standing for every row is a literal's ColumnConst only where a literal bound
```
rjb32: What does this mean?

query/analyzer/ExprAnalyzer.cpp
rjb32: It should not, and no longer is. An index into a list homogeneous in one type is now typed by that type, as an UNWIND of the same list already was (`ListShape::unwoundType`), so `[1,2,3][1] + 1` answers the integer 3 rather than 3.0. `ListItem` -> Double is left for a list that names no one type.

The engine follows: `binaryResultElement`'s Index case returns `nullable<i64>` where the list chunk is a `list<i64>`, and `nl.list_index` binds a kernel that reads each cell out as that type (`ValueListIndexImpl`). A mixed list still lowers to `nullable<list_element>` and keeps the tagged-cell kernel.

```
RETURN [7,2][0] / 2                -> 3      (was 3.5)
MATCH (n) ... RETURN [n.age][0] / 5 -> 6     (was 6.4)
RETURN [7,2.0][0] / 2              -> 3.5    (mixed list, unchanged)
RETURN ['ab','cd'][0] + 'ef'        -> abef   (was rejected)
```

`ListIndexTest.typesAnIndexIntoAMixedListTheSameWay` asserted the old rule and is replaced by two tests, one per side.

query/ir/interpreter/NLExecutor.cpp
rjb32: Unreadable — rewritten. It means: `selectConstantBroadcast` picks this function when the "constant" chunk is not a `ColumnConst`. A kernel computing over literals writes its result as one row of an ordinary column, and that row is what gets laid out over the step.

Say the word if you would rather it just go away.

PR #913, opened 2026-09-15
Implement a GraphSAGE procedure

cyrusknopf:


procedures/GraphSAGEProcedure.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+        GraphSAGESampler::NodeCol* dst = nodeColumn(data, base);
+        GraphSAGESampler::NodeCol* srcs = nodeColumn(data, base + 1);
+        GraphSAGESampler::NodeCol* tgts = nodeColumn(data, base + 2);
+
+        data.sampler->setHopData(hop, srcs, tgts, dst, fanouts[hop]);
+    }
+
+    ColumnNodeIDs nodes;
+    const Column* x = data.getInputColumn(0);
```
rjb32: FInd a better name than x for this variable

storage/samplers/GraphSAGESampler.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "columns/ColumnOptVector.h"
+
+using namespace db;
+
+namespace {
+
+const auto resizeImpl = [](auto* col, size_t size) -> void { col->resize(size); };
+const auto clearImpl = [](auto* col) -> void { col->clear(); };
+
+void colAssign(const ColumnNodeIDs* src, ColumnOptVector<NodeID>* dst) {
```
rjb32: Start the method name by the action verb in general, such as assignColumn

storage/samplers/GraphSAGESampler.cpp
```diff
@@ -0,0 +1,206 @@
...
+            : std::make_unique<NeighbourhoodSampleChunkWriter>(_view, &data._frontier, data._fanout);
+    }
+
+    std::unique_ptr<NeighbourhoodSampleChunkWriter>& writer = data._writer;
+
+    if (writer->isDone()) {
+        return 0;
+    }
+
+    ColumnNodeIDs srcs;
```
rjb32: Why are there columns recreated here?

cyrusknopf: The neighbourhood sample iterator takes `ColumnNodeIDs` as input, but the output of the sampler is `ColumnOptVector<NodeID>` so there is a conversion that has to occur. This could be fixed by templating the neighbourhood sampler on the column type

procedures/GraphSAGEProcedure.cpp
cyrusknopf: Done

storage/samplers/GraphSAGESampler.cpp
cyrusknopf: Fixed

rjb32 (approved):

PR #914, opened 2026-09-15
Test: test shortest path reads the newest committed edge weight

rjb32:
Shortest path uses an iterator without null that does not see correctly the newer version of a property

```
CREATE (a:City {id: 'a'}), (b:City {id: 'b'}), (a)-[:ROAD {distance: 1.0}]->(b)
COMMIT; CHANGE SUBMIT

MATCH (a:City)-[r:ROAD]->(b:City) WHERE a.id = 'a' AND b.id = 'b' SET r.distance = 5.0
COMMIT; CHANGE SUBMIT

getEdgePropertiesWithNull             distance=5
shortest path a -> b                         sees old distance=1
```

ShortestPathStaleWeightTest.readsTheWeightCommittedByTheNewerDatapart fails on main.


rjb32:
@cyrusknopf test case currently failing in main related to GetEdgeProperties iterator without nulls

PR #922, opened 2026-09-16
Add Map Literal Support

sulaimansuhas:
Implement map literals, and change db.listNodes to take one properties MAP in place of its propertyKeys/propertyValues list pair.

```
before
db.listNodes(labels :: LIST, propertyKeys :: LIST, propertyValues :: LIST, skip :: INT64, limit :: INT64)
CALL db.listNodes([], ['name'], ['remy'], 0, 1000) YIELD id, labels, properties

after
db.listNodes(labels :: LIST, properties :: MAP, skip :: INT64, limit :: INT64)
CALL db.listNodes([], {name: 'remy'}, 0, 1000) YIELD id, labels, properties

RETURN {a: 1, b: "hello", c: true, d: 1.5, e: null}
RETURN {a: 1, b: [1, 2, 3], c: {d: 3, e: "x"}}
RETURN {outer: {mid: {inner: 1}}, tail: 9}
```


query/ir/codegen/DBProgramGenerator.cpp
```diff
@@ -806,6 +814,42 @@ mlir::Value DBProgramGenerator::translateListOfColumns(const ListLiteral* list)
...
+                                                                            _opBuilder.getDictionaryAttr(entries));
+
+    return constant.getResult();
+}
+
+void DBProgramGenerator::translateMapEntries(const MapLiteral* map,
+                                             llvm::SmallVectorImpl<mlir::NamedAttribute>& entries) {
+    entries.reserve(map->size());
+
+    for (const auto& [key, valueExpr] : *map) {
```
rjb32, line 831: We are iterating over an unordered map with a pointer key here.
We should not even use an unordered_map because if 2 keys have the same name it will be two Symbol* pointers for the same string. So the 2 duplicates are kept in MapLiteral, but this is passed later to db.constant as a string dictionary and LLVM trips an assert.

```
RETURN {a: 1, a: 2}
```

rjb32:
LIMIT/SKIP after a map literal is not supported:
```
MATCH (n:Person) RETURN {tag: 'x'} LIMIT 1
```

rjb32 (approved):

PR #929, opened 2026-09-17
Add npm Release Step

sulaimansuhas:


.github/workflows/ci_release.yml
```diff
@@ -30,6 +30,13 @@ jobs:
       dev_version_suffix: ${{ needs.setup.outputs.dev_version_suffix }}
     secrets: inherit
 
+  build-npm:
```
rjb32, line 33: Remove it and put it in ci_build.yml if not already built automatically in main build

.github/workflows/ci_wasm.yml
```diff
@@ -0,0 +1,99 @@
...
+        fetch-depth: 500
+
+    - name: Fetch tags
+      run: git fetch --tags --force
+
+    # The decoder compiles spdlog's bundled fmt header-only; external/expected is
+    # in-tree. Nothing else the wasm build needs comes from a submodule, so this
+    # skips the LLVM/OpenBLAS/curl trees the native build pulls.
+    - name: Checkout spdlog
+      run: |
```
rjb32, line 45: Remove all dependencies and emscripten install from this workflow and use the same cache key as for ci_build.yml

PR #933, opened 2026-09-18
CYPHER: Implement `REMOVE`

cyrusknopf:
`REMOVE n.X` sets property `X` to null for `n` if it exists, or if it does not exist, is a noop


rjb32:
Please generate more SET to null cypher tests, particularly involving lists, embeddings, optional match, multipart queries

cyrusknopf:
> Please generate more SET to null cypher tests, particularly involving lists, embeddings, optional match, multipart queries

Done. Also I added your commit in #951 directly here rather than stacking, as the stack had conflicts that GitHub doesn't let you fix without installing the GitHub CLI tool

rjb32 (approved):

PR #937, opened 2026-09-18
Add Edge Disjunction

sulaimansuhas:


rjb32:
Check the db dialect IR generated for:
```Cypher
MATCH (a)-[e]->(b) WHERE e:KNOWS|KNOWS RETURN a
``` 
and check it does not emit twice the edge type check

rjb32:
Add a test for a merge query such that:
```
MERGE (a:Person)-[:KNOWS_WELL|INTERESTED_IN]->(b:Person)
```

sulaimansuhas:
> Add a test for a merge query such that:
> 
> ```
> MERGE (a:Person)-[:KNOWS_WELL|INTERESTED_IN]->(b:Person)
> ```

This query is supposed to fail right?

rjb32:
> > Add a test for a merge query such that:
> > ```
> > MERGE (a:Person)-[:KNOWS_WELL|INTERESTED_IN]->(b:Person)
> > ```
> 
> This query is supposed to fail right?

Yes it must be rejected because only one edge type can be created.

sulaimansuhas:
> Check the db dialect IR generated for:
> 
> ```cypher
> MATCH (a)-[e]->(b) WHERE e:KNOWS|KNOWS RETURN a
> ```
> 
> and check it does not emit twice the edge type check

The DB dialect should be generated as plainly as possible if I am recalling correctly - so I have added a pass here to find filter out these cases.

rjb32 (approved):

PR #941, opened 2026-09-21
Change `NeighbourhoodSampleIterator` to fetch incoming edges

cyrusknopf:


examples/SimpleGraph.cpp
```diff
@@ -65,6 +65,28 @@ void SimpleGraph::findOutEdges(Graph* graph,
...
+    edgeTypes.clear();
+    sourceNodeIDs.clear();
+
+    const auto transaction = graph->openTransaction();
+    const auto reader = transaction.readGraph();
+
+    ColumnNodeIDs columnNodeIDs(nodeIDs);
+
+    const auto edges = reader.getInEdges(&columnNodeIDs);
+    for (const auto& edge : edges) {
```
rjb32, line 83: Could we use reserve here if we know the size of the range?

cyrusknopf: The size is not known ahead of time, it is lazily generated via `edges` which wraps an iterator

rjb32 (approved):

